PR #171, opened 2013-11-27
drop http_proxy clearing

vapier:
The boto logic takes care of merging the env vars and the settings in its
boto config file, so there shouldn't be a need to clear things.


vapier:
with 4.8 now supporting things, we don't need this anymore.  thanks all!

PR #240, opened 2014-12-31
Only use proxy settings from boto if they are valid.

rakuco:
Do not assume .boto always has proper proxy settings. For example, this
is not the case for Chromium+depot_tools users, which are not even
required to have a boto configuration file around.

If the `ProxyInfo` object created by using the settings from boto is not
valid, do not pass `proxy_info` to `httplib2.Http` and instead rely on it
figuring out the proxy settings on its own (at present, it will look at
the user's proxy environment variables).

BUG=chromium:443523


rakuco:
> Can you file an issue that clearly describes the problem you're trying to solve?

Done in #241. I've also mentioned there why #171 is actually related to this problem.

rakuco:
Closing, according to https://github.com/GoogleCloudPlatform/gsutil/issues/241#issuecomment-69862763 a different solution was implemented and will be part of version 4.8.

PR #250, opened 2015-02-11
Modified acl commands to support project entities.

fishjord:


thobrla:
Waiting for mfschwartz's LGTM on the review, then will merge.

PR #251, opened 2015-02-13
Don't retry on HTTP 401 errors but fail immediately.

primiano:
Under some rare circumstances (details below) gsutil can end up
spending long times retrying on HTTP 401 (Unauthorized) errors. There
is no point retrying when the request auth fails. In such case gsutil
should just bail out immediately.

Description of the problem:
Not sure what triggered this problem in the first place, but my gsutil
ended up in a state where the SSO cookie was still in place (so gsutil
did NOT bail out with the "No handler was ready to authenticate" msg)
but the server was rejecting the cookie and returning 401 requests.
Under the hoods the code in http_wrapper was retrying the request with
backoff as if there was no tomorrow.

### DIFF
--- a/gslib/third_party/storage_apitools/http_wrapper.py
+++ b/gslib/third_party/storage_apitools/http_wrapper.py
@@ -251,7 +251,8 @@ def HandleExceptionsAndRebuildHttpConnections(retry_args):
   elif isinstance(retry_args.exc, exceptions.RequestError):
     logging.debug('Request returned no response, retrying')
   # API-level failures
-  elif isinstance(retry_args.exc, exceptions.BadStatusCodeError):
+  elif (isinstance(retry_args.exc, exceptions.BadStatusCodeError) and
+        retry_args.exc.status_code != httplib.UNAUTHORIZED):
     logging.debug('Response returned status %s, retrying',
                   retry_args.exc.status_code)
   elif isinstance(retry_args.exc, exceptions.RetryAfterError):


primiano:
More context: gsutil is used by chromium as a building block of its infrastructure.
Under the circumstances explained in the PR, this can causes long (and useless) waiting times in the bot / contributor workflows.

Capstan:
Will this neuter the access_token refreshing behavior or is that accounted for elsewhere?

thobrla:
We put this in back in 4.5 (commit https://github.com/GoogleCloudPlatform/gsutil/commit/1a0e4d8d46fa4095e1defe4c713945bc8e242262) because we were seeing 401's that would terminate long-running uploads on Google Compute Enigne.  In theory, httplib2+oauth2client should protect us from this, but in practice it was not.

I would love to get rid of this.  Unfortunately, before we do that, we'll need extensive testing and debugging to validate that the Google Compute Engine 401 issues no longer occur.  Once we've done that, instead of your change, we should just drop the one-off code from CheckResponse that treats a 401 as a BadStatusCodeError in the first place.

primiano:
Not sure I fully understand the proposed solution: how are we going to deal with this problem?

thobrla:
This pull request is now obsolete.  Starting with gsutil 4.13 we moved past this [apitools 0.4.5](https://github.com/google/apitools/commit/0b6a9f5a30f8dc73e4f5fc8176b94d3e96d19f78) commit so we don't retry 401s anymore.

PR #268, opened 2015-04-29
Fix stat.py - error printing goes to stderr

7heo:
I am currently writing a script that uses gsutil, and having the error printing go to stdout is a real bummer.

### DIFF
--- a/gslib/commands/stat.py
+++ b/gslib/commands/stat.py
@@ -16,7 +16,7 @@
 
 from __future__ import absolute_import
 
-import logging
+import logging, sys
 
 from gslib.bucket_listing_ref import BucketListingObject
 from gslib.cloud_api import AccessDeniedException
@@ -145,7 +145,7 @@ class StatCommand(Command):
         pass
       if not arg_matches:
         if logging.getLogger().isEnabledFor(logging.INFO):
-          print 'No URLs matched %s' % url_str
+          print >>sys.stderr, 'No URLs matched %s' % url_str
         found_nonmatching_arg = True
     if found_nonmatching_arg:
       return 1


Capstan:
Why does it `print` at all, instead of calling `logging.getLogger().info(...)`? Is this to ensure actual output even if logging is disabled?

jterrace:
I think this should actually be a `raise CommandException('No URLs matched')` since that's what we do in several other places.

thobrla:
Fix for this is in the works; we'll also be fixing the ACL message.

gslib/commands/stat.py
```diff
@@ -16,7 +16,7 @@
 
 from __future__ import absolute_import
 
-import logging
+import logging, sys
```
thobrla: For future reference, gsutil style is imports on a single line.

7heo: @thobrla thanks, I'll keep that in mind.

7heo:
@Capstan @jterrace I usually try to keep the changes minimal when I contribute to a project; hence the diff. I however agree than raising an exception would be much more consistent with the rest of the code; and overall better.

thobrla:
This is fixed for gsutil 4.13 with https://github.com/GoogleCloudPlatform/gsutil/commit/3b9dac53b220b62e71254b8f53987af5b4f3f4eb

Note: raising a CommandException would diverge from Unix stat, which outputs an error and continues when multiple arguments are specified and an earlier argument does not exist.

Thus, the fixed code continues to iterate until all arguments have been evaluated, and returns a nonzero exit code if any of the arguments resulted in a failure.

7heo:
@thobrla when is the 4.13 release planned? 'cause in the meanwhile, I have to work with the broken 4.12... So the earlier the better :)

thobrla:
We don't have an official release date of gsutil 4.13 yet, hopefully it will be within the next month.  In the meantime, if you're using the tarball, you can update to a prerelease version which includes this fix by running:
gsutil update gs://prerelease/gsutil_4.13pre_rel1.tar.gz

(note: you shouldn't depend on this version as it will go away after 4.13 official release).

PR #286, opened 2015-07-08
Updating wording according to suggestions by Mike

feczo:
same thing slightly different wording


feczo:
Updating wording according to suggestions by Mike same thing slightly different wording

PR #299, opened 2015-09-25
Fix rsync doc error about RE needed to match end of string

zbjornson:
The docs said 

> matching any string ending in 'abc' would be specified using '.*abc'

but that will match "foo.abc.bar".


mfschwartz:
Thanks for the doc fix

PR #328, opened 2016-01-27
added -t/--token flag for google oauth2 token

ayush--s:


ayush--s:
eh..sorry..wrong pull req.. :)

PR #339, opened 2016-03-05
Update dependent libraries to pick up GCE metadata client changes.

Galabar001:
The GCE metadata server legacy endpoints (0.1, v1beta1) have been deprecated.
This change brings in dependent libraries that eliminate the use of these endpoints.


thobrla:
There's still some internal work and discussion that likely needs to go on before we can accept this PR.

PR #341, opened 2016-03-15
Disable all gsutil config flags except -a when running in the Cloud SDK.

dlorenc:
https://codereview.appspot.com/294720043/


dlorenc:
cc @thobrla

PR #377, opened 2016-08-04
Clarification to Parallel composite uploads docs

melissachang:


thobrla:
Bypassing checks and merging since this is a comment-only change.  Thanks for the doc update.

PR #387, opened 2016-10-06
Mark protection mode dashes as source code

fredsa:


thobrla:
Following up internally to make sure this will work as intended.

houglum:
Closing since this is a couple years old.

Also, to follow up on thobrla's comment, I believe that should be provided in double-backticks, e.g.

```
``These three words`` get turned into code format when we generate web docs
```

PR #391, opened 2016-11-16
Reduce duplication in defacl ch error message

ixdy:
Before this change:

```console
$ gsutil defacl ch -u foo@example.com:bar gs://example
CommandException: foo@example.com:bar is not a valid ACL change
Allowed permissions are READER, OWNER, OWNER, WRITER, OWNER, WRITER, READER
```

After this change:
```console
$ ./gsutil defacl ch -u foo@example.com:bar gs://example
CommandException: foo@example.com:bar is not a valid ACL change
Allowed permissions are OWNER, WRITER, READER
```


thobrla:
Thanks!

PR #408, opened 2017-02-04
- SSL certs

rjschwei:
+ Recognize the "system" keyword in boto.cfg to support the use of the
    system installed certs

### DIFF
--- a/gslib/util.py
+++ b/gslib/util.py
@@ -680,6 +680,12 @@ def GetCertsFile():
     string filename of the certs file to use.
   """
   certs_file = boto.config.get('Boto', 'ca_certificates_file', None)
+  # The 'system' keyword indicates to use the system installed certs. Some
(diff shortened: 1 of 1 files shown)


gslib/util.py
```diff
@@ -704,6 +704,12 @@ def GetCertsFile():
           configured_certs_files.append(fname)
           disk_certs_file = fname
       certs_file = disk_certs_file
+  # The 'system' keyword indicates to use the system installed certs. Some
+  # Linux distributions patch the stack such that the Python SSL
+  # infrastructure picks up the system installed certs by default, thus no
+  #  action necessary on our part
+  if certs_file == 'system':
```
thobrla: Since this is mutually exclusive with the code block above, could you please move it to the top of the function and just return "None" there?

rjschwei: OK, done

thobrla:
Thanks for adding this!

PR #414, opened 2017-03-15
rsync: add support for canned ACLs

ishitatsuyuki:
The implementation is mostly copied from CpCommand.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -1907,6 +1907,33 @@ class TestRsync(testcase.GsUtilIntegrationTestCase):
           return_stderr=True))
     _Check2()
 
+  def test_rsync_canned_acl(self):
+    """Tests that rsync -a applies ACLs."""
+    bucket1_uri = self.CreateBucket()
+    bucket2_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket1_uri, object_name='obj1',
+                      contents='obj1')
+
+    # Use @Retry as hedge against bucket listing eventual consistency.
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _Check():
+      """Tests rsync -a works as expected."""
+      self.RunGsUtil(['rsync', '-d', '-a', 'public-read', suri(bucket1_uri),
+                      suri(bucket2_uri)])
+      listing1 = TailSet(suri(bucket1_uri), self.FlatListBucket(bucket1_uri))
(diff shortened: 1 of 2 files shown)


houglum:
Seems pretty reasonable. Could you include a test, too? (See `test_canned_acl_cp()` in `gslib/tests/test_cp.py`.)

ishitatsuyuki:
Added, please take a look!

gslib/tests/test_rsync.py
```diff
@@ -1907,6 +1907,33 @@ def _Check2():
           return_stderr=True))
     _Check2()
 
+  def test_rsync_canned_acl():
```
houglum: This test method should have `self` as a parameter.

ishitatsuyuki:
Seems I can't manage to make the test to run. Please take a look where I am wrong.

houglum:
I'm not sure how to troubleshoot without more details - what do you mean by you couldn't manage to make the test run? What command did you try, what error did you get, etc.?

Assuming you have a `~/.boto` file set up with valid credentials, you should be able to run the rsync tests for both of the GCS APIs via the two commands below (which assume your cwd is the root of your local gsutil repository):

    NUM_PROCESSES=50
    ./gsutil -o 'GSUtil:prefer_api=json' test -p $NUM_PROCESSES rsync
    ./gsutil -o 'GSUtil:prefer_api=xml' test -p $NUM_PROCESSES rsync

ishitatsuyuki:
I didn't see the test count increase in CI. I guessed that the test haven't run.

houglum:
A run of Travis CI will be triggered when you push a new commit. Looking at your patch-1 branch, I only see two commits -- I assumed there would be a third commit containing the addition of the `self` parameter.

Anyway, looking at the "Show all checks" > "Details" link, it shows those tests were run at [0253355](https://github.com/GoogleCloudPlatform/gsutil/commit/025335523a4573d38c674292a00702af0f98af70), which appears to include the most recent changes.

As this is an integration test (which Travis CI can't run without us publishing some credentials open to the world -- we only make it run `gsutil test -u` to do the unit tests), I also patched these changes in manually and ran the integration tests myself -- looks good.

ishitatsuyuki:
I see, it was an integration test not unit test. Thanks for verifying! Is this ready to merge now?

ishitatsuyuki:
(I used force push to fix the second commit, since I don't like dirty histories)

houglum:
Done! Speaking of "dirty" histories, GitHub apparently lets me squash all the commits into one when I merge a PR now. Neat.

Thanks for the fix!

Issue #423, opened 2017-04-11
invalid_grant caused by copy/paste selection - we should offer helpful guidance on this


miketruty:
# Problem
When you run **gsutil config** you copy/paste an authorization code from your browser.  But selecting this code (on my desktop/browser - Ubuntu+Chrome) takes a little diligence. The code is not autoselected.  The code starts with **4/** and this is not selected via double-click.  There can be other characters present that confuse the tokenizer on double-click. The code extends beyond the space of the box width so you have to be careful about selecting it with click drag.

# Suggestion
When we raise this particular exception (invalid_grant) perhaps we can/should offer a set of tips to resolve?

```
You should try:
* make sure you've selected and pasted the ENTIRE authorization code (including any #/ prefix).
* try running 'gcloud auth login' to refresh your credentials.
* try running 'gcloud components update' to use the latest version of the tools.
```

# Example of authorization code mis-selection
![image](https://cloud.githubusercontent.com/assets/2963569/24930055/cf7723b6-1ecd-11e7-8d1e-4da9f7ef5962.png)

miketruty:
My patch would be to adjust step2_exchange() where it raises FlowExchangeError(error_msg).  If error_msg == 'invalid_grant', then add some help text.

If I get to do this, I'd probably also adjust the help text 'No code was supplied in the query parameters.' to be a little clearer. code itself is ambiguous.

houglum:
We've actually got a bug open to implement a separate auth flow that doesn't require manual copy/paste, assuming the user is in an environment where we can spawn a web browser for them (the gcloud CLI tool does something similar, so I should be able to re-use their logic).  I'll keep this issue open to track this effort externally :)

houglum:
Fixed the error message in https://github.com/GoogleCloudPlatform/gsutil/pull/425

PR #425, opened 2017-04-13
Add helpful error message when 'invalid_grant'. #423

miketruty:
OK, I squashed it and here is the change.
https://codereview.appspot.com/316490043/


houglum:
Merged. Thanks!

Issue #448, opened 2017-08-06
Topic/event flag in `gsutil help notification` example


cwbeitel:
Maybe 

```bash
    gsutil notification create -f json -t OBJECT_FINALIZE gs://example-bucket
```

was meant to be

```bash
    gsutil notification create -f json -t [some topic] -e OBJECT_FINALIZE gs://example-bucket
```

?

PR #449, opened 2017-08-06
Fixed `gsutil notification create` example

cwbeitel:
https://github.com/GoogleCloudPlatform/gsutil/issues/448


houglum:
Going through old PRs -- looks like this has since been fixed:
https://github.com/GoogleCloudPlatform/gsutil/blob/c96d9dd34dad7f58f076a20f3d3da73c2f2de31b/gslib/commands/notification.py#L169

Marking as closed.

PR #458, opened 2017-09-07
Update gsutil signurl command to use the v4 signature format.

fishjord:


houglum:
This failed TravisCI because of #459.

Just submitted 1ae519c to take care of #459... and a couple more to fix resulting errors - can you pull those into your fork so the tests get re-run?

fishjord:
Done

PR #482, opened 2017-10-29
Adds charset to Content-Type metadata field set through magic file detection

erichmachado:
Related issue: https://codereview.appspot.com/335860043/


erichmachado:
I signed it!

PR #494, opened 2018-01-09
Rename COPYING to LICENSE

michaelawyu:


tswast:
@michaelawyu I believe this repository may be a mirror of an internal (google3) code directory. You may need to update it there too / instead.

michaelawyu:
Sure :)

houglum:
While other licenses allow "COPYING" or "LICENSE", it seems that the Apache license requires the license file to be named "LICENSE" (as mentioned at http://www.apache.org/dev/apply-license.html).  Given this, I agree with the suggested change.

That being said, we have code that expects "COPYING" to be present when doing updates, etc.

Could you also change "COPYING" to "LICENSE" in these locations?

* https://github.com/GoogleCloudPlatform/gsutil/blob/114fc88c47657ab9b1669be7a82dfa847f2a856f/MANIFEST.in#L4
* (and also please put it in the appropriate alphabetic spot) https://github.com/GoogleCloudPlatform/gsutil/blob/114fc88c47657ab9b1669be7a82dfa847f2a856f/gslib/tests/test_update.py#L81

houglum:
And to reply to the comment by @tswast: This repository is the source of truth for gsutil code.  Any other repositories are mirroring this code, not vice versa :)

michaelawyu:
Updated the PR as requested :)

PR #517, opened 2018-04-11
Adds validation to signurl valid duration value.

chestercun:


haizaar:
Can you please clarity the reasons behind setting the _MAX_EXPIRATION to 7 days? I couldn't find any matching limitation in the API [documentation](https://cloud.google.com/storage/docs/access-control/signed-urls) and manually tweaking the value for signing URL for 10 years worked (though time will tell).

Here is why I prefer to have longer expiration days: We use Cloud Storage to host different version of Chrome for use in CI. I prefer not to have them publicly shared hence use signed URLs with looong expiration dates.

houglum:
+@chestercun for comment, in case I'm missing anything.

This PR was a follow-up from where we changed the `signurl` command to use V4-style signing, similar to [AWS's v4 style](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) -- those docs specify a max of 604800 seconds.  The GCS docs you linked are for V2-style signing; it looks like we need to update our docs to cover the V4 stuff as well. +@starsandskies, FYI. 

I think you *can* actually create the signed URL if you specify a longer expiration, but it won't actually work after 7 days' time.  At least, this was the case when someone complained about this about a year ago (bug ID with details, for Google-internal folks: 77830441), which prompted this PR.

haizaar:
Thanks for the quick response and the details. Looks like I was too hasty on conclusions - I was able to generate signed URL with long expiration, but it didn't actually work resulting in Bad Request.

P.S. Do you know any way to generate signed URLs with >7d expiration times? (Any GCP-specific API, etc.?)

houglum:
V2-style signing still works -- you could either download an older version of gsutil to create those, or use one of the GCS client libraries ([permalink to the relevant part of our Python client lib](https://github.com/googleapis/google-cloud-python/blob/eccea4a262ee7b29942fd8273d3cca7447d06ee9/storage/google/cloud/storage/_signing.py#L112)).

haizaar:
Crafted a tool to make it easier: https://github.com/haizaar/gcs-signurl. Thanks again for the API link!
Signed a URL for 10 years. So far it works, but time will tell :)

```
$ pip install --user gcs-signurl
$ gcs-signurl --help
Usage: gcs-signurl [OPTIONS] KEY_FILE RESOURCE

  Generate a signed URL that embeds authentication data so the URL can be
  used by someone who does not have a Google account.

  This tool exists to overcome a shortcomming of gsutil signurl that limits
  expiration to 7 days only.

  KEY_FILE should be a path to a JSON file containing service account
  private key. See gsutil signurl --help for details

  RESOURCE is a GCS location in the form <bucket>/<path> (don't add neither
  "gs://" nor "http://...")

  Example: gcs-signurl /tmp/creds.json /foo-bucket/bar-file.txt

Options:
  -d, --duration TEXT  Specifies the duration that the signed url should be
                       valid for.
                       Times may be specified with no suffix
                       (default hours), or
                       with s = seconds, m = minutes, h =
                       hours, d = days.  [default: 1h]
  --help               Show this message and exit.
```

PR #545, opened 2018-07-16
Faint starts to Python 2-3 compatibility

edhodapp:
Fixed up print and except statements. Fixed up some imports with six.moves.

### DIFF
--- a/gslib/boto_resumable_upload.py
+++ b/gslib/boto_resumable_upload.py
@@ -44,12 +44,12 @@ TODO: gsutil-beta: Add a similar comment to the boto code.
 from __future__ import absolute_import
 
 import errno
-import httplib
+from six.moves import http_client
 import random
 import re
 import socket
 import time
-import urlparse
+from six.moves import urllib
 from boto import UserAgent
 from boto.connection import AWSAuthConnection
 from boto.exception import ResumableTransferDisposition
@@ -64,7 +64,7 @@ class BotoResumableUpload(object):
   """Upload helper class for resumable uploads via boto."""
 
   BUFFER_SIZE = 8192
-  RETRYABLE_EXCEPTIONS = (httplib.HTTPException, IOError, socket.error,
+  RETRYABLE_EXCEPTIONS = (http_client.HTTPException, IOError, socket.error,
                           socket.gaierror)
 
   # (start, end) response indicating service has nothing (upload protocol uses
@@ -108,7 +108,7 @@ class BotoResumableUpload(object):
 
     Raises InvalidUrlError if URL is syntactically invalid.
     """
-    parse_result = urlparse.urlparse(url)
+    parse_result = urllib.parse.urlparse(url)
     if (parse_result.scheme.lower() not in ['http', 'https'] or
         not parse_result.netloc):
       raise InvalidUrlError('Invalid upload URL (%s)' % url)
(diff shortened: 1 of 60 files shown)


edhodapp:
I signed it!

gslib/utils/boto_util.py
```diff
@@ -198,11 +205,11 @@ def GetGcsJsonApiVersion():
 # in multiples of 256KiB). Overridable for testing.
 def GetJsonResumableChunkSize():
   chunk_size = config.getint('GSUtil', 'json_resumable_chunk_size',
-                             1024*1024*100L)
+                             long(1024*1024*100))
```
rooterkyberian: AFAIK long is not even needed here (or in fact anywhere else)
1024*1024*100L is below 2147483647, the upper limit of int on 32bit python 2
also
https://www.python.org/dev/peps/pep-0237/
means that if it exceeds the limit it will be changed to long automatically

gslib/utils/unit_util.py
```diff
@@ -66,7 +73,7 @@ def _GenerateSuffixRegex():
 EIGHT_MIB = 8 * ONE_MIB
 TEN_MIB = 10 * ONE_MIB
 
-SECONDS_PER_DAY = 60 * 60 * 24L
+SECONDS_PER_DAY = long(60 * 60 * 24)
```
rooterkyberian: why would `long` be even needed here?

gslib/boto_resumable_upload.py
```diff
@@ -44,12 +44,12 @@
 from __future__ import absolute_import
 
 import errno
-import httplib
+from six.moves import http_client
```
rooterkyberian: please sort imports

rooterkyberian:
Nice someone actually tries to port it!
I think its crucial to include a change to `.travis.yml` in this PR:
```
python:
  - "2.7"
  - "3.6"
  - "pypy"
```
seeing these `iteritems` calls in the code makes me think it is possible that it won't pass on python 3 first time around (assuming the test coverage is good enough)

less crucial, but important for visibility is adding Python 3 classifiers to `setup.py`
```python
    classifiers=[
        'Development Status :: 5 - Production/Stable',
        'Environment :: Console',
        'Intended Audience :: Developers',
        'Intended Audience :: System Administrators',
        'License :: OSI Approved :: Apache Software License',
        'Natural Language :: English',
        'Programming Language :: Python',
        'Programming Language :: Python :: 2',
        'Programming Language :: Python :: 2.6',
        'Programming Language :: Python :: 2.7',
        'Programming Language :: Python :: 3',
        'Programming Language :: Python :: 3.6',
        'Topic :: System :: Filesystems',
        'Topic :: Utilities',
]
```

gslib/utils/unit_util.py
houglum: Not sure off the top of my head.  But, that's outside the scope of this changelist -- this is performing syntax conversions for Py 3 compatibility only :)

gslib/utils/boto_util.py
houglum: We can probably make a TODO to drop `long` completely once Python 2 reaches end of life and this only runs under Python 3 (https://pythonclock.org/).  In the interim, I'd rather not go changing things without a good reason.

houglum:
@edhodapp, can you change the title of this PR? It makes it seem like this will introduce Py 3 compatibility once merged, when (IIUC) it's just a step toward that.

PR #559, opened 2018-09-21
update references to Mac OS X

ToddKopriva:
Apple rebranded Mac OS X as macOS. When referring to the operating system in general, we use macOS now.
https://developers.google.com/style/word-list#letter-M

This change is being made throughout our documents.

### DIFF
--- a/gslib/addlhelp/crc32c.py
+++ b/gslib/addlhelp/crc32c.py
@@ -36,7 +36,7 @@ _DETAILED_HELP_TEXT = ("""
   The crcmod module contains a pure-Python implementation of CRC32C, but using
   it results in very poor performance. A Python C extension is also provided by
   crcmod, which requires compiling into a binary module for use. gsutil ships
-  with a precompiled crcmod C extension for Mac OS X; for other platforms, see
+  with a precompiled crcmod C extension for macOS; for other platforms, see
   the installation instructions below.
 
   At the end of each copy operation, the gsutil cp and rsync commands validate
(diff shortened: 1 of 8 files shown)


houglum:
Thanks for the PR! There are a few references that this PR didn't catch; would you mind updating it in the other files as well? They look to all be under the `gslib/` directory.

```
$ cd <root of gsutil repository>
$ grep -ir --exclude-dir=third_party/ "mac[ ]*os" .
./gslib/utils/parallelism_framework_util.py:On MacOS, you may also need to run a command like this once (in addition to the
./gslib/addlhelp/throttling.py:  various tools available on Linux and MacOS that can be used to throttle
./gslib/addlhelp/metadata.py:  to text/plain. If you're running gsutil on Linux or MacOS and would prefer to
./gslib/addlhelp/crc32c.py:  with a precompiled crcmod C extension for Mac OS X; for other platforms, see
./gslib/addlhelp/crc32c.py:  Mac OS X
./gslib/addlhelp/encoding.py:  locale configuration on Linux and Mac OSes. If your file paths are Unicode
./gslib/addlhelp/encoding.py:  using cygwin (on Windows) or Linux or MacOS - all of which support Unicode.
./gslib/addlhelp/encoding.py:<B>USING UNICODE FILENAMES ON MAC OS</B>
./gslib/addlhelp/encoding.py:  MacOS stores filenames in decomposed form (also known as
./gslib/addlhelp/encoding.py:  form (like Ubuntu) from one that does (like MacOS).
./gslib/addlhelp/encoding.py:  a MacOS machine and edit the file, then when you use gsutil rsync to bring
./gslib/addlhelp/encoding.py:  of overwriting the original. This is because MacOS converted the filename to
./gslib/addlhelp/encoding.py:    and MacOS are not. Thus, for example, if you have two filenames on Linux
./gslib/addlhelp/encoding.py:  - Mac OS performs character encoding decomposition based on tables stored in
./gslib/addlhelp/encoding.py:  downloading from Google Cloud Storage to a machine running MacOS).
./gslib/tests/test_cp.py:  # skipped, but it turns out os.walk() on MacOS doesn't have problems with
./gslib/commands/cp.py:  ends with a "/" is not allowed on Linux and MacOS. Because of this, it is
./gslib/commands/cp.py:  (e.g., it comes pre-installed on MacOS), but in other cases some users have
./gslib/commands/cp.py:  temp files by setting the TMPDIR environment variable. On Linux and MacOS
./gslib/commands/cp.py:  is not necessary after running the export command on Linux and MacOS.)
./gslib/commands/config.py:# MacOS and Windows users should see
./gslib/commands/config.py:# mechanism. Available on UNIX and MacOS (and possibly on Windows, if you're
```

ToddKopriva:
This change is related to this bug record:
https://issuetracker.google.com/issues/116161063

ToddKopriva:
I just added commits for changes in additional files, based on Matt Houglum's review.

PR #570, opened 2018-10-11
Fix bug in gsutil -m note outputting first arg instead of command name.

jterrace:
Tested locally:

```
$ ./gsutil -o GSUtil:x=y cp gs://pub/** /tmp/pub/
==> NOTE: You are performing a sequence of gsutil operations that may
run significantly faster if you instead use gsutil -m cp ... Please
see the -m section under "gsutil help options" for further information
about when gsutil -m can be advantageous.
```


houglum:
LGTM, merged.

PR #582, opened 2018-11-15
Fix link to storage 'locations' page (2)

garyleesmith3:
Former link:
https://cloud.google.com/storage/docs/bucket-locations
New link
https://cloud.google.com/storage/docs/locations


houglum:
Will merge this along with #581 when the link is live.

Issue #524, opened 2018-05-02
Add option for rsync not overwrite newer files


robert-kiessling:
Currently gsutil rsync performs a force one-directional sync which will always overwrite the destination file.

This feature request is to add an option `--preserve-newer` that changes the behaviour to not copy a file if both source and destination have mtimes and the destination mtime is newer than the source mtime.

In my case I want to use gsutil rsync to synchronise between two directories/buckets where changes can happen in both places. While a full-blown bidirectional sync would be difficult to implement, much of this (except deletions) could be achieved with `gsutil rsync -r --preserve-newer /some/path /some/other/path; gsutil rsync -r --preserve-newer /some/other/path /some/path`  .

PR #583, opened 2018-11-21
Add -u option to rsync, Skip files that are newer on the receiver.

m5-sugi:
please marge

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -475,7 +475,9 @@ _DETAILED_HELP_TEXT = ("""
                  only the top-level directory in the source and destination URLs
                  match, skipping any sub-directories.
 
-  -u             Skip files that are newer on the receiver.
+  -u             When a file/object is present in both the source and
+                 destination, if mtime is available for both, do not perform
+                 the copy if the destination mtime is newer.
                  
   -U             Skip objects with unsupported object types instead of failing.
                  Unsupported object types are Amazon S3 Objects in the GLACIER
@@ -1055,7 +1057,8 @@ class _DiffIterator(object):
     use_hashes = (self.compute_file_checksums or
                   (StorageUrlFromString(src_url_str).IsCloudUrl() and
                    StorageUrlFromString(dst_url_str).IsCloudUrl()))
-    if self.skip_old_files and has_src_mtime and has_dst_mtime and src_mtime < dst_mtime:
+    if (self.skip_old_files and has_src_mtime and has_dst_mtime and
+            src_mtime < dst_mtime):
       return False, has_src_mtime, has_dst_mtime
     if not use_hashes and has_src_mtime and has_dst_mtime:
       return (src_mtime != dst_mtime or src_size != dst_size, has_src_mtime,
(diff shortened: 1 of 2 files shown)


m5-sugi:
I signed it!

houglum:
Thanks for the pull request!  I don't have much context here... I assume this is intended to fix #524?

gslib/tests/test_rsync.py
```diff
@@ -2146,6 +2146,55 @@ def _Check():
...
+      self._VerifyObjectMtime(dst_bucket.bucket_name, 'obj1', '10')
+      self._VerifyObjectMtime(dst_bucket.bucket_name, 'obj2', '10')
+    _Check2()
+
+    self.CreateObject(bucket_uri=src_bucket2, object_name='obj3',
+                      contents='obj3-2', mtime=11)
+
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _Check3():
+      self.RunGsUtil(['rsync', '-u', suri(src_bucket2), suri(dst_bucket)], return_stderr=True)
```
houglum: There's no need for `return_stderr=True` here, since we're not using the returned value.

gslib/tests/test_rsync.py
```diff
@@ -2146,6 +2146,55 @@ def _Check():
...
+      self.RunGsUtil(['rsync', '-u', suri(src_bucket2), suri(dst_bucket)])
+      self.assertEquals('obj1-1', self.RunGsUtil(
+          ['cat', suri(dst_bucket, 'obj1')], return_stdout=True))
+      self.assertEquals('obj2-1', self.RunGsUtil(
+          ['cat', suri(dst_bucket, 'obj2')], return_stdout=True))
+      self._VerifyObjectMtime(dst_bucket.bucket_name, 'obj1', '10')
+      self._VerifyObjectMtime(dst_bucket.bucket_name, 'obj2', '10')
+    _Check2()
+
+    self.CreateObject(bucket_uri=src_bucket2, object_name='obj3',
```
houglum: Why is another iteration of rsync-and-check needed here? It seems like you could add this object creation to the two above (i.e. under line 2174), and include the assertions below into `_Check2()`.

Reasoning: The integration tests already take long enough to run, and an extra call here and there adds up.

m5-sugi:
Yes, this is intended to fix #524

gslib/commands/rsync.py
```diff
@@ -1052,6 +1055,8 @@ def _CompareObjects(self, src_url_str, src_size, src_mtime, src_crc32c,
     use_hashes = (self.compute_file_checksums or
                   (StorageUrlFromString(src_url_str).IsCloudUrl() and
                    StorageUrlFromString(dst_url_str).IsCloudUrl()))
+    if self.skip_old_files and has_src_mtime and has_dst_mtime and src_mtime < dst_mtime:
```
houglum: This is predicate is longer than 80 characters; please wrap it in parentheses:

        if (self.skip_old_files and has_src_mtime and has_dst_mtime and
            src_mtime < dst_mtime):
          return False, has_src_mtime, has_dst_mtime

m5-sugi:
Thanks for verifying!
I made some changes. Would you be able to check them?

PR #597, opened 2019-01-09
Black Lint

walkerjoe:
Using black package to lint all files - verified that tests can run
under both 2.7 and 3.6 after lint. (still some test failures)


walkerjoe:
Please check again

PR #600, opened 2019-01-09
Exception fixes, boto fixes

catleeball:


catleeball:
Updated google address to primary

PR #605, opened 2019-01-16
Fix #591: Don't clobber local files with differing size

### LINKED ISSUE #591: gsutil cp -n clobbers local files with differing size
I expected `gsutil cp -n` not to overwrite existing files when copying from the cloud to the local machine.  That expectation is based on the `--help` output, which describes `-n` as "No-clobber. When specified, existing files or objects at the destination will not be overwritten. [...]"  In practice, if the existing local file does not have the same size as the file in the cloud it will be overwritten.  For example:

```sh
echo 'Original content.' > test.txt
gsutil cp test.txt "gs://$BUCKET_NAME"
echo 'Modified content with differing size.' >| test.txt
gsutil cp -n "gs://$BUCKET_NAME/test.txt" test.txt
cat test.txt
```

Which prints `Original content.` instead of `Modified content with differing size.`, demonstrating that the file has been overwritten by `gsutil cp -n`.

The code responsible is at [copy_helper.py:3438](https://github.com/GoogleCloudPlatform/gsutil/blob/v4.35/gslib/utils/copy_helper.py#L3438) which was first introduced in a768a58 to support resuming partial downloads.  However, I don't think that has worked since 3e3d6dc, which started using temporary files (named with `_.gstmp`) for all downloads.

Would it be possible to either make `gsutil cp -n` behave as documented (my preference) or update the documentation to match the behavior?

Thanks,
Kevin

kevinoid:
As discussed in #591, `gsutil cp -n` will overwrite an existing local file if its size differs from the size of the source file.  This behavior is contrary to user expectations, POSIX convention, and the documentation.

Remove the special case for files with differing size, which was added in a768a58 to support resuming partial downloads.  This code has not been useful since 3e3d6dc, which started using temporary files (named with `_.gstmp`) for all downloads.

Fixes: #591
CC: @houglum
CC: @thobrla

Thanks for considering,
Kevin

P.S.:  To get any of the integration tests to run I had to apply boto/boto#3843.  I'm not sure why this apparently isn't affecting others.  Feel free to comment on that issue if you have any insights.

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -466,6 +466,20 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
       self.assertIn('Skipping existing item: %s' % suri(f), stderr)
       self.assertEqual(f.read(), 'bar')
 
+  @SequentialAndParallelTransfer
+  def test_noclobber_different_size(self):
+    key_uri = self.CreateObject(contents='foo')
+    fpath = self.CreateTempFile(contents='quux')
+    stderr = self.RunGsUtil(['cp', '-n', fpath, suri(key_uri)],
+                            return_stderr=True)
+    self.assertIn('Skipping existing item: %s' % suri(key_uri), stderr)
+    self.assertEqual(key_uri.get_contents_as_string(), 'foo')
+    stderr = self.RunGsUtil(['cp', '-n', suri(key_uri), fpath],
+                            return_stderr=True)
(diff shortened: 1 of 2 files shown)


kevinoid:
My pleasure.  Thanks for reviewing and merging it!

houglum:
I also added some notes on that Boto PR about the issues you were seeing with OpenSSL v1.1.1.  Thanks for mentioning those as well; I'll keep those on our radar.

PR #608, opened 2019-01-25
Updates helptext for bucketpolicyonly command.

houglum:
The old text was never changed after the command was revised to have two
subcommands. These changes make the explanations correct, as well as
isolating helptext between `get` and `set` so that `gsutil help
bucketpolicyonly get` and `gsutil help bucketpolicyonly set` will return
only the appropriate segments of the helptext.


houglum:
@chestercun PTAL, just a quick docs change after the merge today.

chestercun:
LGTM

PR #612, opened 2019-02-06
Update crcmod install instructions for CentOS 6

houglum:
CentOS 6 uses Python 2.6 by default (at least, it does on our GCE images).  This updates the instructions to show steps necessary to enable Python 2.7 for a session, as well as compiling/installing crcmod for Python 2.7.

This addresses https://issuetracker.google.com/issues/63114773


houglum:
@starsandskies PTAL

PR #616, opened 2019-02-12
Start to max diff cball

catleeball:
Cherry picked commits 6be1834 through fc0ba95 from catleeball/py-six-current, plus a merge commit.


catleeball:
Oopsies wrong target branch, ignore this

PR #619, opened 2019-02-13
Compatibility fixes for StringIO imports, exception syntax, hashlib encoding and tests updates, and verbose test comparisons

catleeball:
This is a cherry pick from [6be1834](https://github.com/GoogleCloudPlatform/gsutil/commit/6be18341c4acee3fbbc13f078e9104129836a409) through [db1e453](https://github.com/catleeball/gsutil/commit/db1e4539ae84a179ab38f4b4fbdda53a4984b297)

It includes Python 2.7 / 3.x compatibility fixes for:
- StringIO imports now from six.moves
- exception syntax updated to use `as`
- Hashlib string encoding updated
  - Tests updated since changing the encoding also changes the hash
- More verbose asserts in test failure output when comparing hashes


gslib/tests/signurl_signatures.py
```diff
@@ -0,0 +1,79 @@
...
+                         '0b0bbe0f300776d0&x-goog-algorithm=GOOG4-RSA-SHA256&x-goog-cred'
+                         'ential=test%40developer.gserviceaccount.com%2F19000101%2Fus-ea'
+                         'st%2Fstorage%2Fgoog4_request&x-goog-date=19000101T000555Z&x-go'
+                         'og-expires=3600&x-goog-signedheaders=host%3Bx-goog-resumable')
+
+TEST_SIGN_PUT_SIG_PY3 = "https://storage.googleapis.com/test/test.txt?x-goog-signature=b'2a045db18685eae0817a536d2406f544447959fa9bfac6461bdca2ae3e8e05e46f4516c6c89ecb7e9874ef02718432d2e7c852ef9a5d928f93dc76d96f9104287c053f4074d55d0be58478b658cff7ff256f02f9b97ed407f6a2d2a1b55adfb67ddbf81379922fdcaa5a3bfcee0a227bf5d40155ad4ff668940746a38683e09f268ad5f02e75da1a1ba02421ff17bcfc20089bbff05ef10da5281bbcffc096686bfeaad8639326e44ca78c369b00544286e303e3f91f3141c4b3b58f87dfcdde7e0bb62c3a69ba807758e1ddbb996b2435fd3776276cd0e18b7b29e2f6b2695b751131d76e021eb9f80a125e2a3ba7c77edc9b7aea655f76122bb75ae10499db'&x-goog-algorithm=GOOG4-RSA-SHA256&x-goog-credential=test%40developer.gserviceaccount.com%2F19000101%2Fus-east%2Fstorage%2Fgoog4_request&x-goog-date=19000101T000555Z&x-goog-expires=3600&x-goog-signedheaders=host%3Bx-goog-resumable"
```
houglum: Here and for all the PY3 alternatives below, I don't think the signature should contain the PY3 artifact of ` b' `. And unless I'm missing something, the signature should be the same between PY2 an PY3, right?

catleeball: This change is due to PyOpenSsl having different implementations between Python 2.x and 3.x, and different method signatures for their sign methods, hence the change [here](https://github.com/GoogleCloudPlatform/gsutil/pull/619/files/19fe4b93543d1423f10792c96de2edc4a5a06db9#diff-2203ca5e293c09b69dae32ab51ee067aR298) and [here](https://github.com/GoogleCloudPlatform/gsutil/pull/619/files/19fe4b93543d1423f10792c96de2edc4a5a06db9#diff-2203ca5e293c09b69dae32ab51ee067aR281). After making that change, I was able to generate signatures across both versions, but the signatures differed, including these artifacts.

testSignPut and related tests were able to succeed with these values, and despite the artifact showing up in the URL, it's still a valid URL. The values were taken from running the integ tests in 3.5 and grabbing the values from the actual output of the asserts, after having made changes to work with both versions of PyOpenSsl.

That said, I think you're right that it *shouldn't* act like this, but it seems like a side effect of the differing implementations of the PyOpenSsl lib. We could manually strip off the enclosing `b' '` if you'd like though.

houglum: This ends up happening because the string_to_sign *also* contains the `b'...'` wrapper:

Using a Py3 interpreter (taken from the py-six-current branch):
(Pdb) string_to_sign
"GOOG4-RSA-SHA256\n20190213T074138Z\n20190213/US/storage/goog4_request\nb'e7f91afd09b584e12e0a9b96d4f2c1f150edb7cd2b6b8563b646e5a17f9b77bc'"

Using a Py2 interpreter:
(Pdb) string_to_sign
u'GOOG4-RSA-SHA256\n20190213T074138Z\n20190213/US/storage/goog4_request\ne7f91afd09b584e12e0a9b96d4f2c1f150edb7cd2b6b8563b646e5a17f9b77bc'

The string to sign should have those removed before actually performing the signature.  If you manually replace that string with the "fixed" one, the signature comes out correctly. Plz fix :)

catleeball: Oh!! Thanks for noticing that, I'll apply the fix. Thank you, will fix. 👍

catleeball: Okay, I've added a little commit fixing the artifact issue. Thanks again for the insight around the issue. 😃

houglum: FYI for others following this thread: The issue here was with plugging `byte`-types into unicode format strings, ending up with unicode strings that look like `argument_is:b'MyString'` instead of `argument_is:MyString`.

Also, the integration tests aren't catching this stuff.  We should have unit tests that ensure we always get the same resulting signature when using a given set of values for all the variables that go into creating a signature (same datetime, bucket/object name, private key, etc.).

catleeball: Small note: it looks like `hashed_canonical_request` is also getting wrapped in `b'...'`. Going to debug `_GenSignedKey` to try and find if there's any other instances of this happening elsewhere too

catleeball: It looks like this happens after doing `canonical_request = canonical_request.encode()`. When trying things like `canonical_request = canonical_request.encode()[2:-1]` in attempt to strip the `b' '`, it just strips characters inside the `b' '`. Trying to search more info on why this happens.

catleeball: Oh, here's something from Stack Overflow about this: https://stackoverflow.com/questions/21810776/str-encode-adds-a-b-to-the-front-of-data

Adding `.decode()` to the end of `hashed_canonical_request = base64.b16encode(canonical_request_hasher.digest()).lower()` removed the wrapping `b' '`

catleeball: Okay, I've added a couple `.decodes()` and set tests to point back to the same old hashes and removed the Py3 specific hashes. :)

gslib/tests/test_signurl.py
```diff
@@ -22,6 +22,7 @@
 from datetime import datetime
 from datetime import timedelta
 import pkgutil
+from six import PY2, PY3
```
houglum: For consistency's sake (with the rest of the code base), can we just import `six` and reference these below as `six.PY2` and `six.PY3`?

catleeball: Sounds good!

catleeball: After removing the extra hashes, PY2 and PY3 weren't used at all in this file anymore, so I just removed the import entirely.

PR #620, opened 2019-02-14
Adding the python executable path

catleeball:
Commit cherry-picked: 779903967a1e2964a004665e4a3f7f285a62de25 by @walkerjoe 

The full python executable path was not propagating through to the test
processes and it was defaulting to Python2.7 even when the test was
started with python 3.x. Adding the path to the beginning of the cmd
sent to Popen forces the tests to use the correct python version.

### DIFF
--- a/gslib/commands/test.py
+++ b/gslib/commands/test.py
@@ -292,7 +292,7 @@ def CreateTestProcesses(parallel_tests, test_index, process_list, process_done,
     Index of last created test.
   """
   orig_test_index = test_index
-  executable_prefix = [sys.executable] if sys.executable and IS_WINDOWS else []
+  executable_prefix = [sys.executable] if sys.executable else []
   s3_argument = ['-s'] if tests.util.RUN_S3_TESTS else []
   multiregional_buckets = ['-b'] if tests.util.USE_MULTIREGIONAL_BUCKETS else []
   project_id_arg = []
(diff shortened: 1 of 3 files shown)


catleeball:
LGTM :ship: 

Any idea why this might have been exclusively Windows before?

walkerjoe:
it has to do with the keywords being registered in terminal vs cmd... unix has python registered but windows usually doesn't so it requires the full path.

houglum:
This was added by @jterrace in https://github.com/GoogleCloudPlatform/gsutil/commit/28e356a7ec2398eba148c6178a2eab2b12463c1d.  That commit message isn't particularly detailed, so I'd like to get his input as to why it was added before removing it again.

gslib/commands/test.py
```diff
@@ -292,7 +292,7 @@ def CreateTestProcesses(parallel_tests, test_index, process_list, process_done,
     Index of last created test.
   """
   orig_test_index = test_index
-  executable_prefix = [sys.executable] if sys.executable and IS_WINDOWS else []
+  executable_prefix = [sys.executable] if sys.executable else []
```
houglum: I think this was originally added with the intention of correctly running from a Python archive file, which bundles the interpreter and the code together. Context, if you're interested: https://github.com/google/subpar.

Can you add a bit to this that checks if we're running in an archive, and if so, doesn't add `sys.executable`? Something like this should do:

```python
  def InvokedFromParFile():
    loader = globals().get('__loader__', None)
    if not loader:
      return False
    return 'zipimport' in loader.__class__.__module__

  executable_prefix = []
  if not InvokedFromParFile():
    executable_prefix.append(sys.executable)

  orig_test_index = test_index
  # ...
```

walkerjoe: Code in second commit address the changes requested to omit executable prefix when run from python archive files

PR #623, opened 2019-02-18
Refactoring, encoding fixes, and concat fixes

catleeball:
- 4faf087 Refactored; make get_random_ascii_chars method
> Made a new method get_random_ascii_chars as part of a small
> refactor. There was some minor redundant behavior. This method
> was made to follow DRY practices, and to increase readability.
 
- a194cc1 Unicode fix for Python 2
> One of the Unicode tests gets upset in Python 2.7 after recent
> commits. Explicitly specifying encoding makes it compatible
> across versions.

- 11e90c6 Unit test printer now safely concats strings
> Now using six.ensure_text to wrap test IDs before printing them
> in order to ensure compatibility between Python 2 and 3.

- 9921822 Refactored integ test for readability.
> While working with the integ test test_gzip_transport_encoded_upload_and_download
> I refactored it into several sub-methods, added docstrings, and did
> some general cleanup.

- 604712c Wrapping `cmd` list elements in six.ensure_text
> This is to prevent issues with unicode in the command text when
> running in Python 2.7.

- e81f693 Expanded docstrings for two tests.
> Docstrings were made more verbose, explaining two integ tests:
> - test_gzip_transport_encoded_parallel_upload_resumable
> - test_gzip_transport_encoded_parallel_upload_non_resumable
>
> This was based on feedback given in the following bug:
> - http://issuetracker.google.com/123027395

### DIFF
--- a/gslib/tests/testcase/integration_testcase.py
+++ b/gslib/tests/testcase/integration_testcase.py
@@ -276,25 +276,6 @@ class GsUtilIntegrationTestCase(base.GsUtilTestCase):
       self.xml_api.PatchObjectMetadata(bucket_name, object_name, obj_metadata,
                                        provider=provider)
 
-  def get_random_ascii_chars(self, size, seed=None):
-    """Generates random ASCII characters up to a given size.
-
-    Args:
-      size: Integer quantity of characters to generate.
-      seed: A seed may be specified for deterministic behavior.
-            None is used as the default value.
-
-    Returns:
-      String of length equal to size argument.
-    """
-    random.seed(seed)
-    charset = string.printable.encode('ascii')
-
-    contents = ''.join([random.choice(charset) for _ in range(size)]
-                        ).encode('ascii')
-
-    random.seed()  # Reset the seed for any other tests.
-    return contents
 
   def SetPOSIXMetadata(self, provider, bucket_name, object_name, atime=None,
                        mtime=None, uid=None, gid=None, mode=None):
--- a/gslib/tests/testcase/integration_testcase.py
+++ b/gslib/tests/testcase/integration_testcase.py
@@ -25,12 +25,15 @@ import datetime
 import locale
 import logging
 import os
+import random
+import string
 import subprocess
 import sys
 import tempfile
 import time
 
 import six
+from six.moves import range
 
 import boto
 from boto import config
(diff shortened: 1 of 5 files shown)


gslib/tests/testcase/integration_testcase.py
```diff
@@ -273,6 +276,26 @@ def _SetObjectCustomMetadataAttribute(self, provider, bucket_name,
...
+  def get_random_ascii_chars(self, size, seed=None):
+    """Generates random ASCII characters up to a given size.
+
+    Args:
+      size: Integer quantity of characters to generate.
+      seed: A seed may be specified for deterministic behavior.
+            None is used as the default value.
+
+    Returns:
+      String of length equal to size argument.
```
houglum: In Python 3, this isn't a "string" :) Please differentiate between unicode string and bytes/byte-sequences in your docstrings (maybe by using `six.text_type` and `bytes`/`six.binary_type`?)

gslib/tests/testcase/integration_testcase.py
```diff
@@ -273,6 +276,26 @@ def _SetObjectCustomMetadataAttribute(self, provider, bucket_name,
       self.xml_api.PatchObjectMetadata(bucket_name, object_name, obj_metadata,
                                        provider=provider)
 
+  def get_random_ascii_chars(self, size, seed=None):
```
houglum: [no action required]

This isn't specific to integration tests, and thus should probably go in the testing `util` module instead.  We should go through and reorganize things to be in their correct/appropriate modules at some point, but that's fairly low priority, so if it will make diffs and merge conflicts easier to deal with (in doing the rest of your cherry pick PRs), just leave it here :)

catleeball: That sounds good! I dropped it into `gsutil/utils/text_util.py` if that sounds like a reasonable new home for the function. :house_with_garden:

gslib/tests/testcase/integration_testcase.py
catleeball: Yes!! I updated the docstring to clarify explicitly that it's returning an ASCII encoded bytestring, with a note about Py2 / Py3 strings :spiral_notepad:

PR #624, opened 2019-02-18
Fix du for 6-digit sizes (#613)

### LINKED ISSUE #613: du summary is badly formatted on sizes with 6 digits
Minor issue, but when you have a path prefix that weight between 1000 and 1023.99 {…,G,k}iB, it uses all the available space _plus one character_:

```
872.78 KiB  gs://my-bucket/dir1
41.06 GiB   gs://my-bucket/dir2
1011.74 GiB  gs://my-bucket/dir3  # <-- here
203.22 MiB  gs://my-bucket/dir4
3.26 GiB    gs://my-bucket/dir5
```

It should probably use 11 chars rather than 10 for the size. I can make a PR ([following the guide](https://cloud.google.com/storage/docs/gsutil/addlhelp/ContributingCodetogsutil)) if you want.

bfontaine:
Fixes #613.

### DIFF
--- a/gslib/commands/du.py
+++ b/gslib/commands/du.py
@@ -141,7 +141,7 @@ class DuCommand(Command):
   def _PrintSummaryLine(self, num_bytes, name):
     size_string = (MakeHumanReadable(num_bytes)
                    if self.human_readable else str(num_bytes))
-    sys.stdout.write('%(size)-10s  %(name)s%(ending)s' % {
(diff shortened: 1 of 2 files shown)


catleeball:
Looks good to me! 👍 Adding Matt to also review though, since he has a little more context on this.

Thank you for submitting this PR!

PR #630, opened 2019-02-26
Getting Py-Six-Current up to date - PR02

walkerjoe:
Mostly small optimizations in this PR. (ex: using curly brackets to define sets instead of a list type-casted into a set, etc...)


gslib/boto_translation.py
```diff
@@ -405,8 +405,15 @@ def ListObjects(self, bucket_name, prefix=None, delimiter=None,
           key_to_convert = key
 
           # Listed keys are populated with these fields during bucket listing.
-          key_http_fields = set(['bucket', 'etag', 'name', 'updated',
-                                 'generation', 'metageneration', 'size'])
+          key_http_fields = {
```
houglum: I'd argue that this decreases readability.  When I see curly braces, I immediately jump to thinking it's a `dict`.   Also, there's no syntax to initialize an empty `set` like this, (`{}` would create a `dict`), and I'd rather be consistent with how we initialize `set`s across the codebase.

gslib/tests/test_json.py
```diff
@@ -0,0 +1,36 @@
...
+      cls.binary_json = str('[{"integer": 1,"decimal": 3.14,"text": '
+                          '"testing, 1, 2"}]')
+    cls.keys = {"integer", "decimal", "text"}
+    cls.json = [
+      {
+        "integer": 1,
+        "decimal": 3.14,
+        "text": "testing, 1, 2"
+      }
+    ]
```
houglum: Need a blank line before a method def'n in a class.

gslib/utils/json_six.py
```diff
@@ -0,0 +1,46 @@
...
+
+import json
+from six import PY3
+from six import binary_type
+
+# used for including both loads and dumps methods for ease of importing
+from json import dumps
+
+def loads(json_data):
+  """
```
houglum: This docstring isn't consistent with google py style.  Doc strings should look like:

```
def foo():
  """Summary line less than 80 chars.

  Additional text if necessary.
  """
```

gslib/utils/json_six.py
```diff
@@ -0,0 +1,46 @@
...
+  decode_message: Merge from a JSON string in to a message.
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import json
+from six import PY3
```
houglum: I'd like to be consistent through the codebase that we import `six`, and address the members in that module as `six.binary_type`, `six.PY3`, etc.

gslib/utils/posix_util.py
```diff
@@ -430,11 +430,11 @@ def ParseAndSetPOSIXAttributes(path, obj_metadata, is_rsync=False,
     if atime > NA_TIME and mtime > NA_TIME:
       # Set both atime and mtime.
       os.utime(path, (atime, mtime))
-    elif atime > NA_TIME and mtime <= NA_TIME:
+    elif atime > NA_TIME >= mtime:
```
houglum: I think this actually makes the predicate harder to understand - we should be wary of adding the three-way comparisons unless they're semantically checking "make sure that <<value in the middle>> is between the two values surrounding it".

houglum:
Most of my nits here are about the changes to the `set` init style.  I'd prefer not to use the "set display" style (curly brackets), as it carries the overhead for readers along the lines of "ah, curly brackets, is this a dict or a set?".   But I've also got a few other nits here and there.

gslib/boto_translation.py
walkerjoe: There are many places where sets are defined only with curley brackets. Should they all be changed to set([])?

gslib/tests/test_json.py
walkerjoe: Formatting change submitted

gslib/utils/json_six.py
walkerjoe: Docstring formatting changed changed and expanded to include more information

gslib/utils/json_six.py
walkerjoe: imports revised

gslib/utils/posix_util.py
walkerjoe: It can be reverted, but I made this change because it seemed to me the more readable way of displaying that information. Maybe that's just me though.

gslib/boto_translation.py
houglum: Discussed offline - in order to get caught up with local branches, I'm willing to pull the "set display" syntax in, but I'd like to change it back to use the explicit `set([...])` syntax once we've got this branch caught up.

gslib/utils/posix_util.py
houglum: This is probably fine --  looking at it again, using lt/gt operators to compare against an "NA" value is confusing enough in the first place. We should really go in an add helper methods or figure out a way to make this read a bit more cleanly, but it's fine for now.

gslib/tests/test_json.py
```diff
@@ -0,0 +1,31 @@
+from gslib.tests import testcase
+from gslib.utils import json_six
+import six
+
+class TestJsonConversion(testcase.GsUtilUnitTestCase):
+  @classmethod
+  def setUpClass(cls):
+    cls.text_json = six.ensure_text('[{"integer": 1,"decimal": 3.14,"text": '
+                        '"testing, 1, 2"}]')
```
houglum: W.r.t. alignment of args that you can't fit all on the same line, either use the "Aligned with opening delimiter" style, or the " 4-space hanging indent; nothing on first line" style.  If you go with a 4-space indent, don't put any args on the same line as the method name:

http://google.github.io/styleguide/pyguide.html#34-indentation

gslib/tests/test_json.py
```diff
@@ -0,0 +1,31 @@
+from gslib.tests import testcase
```
houglum: Q: Any reason this file isn't named `test_<name-of-module>.py`?  I'd expect this to be named `test_json_six.py` if you're testing the `json_six` module.

gslib/boto_translation.py
walkerjoe: As per chat, these changes will be re-addressed at a later time.

gslib/tests/test_json.py
walkerjoe: Yes, that's probably what it should be named. But now that I think about it, I'm pretty sure we can actually delete this test and the json_six file. I think my original intention was to use it throughout the program, but with the six.ensure methods, I think this may be irrelevant.

gslib/tests/test_json.py
walkerjoe: Yep, you're right, formatting changed in that file.

houglum:
LGTM - one nit though, can you remove the submodule changes?  I'm assuming those are local changes, rather than actual version bumps, right? (In which case we should not push those.)

walkerjoe:
recreating without third_party changes

PR #631, opened 2019-02-26
Fixes #606 - use SNI for boto cxns if possible

### LINKED ISSUE #606: On Python installs built against OpenSSL 1.1.1, Boto fails with SNI errors
As mentioned in #605 and https://github.com/boto/boto/pull/3843, Boto will fail to connect to Google servers if the version of OpenSSL that Python is compiled against is 1.1.1:

```
  File "./gsbototest.py", line 6, in <module>
    boto.storage_uri('bucket-name', 'gs').create_bucket()
  File "/tmp/boto/boto/storage_uri.py", line 574, in create_bucket
    storage_class)
  File "/tmp/boto/boto/gs/connection.py", line 95, in create_bucket
    data=get_utf8_value(data))
  File "/tmp/boto/boto/s3/connection.py", line 682, in make_request
    retry_handler=retry_handler
  File "/tmp/boto/boto/connection.py", line 1074, in make_request
    retry_handler=retry_handler)
  File "/tmp/boto/boto/connection.py", line 1033, in _mexe
    raise ex
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:727)
```

Until that Boto pull request gets merge, a new version gets released (this is the part we don't have control over :( ), and we can start pinning to the new version of Boto in our setup.py code, we should probably add this patch to the Boto monkeypatches we run at startup.

Thanks again to @kevinoid for pointing this out.

houglum:
Boto https connections weren't using SNI before. With the recent-ish
introduction of OpenSSL 1.1.1, we start to see SSL errors when making
requests to storage.googleapis.com (GCS's XML API) where the server
hostname was not specified. This pulls in
https://github.com/boto/boto/pull/3843 as a hotfix into gsutil, since
there isn't a released version of Boto where this PR is included yet,
and we'd like to have it sooner rather than later.

### DIFF
--- a/gslib/utils/boto_util.py
+++ b/gslib/utils/boto_util.py
@@ -348,6 +348,8 @@ def JsonResumableChunkSizeDefined():
 def MonkeyPatchBoto():
   """Apply gsutil-specific patches to Boto.
 
+  Here be dragons. Sorry.
+
   Note that this method should not be used as a replacement for contributing
   fixes to the upstream Boto library. However, the Boto library has historically
   not been consistent about release cadence, so upstream fixes may not be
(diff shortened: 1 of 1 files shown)


gslib/utils/boto_util.py
```diff
@@ -387,6 +392,66 @@ def _PatchedGetPluginMethod(cls, requested_capability=None):
...
+        self.sock.ciphers = None
+    else:
+        self.sock = ssl.wrap_socket(sock, keyfile=self.key_file,
+                                    certfile=self.cert_file,
+                                    cert_reqs=ssl.CERT_REQUIRED,
+                                    ca_certs=self.ca_certs)
+    cert = self.sock.getpeercert()
+    hostname = self.host.split(':', 0)[0]
+    if not ValidateCertificateHostname(cert, hostname):
+        raise InvalidCertificateException(hostname,
```
catleeball: Nitpick: is there any point we would want to print the __str__ method of this error? The only extra attribute you don't have in your print string is InvalidCertificateException.reason, and I'm not sure what that would contain that you don't already list. Might be ok for living behind the debug flag.

houglum: Just copy/pasting what was added in the Boto PR.  This should suffice for most cases :)

PR #634, opened 2019-02-27
Getting Py-Six-Current up to date - PR04

walkerjoe:
* syntax changes from py2 to py3
* ascii/unicode conversions
* string formatting changes for clarity & readability


walkerjoe:
RE: CLA - Contributors in this PR are as follows:
* Joe Walker (walkerjoseph@)
* Lee Ball (cball@)
Both contributors have active CLA's and are cleared to contribute to this project

walkerjoe:
Verified that Travis CI check failures are false negatives. All tests passed.

gslib/tests/testcase/base.py
```diff
@@ -194,9 +195,10 @@ def CreateTempFile(self, tmpdir=None, contents=None, file_name=None,
     Returns:
       The path to the new temporary file.
     """
-    tmpdir = tmpdir or self.CreateTempDir()
-    file_name = file_name or self.MakeTempName('file')
-    if isinstance(file_name, six.text_type):
+
+    tmpdir = six.ensure_str(tmpdir or self.CreateTempDir())
+    file_name = file_name or self.MakeTempName(str('file'))
```
catleeball: Should we do `six.ensure_string('file')` here instead of `str('file')`? Since `str` can return either bytes or unicode depending on version, and a string literal should also imply`str`.

gslib/tests/testcase/base.py
```diff
@@ -205,7 +207,11 @@ def CreateTempFile(self, tmpdir=None, contents=None, file_name=None,
 
     with open(fpath, 'wb') as f:
       contents = (contents if contents is not None
-                  else self.MakeTempName('contents').encode(UTF8))
+                  else self.MakeTempName(str('contents')))
```
catleeball: Same as about about six.ensure_string

gslib/tests/testcase/integration_testcase.py
```diff
@@ -276,6 +276,26 @@ def _SetObjectCustomMetadataAttribute(self, provider, bucket_name,
       self.xml_api.PatchObjectMetadata(bucket_name, object_name, obj_metadata,
                                        provider=provider)
 
+  def get_random_ascii_chars(self, size, seed=None):
```
catleeball: This method can be removed. It was merged in PR #623 , but moved to utils/text_util.py

gslib/tests/util.py
```diff
@@ -623,14 +623,19 @@ def MakeBucketNameValid(name):
   """Returns a copy of the given name with any invalid characters replaced.
 
   Args:
-    name (str): The bucket name to transform into a valid name.
+    name: The bucket name to transform into a valid name.
```
catleeball: May want to specify if it will be a byte string, unicode string, or either for clarity. That way if someone contributes but isn't aware of the string handling differences between Py 2 / 3, they'll know

gslib/tests/testcase/base.py
walkerjoe: ensure_string should return the same as str(). The reason for using str instead of text/binary is that join() requires str() type and does not work with both text/binary types in all version of python.

gslib/tests/testcase/integration_testcase.py
walkerjoe: removed

gslib/tests/testcase/base.py
catleeball: Got it. Is it necessary to wrap with `str()` at all then? Since an inline string literal should evaluate to the interpreter's default string type.

gslib/tests/util.py
walkerjoe: Should it mirror the return types--(unicode, str, bytes)?

gslib/tests/testcase/base.py
walkerjoe: I understand that there will be different encodings when declared with str() and I try to avoid it when possible, but I believe it is the more universal situation in this instance.

gslib/tests/util.py
catleeball: Yeah, that makes sense to me. (str (py2: bytes / py3: unicode)) or however you want to notate it

gslib/tests/testcase/base.py
walkerjoe: The default is overridden with the top line `# -*- coding: utf-8 -*-` and can cause join to not work properly (at least that is the behavior I observed during my testing)

catleeball: Right, but `'foo' == str('foo')` is what I was meaning. The `str()` would be redundant. :slightly_smiling_face:

catleeball: Oooh I see now, re: the utf-8 string at the top

gslib/tests/util.py
walkerjoe: Added type hint to mirror return types

gslib/tests/util.py
```diff
@@ -623,14 +623,19 @@ def MakeBucketNameValid(name):
   """Returns a copy of the given name with any invalid characters replaced.
 
   Args:
-    name (str): The bucket name to transform into a valid name.
+    name (str, unicode, bytes): The bucket name to transform into a valid name.
 
   Returns:
-    (str) The version of the bucket name containing only valid characters.
+    (str, unicode, bytes) The version of the bucket name containing only
```
houglum: Nit, here and in the line above: Can you make these "(str, unicode, or bytes)" or "Union[str, unicode, bytes]"? The versions without the "or" sort of makes it look like a tuple.

gslib/tests/util.py
```diff
@@ -623,14 +623,19 @@ def MakeBucketNameValid(name):
...
+      valid characters.
   """
   # Neither underscores nor uppercase letters are valid characters for a
   # bucket name. Replace those with hyphens and lowercase characters.
-  return name.replace('_', '-').lower()
+  if isinstance(name, (six.text_type, six.binary_type)):
+    return name.replace('_', '-').lower()
+  else:
+    raise TypeError("Unable to format name. Incorrect Type: {0}".format(
```
houglum: Nit: single quotes instead of double quotes here.

gslib/tests/util.py
walkerjoe: Good point, I will change it.

gslib/tests/util.py
walkerjoe: Changed

PR #637, opened 2019-02-28
Getting Py-Six-Current up to date - PR05

walkerjoe:
This is a smaller one, but it's all I could get done before I catch my bus! More to follow tomorrow.

### DIFF
--- a/gslib/tests/testcase/base.py
+++ b/gslib/tests/testcase/base.py
@@ -201,6 +201,7 @@ class GsUtilTestCase(unittest.TestCase):
     if isinstance(file_name, (six.text_type, six.binary_type)):
       fpath = os.path.join(tmpdir, file_name)
     else:
+      file_name = map(six.ensure_str, file_name)
       fpath = os.path.join(tmpdir, *file_name)
     if not os.path.isdir(os.path.dirname(fpath)):
       os.makedirs(os.path.dirname(fpath))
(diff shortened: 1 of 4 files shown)


catleeball:
LGTM ship it :ship:

gslib/tests/testcase/base.py
```diff
@@ -201,6 +201,7 @@ def CreateTempFile(self, tmpdir=None, contents=None, file_name=None,
     if isinstance(file_name, (six.text_type, six.binary_type)):
       fpath = os.path.join(tmpdir, file_name)
     else:
+      file_name = map(six.ensure_str, file_name)
```
houglum: Per the style guide, try to use list/dict/tuple comprehensions instead of `map()`, `filter()`, etc.

http://google.github.io/styleguide/pyguide.html#215-deprecated-language-features

gslib/tests/testcase/integration_testcase.py
```diff
@@ -900,7 +906,7 @@ def RunGsUtil(self, cmd, return_status=False,
     status = p.returncode
 
     if expected_status is not None:
-      cmd = [six.ensure_text(item) for item in cmd]
+      map(six.ensure_text, cmd)
```
houglum: See other comment about not using `map()`.

gslib/tests/testcase/base.py
catleeball: It looks like the decision favors using `map()`: http://google.github.io/styleguide/pyguide.html#2152-decision

Joe and I were actually talking about this earlier. :slightly_smiling_face:

houglum: Ya know, I instinctively looked for that section in the style guide, read the first paragraph, and then linked it. Didn't bother to read the code snippets 🙃 They used to discourage these completely in favor of list comprehensions, but I guess it changed at some point. If the style guide doesn't complain, I won't either :)

PR #638, opened 2019-02-28
fix the daisy_chain_wrapper when file is bigger than _DEFAULT_DOWNLOAD_CHUNK_SIZE

darkloup:
The goal of the PR is to fix the file transfert between gs and aws. For me any copy with file bigger than 100Mo was hanging indefinitely.
When transfering file bigger than _DEFAULT_DOWNLOAD_CHUNK_SIZE, the second download for the checksum hangs because the stop_download threading event isn't reset correctly.

I was unable to use the upload.py specified in the contributing guidelines. But I did the first part(CLA) given the size/complexity of the contribution it didn't want to waste to much time on it.

### DIFF
--- a/gslib/daisy_chain_wrapper.py
+++ b/gslib/daisy_chain_wrapper.py
@@ -298,6 +298,7 @@ class DaisyChainWrapper(object):
           self.bytes_buffered = 0
           self.last_position = 0
           self.last_data = None
+        self.stop_download.clear()
         self.StartDownloadThread(start_byte=offset,
                                  progress_callback=self.progress_callback)
     else:


catleeball:
Thanks for the contribution @darkloup , this looks good to me! Adding @houglum who may have greater context on the daisy chaining method in case there's some spooky threading side effect to calling [clear](https://docs.python.org/2/library/threading.html#threading.Event.clear) that I'm unaware of

houglum:
Thanks for the PR!  I'd like to pull in your changes and run large cross-provider tests before approving and merging - I've just been focused on other tasks this past week.  I'll try to get to this tomorrow; feel free to @ me if I haven't replied in a couple days.

houglum:
I haven't been able to reproduce the issue, even with an object that is more than 2 multiples of the default chunk size.

Notes:
*  The commands below are run using a clone of gsutil from the current master branch, without the proposed fix.
*  I set the chunk size in `gslib/daisy_chain_wrapper.py` to be 2MiB, for faster testing transfers:
    `_DEFAULT_DOWNLOAD_CHUNK_SIZE = 1024 * 1024 * 2`

```shell
$ AWS_BKT="bucket-name-here"
$ GCS_BKT="bucket-name-here"
$ ./gsutil mb gs://$GCS_BKT
Creating gs://[bucket]/...
$ gsutil mb s3://$AWS_BKT
Creating s3://[bucket]/...

$ python2
>>> with open('/tmp/7MiB.txt', 'wb') as f:
...   f.write('a' * (3 * 1024 * 1024))
...   f.write('b' * (4 * 1024 * 1024))
... 
>>> exit()

# Copy from local disk to S3 works fine
$ ./gsutil -o 'GSUtil:prefer_api=xml' \
      cp /tmp/7MiB.txt s3://$AWS_BKT
Copying file:///tmp/7MiB.txt [Content-Type=text/plain]...
- [1 files][  7.0 MiB/  7.0 MiB]                                                
Operation completed over 1 objects/7.0 MiB.

# Copy from S3 to GCS works fine
$ ./gsutil -o 'GSUtil:prefer_api=xml' \
      cp s3://$AWS_BKT/7MiB.txt gs://$GCS_BKT
Copying s3://[bucket]/7MiB.txt [Content-Type=text/plain]...
\ [1 files][  7.0 MiB/  7.0 MiB]
Operation completed over 1 objects/7.0 MiB.

# Copy from GCS to S3 works fine
$ ./gsutil -o 'GSUtil:prefer_api=xml' \
      cp gs://$GCS_BKT/7MiB.txt s3://$AWS_BKT/7MiB_copied_back.txt
Copying gs://[bucket]/7MiB.txt [Content-Type=text/plain]...
- [1 files][  7.0 MiB/  7.0 MiB]
Operation completed over 1 objects/7.0 MiB.
```

I also tried this with changing the default chunk size back to 100 MiB and using a 350MiB object -- all the copies still succeeded.  **Note that while the progress bar seems to indicate that gsutil "hangs" when copying to S3, it's actually writing bytes under the hood -- our progress callbacks just don't seem to work for copies to S3; the bar jumps from 0% to 100% whenever a transfer completes... you can verify that bytes are being sent by using the top-level `-D` flag.** (I *think* there's a bug open for that somewhere?) Anyway, if that's the thing that was leading you to believe your transfer was hanging, try running with `-D` and checking whether the transfer is actually happening, despite the progress bar's lies.

I also tried all of these commands, but with the top-level `-m` flag, and they all succeeded as well.

Can you provide me with a more detailed reproduction for how you're running into this issue?  Maybe there's something special about the AWS region your bucket is in?  I'd also like to see which version of Python and which OS you're encountering this issue on.

Most of the info I'd like to see will be present in the output of `gsutil -D version`, although you should sanitize any sensitive info that comes out of that command (file paths, etc.)

Thanks!

houglum:
^ The above stuff being said, just looking through the logic, I do agree that your fix needs to be added in.  I'd just like to be able to reliably reproduce the issue in order to add a test and ensure this doesn't get broken again.

darkloup:
Thanks for taking interest in this pull request and for your work on gsutil.
On my side it was easy to reproduce any gsutil cp from gs to aws was hanging with a file>100Mb. 
I tested it with gsutil version: 4.37. 
When I investigated by adding some logs in the python I could clearly see that the first transfer worked, but the second (for checksum) was always hanging after the first chunk download.
And it wasn't just an impression:)
I left it running for a long time and nothing happened but with my fix it took only a couple of second to complete the transfert. 
You can find the logs below
[log_gsutil.txt](https://github.com/GoogleCloudPlatform/gsutil/files/2945893/log_gsutil.txt)

houglum:
Thanks for the response!  One quick note: the `-D` logs weren't sanitized; they contained some form of your service account's credentials -- both access tokens and (more importantly) the refresh material.  I've removed the link, but to be safe, you'll probably want to invalidate the keyfile specified in your boto file.

The "authorization" headers, any requests to oauth2.googleapis.com, and anything logged by crypt.py should generally be removed, just FYI.  And generally, I'd suggest either sending log files to gs-team@google.com or ACL-protecting them to be visible only by the desired Google accounts - making them publicly visible is a trap, as you'll inevitably forget to sanitize something :)

houglum:
copy/pasting the rest of your answer, as it contained answers to my above questions:

"""
Thanks for taking interest in this pull request and for your work on gsutil.
On my side it was easy to reproduce any gsutil cp from gs to aws was hanging with a file>100Mb.
I tested it with gsutil version: 4.37.
When I investigated by adding some logs in the python I could clearly see that the first transfer worked, but the second (for checksum) was always hanging after the first chunk download.
And it wasn't just an impression:)
I left it running for a long time and nothing happened but with my fix it took only a couple of second to complete the transfert.
You can find the logs below
"""

mkalmanson:
This was just merged in from another PR.

PR #643, opened 2019-03-01
Added config generator for CI test scripts.

catleeball:
Added a shell script to be called by Kokoro when running
integration tests. It generates a config file allowing access
to the bigstore-gsutil-testing account.

Like the integ test shell scripts, this was pulled from Google3
due to GCP Kokoro VMs being unable to access Piper.

### DIFF
--- a/test/ci/kokoro/config_generator.sh
+++ b/test/ci/kokoro/config_generator.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2018 Google LLC
+# Copyright 2019 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/test/ci/kokoro/config_generator.sh
+++ b/test/ci/kokoro/config_generator.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


test/ci/kokoro/config_generator.sh
```diff
@@ -0,0 +1,35 @@
...
+API=$2
+
+cat  << EOF
+[Credentials]
+gs_service_key_file = "$GSUTIL_KEY"
+
+[GSUtil]
+test_notification_url = https://bigstore-test-notify.appspot.com/notify
+default_project_id = bigstore-gsutil-testing
+prefer_api = "$2"
```
walkerjoe: Question: You define API above but use $2, and similarly, you define GSUTIL_KEY and use it in the config, is this occurrence supposed to be $API? or should $GSUTIL_KEY above be $1?

I'm no shell expert, so I could be completely wrong...

test/ci/kokoro/config_generator.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Copyright 2018 Google LLC
```
walkerjoe: Super small nit, 2018 -> 2019 since this is a new file.

test/ci/kokoro/config_generator.sh
catleeball: Good catch, updated line 30 to be `prefer_api = "$API"`

test/ci/kokoro/config_generator.sh
catleeball: Fixed :)

PR #645, opened 2019-03-02
Vendor in Boto repository

catleeball:
This PR adds the third_party/vendored subdirectory. Resources in this directory will be preferred over those in third_party. We will use this directory to store dependencies we've modified, but the dependency's maintainer hasn't yet released those changes upstream.


gsutil.py
```diff
@@ -73,24 +73,40 @@ def OutputAndExit(message):
...
-        'Please re-install gsutil per the installation instructions.' % (
-            libdir, THIRD_PARTY_DIR))
-  sys.path.insert(0, os.path.join(THIRD_PARTY_DIR, libdir, subdir))
+
+# These are special third party libraries that live in
+# `gsutil/third_party/vendored`. These libraries have modifications specifically
+# for gsutil to operate smoothly that may not yet be present in the latest
+# release of the package maintainer's repositoy.
+VENDORED_THIRD_PARTY_LIBS = [
+    ('boto', 'vendored') # XML API and AWS interop dependency
```
walkerjoe: Reversed; boto is the subdir of vendored. 

Also, with the prepend_to_path() method, it could be changed so the dir being passed in is `gsutil/third_party/vendored/` so that 'vendored' doesn't have to be added to the tuple. This also comes in handy for adding future packages like crcmod where an additional subdir is needed to specify python2 or python3.

catleeball: Good suggestion, I'll pass the path including 'vendored' in to `prepend_to_path` so we'll only need to specify boto.

gsutil.py
```diff
@@ -101,7 +101,7 @@ def prepend_to_path(THIRD_PARTY_LIBS, THIRD_PARTY_DIR):
           'The gsutil command cannot work properly when installed this way.\n'
           'Please re-install gsutil per the installation instructions.' % (
               libdir, THIRD_PARTY_DIR))
-    sys.path.insert(0, os.path.join(THIRD_PARTY_DIR, libdir, subdir))
+    sys.path.insert(1, os.path.join(THIRD_PARTY_DIR, libdir, subdir))
```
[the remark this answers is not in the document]
catleeball: Makes sense to me, I'll set it back to 0 and test

gsutil.py
```diff
@@ -73,24 +73,40 @@ def OutputAndExit(message):
...
+# These are special third party libraries that live in
+# `gsutil/third_party/vendored`. These libraries have modifications specifically
+# for gsutil to operate smoothly that may not yet be present in the latest
+# release of the package maintainer's repositoy.
+VENDORED_THIRD_PARTY_LIBS = [
+    ('boto', 'vendored') # XML API and AWS interop dependency
+]
+
+
+def prepend_to_path(THIRD_PARTY_LIBS, THIRD_PARTY_DIR):
```
houglum: Functions should be named `LikeThis()`, not `like_this()`

Also, parameters should be named `like_this`, not `LIKE_THIS`.

Also-also, I don't think we need to pass in `THIRD_PARTY_DIR` here, since it never changes; we can hard-code it into the function.

I think this is a bit more intuitive to read, thoughts?:
```
def PrependThirdPartyLibToPath(path_component_groups):
```

gsutil.py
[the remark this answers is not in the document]
houglum: +1 for leaving the previous behavior in place.

gsutil.py
```diff
@@ -73,24 +73,40 @@ def OutputAndExit(message):
...
-        'There is no %s library under the gsutil third-party directory (%s).\n'
-        'The gsutil command cannot work properly when installed this way.\n'
-        'Please re-install gsutil per the installation instructions.' % (
-            libdir, THIRD_PARTY_DIR))
-  sys.path.insert(0, os.path.join(THIRD_PARTY_DIR, libdir, subdir))
+
+# These are special third party libraries that live in
+# `gsutil/third_party/vendored`. These libraries have modifications specifically
+# for gsutil to operate smoothly that may not yet be present in the latest
+# release of the package maintainer's repositoy.
```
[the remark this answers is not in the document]
catleeball: Ack, will do :)

gsutil.py
```diff
@@ -73,24 +73,40 @@ def OutputAndExit(message):
     ('pyasn1-modules', ''),  # oauth2client dependency
     ('rsa', ''),  # oauth2client dependency
     ('apitools', ''),
-    ('boto', ''),
+    ('boto', ''), # XML API and AWS interop dependency
```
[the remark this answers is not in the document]
catleeball: We sure can! Unless my logic doesn't work, in which case the problem is worse. :wink: I'll drop boto there

gsutil.py
catleeball: Will update to PascalCase! And whoops my bad on the params

With Joe's comment I was gonna have `THIRD_PARTY_DIR` and `VENDOR_DIR` constants passed into `prepend_to_path` so we don't have to specify the `vendored` subdir every time we add a vendored lib.

Agreed that `path_component_groups` sounds nicer! Given above, how about:

```
def PrependThirdPartyLibToPath(path_component_groups, component_path):
```

catleeball:
Okay, made requested changes. :)

gsutil.py
```diff
@@ -73,24 +73,39 @@ def OutputAndExit(message):
...
+# These are special third party libraries that live in
+# `gsutil/third_party/vendored`. These libraries have modifications specifically
+# for gsutil to operate smoothly that may not yet be present in the latest
+# release of the package.
+VENDORED_THIRD_PARTY_LIBS = [
+    ('boto', '') # XML API and AWS interop dependency
+]
+
+
+def PrependThirdPartyLibToPath(path_component_groups, component_path):
```
houglum: Semantic nit: "component_path" doesn't make much sense to me - as a path is composed of its components.  Maybe "parent_path"?

catleeball: SGTM, updated :+1:

PR #647, opened 2019-03-04
Explicit wording in rm docs about deleting buckets

### LINKED ISSUE #567: `gsutil rm -r gs://bucket/` should NOT remove the bucket
It should remove every item in the bucket and halt, leaving an empty bucket. The short doc string for `rm` says "Remove objects" and there's a separate command (`rb`) for removing buckets. This is a bad design choice that's going to screw people over.

houglum:
Fixes #567.

### DIFF
--- a/gslib/commands/rm.py
+++ b/gslib/commands/rm.py
@@ -53,7 +53,7 @@ _DETAILED_HELP_TEXT = ("""
 
 
 <B>DESCRIPTION</B>
-  The gsutil rm command removes objects.
+  The gsutil rm command removes objects and/or buckets.
   For example, the command:
 
     gsutil rm gs://bucket/subdir/*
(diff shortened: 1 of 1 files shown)


houglum:
@starsandskies Looking for input about the wording here.  Adding "and/or buckets" was the best thing I could think of w.r.t. being explicit and also terse.

gslib/commands/rm.py
```diff
@@ -66,9 +66,9 @@
   will remove all objects under gs://bucket/subdir or any of its
   subdirectories.
 
-  You can also use the -r option to specify recursive object deletion. Thus, for
-  example, either of the following two commands will remove gs://bucket/subdir
-  and all objects and subdirectories under it:
+  You can also use the -r option to specify recursive object and/or bucket
```
starsandskies: I'd recommend not adding "and/or bucket" here, since the accompanying example does not remove the bucket, and because there is discussion of using -r to delete a bucket later on in the page.

PR #650, opened 2019-03-05
Remove boto from git submodules.

houglum:
Following the instructions from https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule, because this was way harder than it should have been :upside_down_face: 

This is a follow-up to #645.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "third_party/boto"]
-	path = third_party/boto
-	url = git://github.com/boto/boto.git
 [submodule "third_party/httplib2"]
 	path = third_party/httplib2
 	url = git://github.com/httplib2/httplib2.git
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,6 @@ management tasks, including:
 
 requires = [
     'argcomplete>=1.9.4',
-    'boto==2.49.0',
     'crcmod>=1.7',
     'fasteners>=0.14.1',
     'gcs-oauth2-boto-plugin>=2.2',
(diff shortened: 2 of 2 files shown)


catleeball:
Oh wait, I see travis failed to build boto:

```
xception: Unable to find required boto test source file at /tmp/pip-YA6XcB-build/third_party/boto/tests/integration/s3/mock_storage_service.py or /tmp/pip-YA6XcB-build/gslib/tests/mock_storage_service.py.
```

We might need to update `setup.py` too maybe

catleeball:
Looks like line 169: `packages=find_packages(exclude=['third_party']),` we could add the vendored subdir and maybe that will satisfy travis running `pip install .`

catleeball:
Also possible culprits in setup.py:

https://github.com/GoogleCloudPlatform/gsutil/blob/6376292583d48bc19d19ae8caf47c862a5f07cd5/setup.py#L68

`BOTO_DIR = os.path.join(CURDIR, 'third_party', 'boto'`

and dependency links on line 59:
```
dependency_links = [
    # Note: this commit ID should be kept in sync with the 'third_party/boto'
    # entry in 'git submodule status'.
    # pylint: disable=line-too-long
    'https://github.com/boto/boto/archive/8fac1878734c5ac085b781f619c70ea4b6e913c3.tar.gz#egg=boto-2.49.0',
    # pylint: enable=line-too-long
]
```

houglum:
Was in a hurry to do this before leaving earlier and completely forgot to also update setup.py. Whoops. This is why we have presubmits :upside_down_face: 

After commit e21ad3c, I manually ran `pip install .` in a new virtualenv while in the root of my local repository, and all was well with the world.  Travis also seems to be happy again, PTAL.

PR #652, opened 2019-03-05
Removing the Python 3 blockers from gsutil.py

catleeball:
This is to test the `ci-testing` branch to make sure Kokoro sees
that a PR is open and that it runs integration tests on the PR.

### DIFF
--- a/gsutil.py
+++ b/gsutil.py
@@ -22,8 +22,8 @@ import warnings
 
 # TODO: gsutil-beta: Distribute a pylint rc file.
 
-if not (2, 6) <= sys.version_info[:3] < (3,):
-  sys.exit('gsutil requires python 2.6 or 2.7.')
+# if not (2, 6) <= sys.version_info[:3] < (3,):
+#   sys.exit('gsutil requires python 2.6 or 2.7.')
 
 
 def UsingCrcmodExtension(crcmod_module):
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -99,7 +99,7 @@ _SET_DESCRIPTION = """
     gsutil -m acl set acl.txt gs://bucket/*.jpg
 
   Note that multi-threading/multi-processing is only done when the named URLs
-  refer to objects, which happens either if you name specific objects or 
+  refer to objects, which happens either if you name specific objects or
   if you enumerate objects by using an object wildcard or specifying
   the acl -r flag.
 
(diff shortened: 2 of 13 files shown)


Tenzer:
[This line](https://github.com/GoogleCloudPlatform/gsutil/blob/master/setup.py#L158) would also have to be updated, otherwise you can't install the package on Python 3.

gsutil.py
```diff
@@ -22,8 +22,8 @@
 
 # TODO: gsutil-beta: Distribute a pylint rc file.
 
-if not (2, 6) <= sys.version_info[:3] < (3,):
-  sys.exit('gsutil requires python 2.6 or 2.7.')
+# if not (2, 6) <= sys.version_info[:3] < (3,):
+#   sys.exit('gsutil requires python 2.6 or 2.7.')
```
walkerjoe: We should probably still have some requirements on the python version, like 2.7 or 3.4+.
```
if not ((2, 7) <= sys.version_info[:3] < (3,) or (3, 4) <= sys.version_info[:3] < (4,):
    sys.exit('gsutil requires python 2.7 or 3.4+.')
```

catleeball: Smart thinking! Or even a warning that we only support 2.7, >=3.5 and that unexpected behavior may occur.

catleeball:
Thanks for the feedback @walkerjoe and @Tenzer ! I'll fix these comments ASAP. This branch is mainly to test our CI testing tool which I'm resolving some permissions issues with, so it may be a smidge before these commits are merged into master branch. :smiley:

Tenzer:
@catleeball You're welcome. I came to look into this because I saw that we weren't using the latest version of the library, and was wondering why things started failing when I tried to use the latest version - when I then realised that version 4.35, 4.36 and 4.37 doesn't allow being installed on Python 3 which seemed odd to me.

catleeball:
@Tenzer Definitely! We're currently implementing Python 2 / 3 compatibility; if you're curious you can take a look at our `py-six-current` branch. :)

gsutil.py
houglum: I think we should require the Python version to be ==2.7 or >=3.5, since that's all we officially develop for and test.

catleeball: Sounds good to me! I'll keep the blocker when we ship things. I'm not planning to merge this PR to master ever, it's just Kokoro testing,

catleeball:
Closing this, will continue Kokoro testing in possible future PRs that I'll label [TEST - KOKORO - DO NOT MERGE] :smiley:

PR #658, opened 2019-03-06
Catch up py-six-current with master [--strategy theirs]

catleeball:
Same as PR #657 but I used `git merge --strategy-option theirs origin/master`. This PR is for comparison which changes from master were excluded. If we'd rather merge this one and make changes on top of it, that's fine too.


catleeball:
Verified that PR #657 looks reasonable, closing this one

PR #661, opened 2019-03-07
Python 3 compatibility fixes in TestCp

walkerjoe:


catleeball:
LGTM!

gslib/tests/testcase/integration_testcase.py
```diff
@@ -906,9 +906,10 @@ def RunGsUtil(self, cmd, return_status=False,
     status = p.returncode
 
     if expected_status is not None:
-      map(six.ensure_text, cmd)
+      if six.PY2:
+        cmd = [item.decode("utf-8") for item in cmd]
```
houglum: QQ: Why does `ensure_text()` not work here?

catleeball: Now that I'm glancing at it, I didn't do list(map(ensure text)) in the removed line, so maybe it was doing the ole lazy evaluation and never actually running over the whole cmd list? And since list comprehensions are eager, maybe ensure_text would be valid in the list comp?

houglum: I'm not advocating for one over the other. I was just curious if there was some gotcha with `ensure_text()` that I didn't know about.

walkerjoe: Initial reason for change: elements in cmd were not being encoded correctly
Upon further review: Apparently I forgot how a map function works and accidentally forgot to assign it back to cmd in the previous commit...

It works either way, once `cmd = ` is added in front of the map function, technically this is quicker though since nothing will be called when running in python 3.

PR #662, opened 2019-03-07
DU Fixes

walkerjoe:
* using new method for formatting strings

* conditions for encoding/decoding depending on python version

* added ensure_str to ls helper to ensure comparisons are against same
type

### DIFF
--- a/gslib/commands/du.py
+++ b/gslib/commands/du.py
@@ -147,8 +147,12 @@ class DuCommand(Command):
   def _PrintSummaryLine(self, num_bytes, name):
     size_string = (MakeHumanReadable(num_bytes)
                    if self.human_readable else six.text_type(num_bytes))
-    text_util.ttyprint('%(size)-11s  %(name)s' % {
-        'size': size_string, 'name': name, 'ending':self.line_ending})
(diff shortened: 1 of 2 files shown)


gslib/commands/du.py
```diff
@@ -177,10 +181,12 @@ def _PrintInfoAboutBucketListingRef(self, bucket_listing_ref):
       num_objs = 1
 
     if not self.summary_only:
-      sys.stdout.write('%(size)-11s  %(url)s%(ending)s' % {
-          'size': size_string,
-          'url': url_str.encode(UTF8),
-          'ending': self.line_ending})
+      if six.PY2:
+        url_str.encode(UTF8)
```
houglum: I'm guessing that since this isn't being stored in a variable, it's just encoding to verify that the url_str can be represented using `utf8`?  I.e. it will throw a `UnicodeDecodeError` in PY2 if `url_str` was a bytes type and used `latin1` chars?

walkerjoe: yeah, it was part of testing, I'm making a quick PR to do some cleanups and this will be part of it.

PR #664, opened 2019-03-07
Cleanup

walkerjoe:
Removing unnecessary code

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -495,20 +495,6 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
       self.assertIn('Skipping existing item: %s' % suri(f), stderr)
       self.assertEqual(f.read(), b'quux')
 
-  @SequentialAndParallelTransfer
-  def test_noclobber_different_size(self):
-    key_uri = self.CreateObject(contents='foo')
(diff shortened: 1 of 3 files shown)


gslib/tests/test_cp.py
```diff
@@ -495,20 +495,6 @@ def test_noclobber_different_size(self):
       self.assertIn('Skipping existing item: %s' % suri(f), stderr)
       self.assertEqual(f.read(), b'quux')
 
-  @SequentialAndParallelTransfer
```
catleeball: Sorry about this merge goof, thanks for catching it. :)

houglum: My test! Don't kill it! ;)

catleeball: Oh!! Sorry, I thought I'd messed up the merge when we saw there was two of those! Thanks for letting us know that's intentionally there lol :smiley:

walkerjoe: @houglum - there are intentionally 2 tests called test_noclobber_different_size() ?

houglum: Derp. I was looking at the master branch (where this wasn't duplicated) and wondering why you removed it -- looks like this test actually got added in twice in the py-six-current branch.  Thanks for catching that :)

PR #670, opened 2019-03-10
Update storage_v1_* apitools generated modules.

houglum:
This change updates the JSON API client files in
gslib/third_party/storage_apitools to be consistent with the source
documents (via the JSON API discovery doc) and updates docstrings to
the new style that apitools generates (using raw strings), which makes
this change a bit noisy.

This commit also adds the "includeTrailingDelimiter" field to some
message types, but that is uninteresting because it's unused in the
code.


houglum:
@catleeball: Note that most of these changes aren't necessarily meaningful, but they're what's autogenerated via apitools from our discovery document... so I've merged these changes in order to minimize the number of diffs for future merges, since that's what future autogenerated files will contain as well :)

catleeball:
LGTM!

PR #671, opened 2019-03-11
Encoding for win

walkerjoe:
* Open file with utf-8 encoding when computing code checksum in commands/version.py.

* Try to write using default encoding in ttyprint, fallback to utf-8 if default is unsuccessful.

Both fixes are in response to errors found on Windows.

### DIFF
--- a/gslib/commands/version.py
+++ b/gslib/commands/version.py
@@ -33,6 +33,7 @@ from gslib.command import Command
 from gslib.utils import system_util
 from gslib.utils.boto_util import GetFriendlyConfigFilePaths
 from gslib.utils.boto_util import UsingCrcmodExtension
+from gslib.utils.constants import UTF8
 from gslib.utils.parallelism_framework_util import CheckMultiprocessingAvailableAndInit
 
 
@@ -169,10 +170,10 @@ class VersionCommand(Command):
         m.update(content)
         f.close()
       else:
-        f = open(filepath, 'r', encoding='utf-8')
+        f = open(filepath, 'r', encoding=UTF8)
         content = f.read()
         content = re.sub(r'(\r\n|\r|\n)', '\n', content)
-        m.update(content.encode('utf-8'))
+        m.update(content.encode(UTF8))
         f.close()
     return m.hexdigest()
 
--- a/gslib/commands/version.py
+++ b/gslib/commands/version.py
@@ -169,7 +169,7 @@ class VersionCommand(Command):
         m.update(content)
         f.close()
       else:
-        f = open(filepath, 'r')
+        f = open(filepath, 'r', encoding='utf-8')
         content = f.read()
         content = re.sub(r'(\r\n|\r|\n)', '\n', content)
         m.update(content.encode('utf-8'))
--- a/gslib/utils/text_util.py
+++ b/gslib/utils/text_util.py
@@ -302,30 +302,31 @@ def ttyprint(*objects, **kwargs):
     ttywrite(file, data)
   else:  # PY3
     def Py3Print(encoding, sep=sep, end=end):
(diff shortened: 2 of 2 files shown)


gslib/commands/version.py
```diff
@@ -169,7 +169,7 @@ def _ComputeCodeChecksum(self):
         m.update(content)
         f.close()
       else:
-        f = open(filepath, 'r')
+        f = open(filepath, 'r', encoding='utf-8')
```
houglum: At some point once all of this works, we should go back and replace all the `"utf-8"` occurrences with `UTF8` from `gslib.utils.constants`.  There are mixed usages scattered throughout the code already though, so no action needed on this PR :)

gslib/utils/text_util.py
```diff
@@ -301,22 +301,37 @@ def ttyprint(*objects, **kwargs):
...
+      for object in objects:
+        if isinstance(object, bytes):
+          byte_objects.append(object)
+        elif isinstance(object, str):
+          byte_objects.append(object.encode(encoding))
+        else:
+          byte_objects.append(str(object).encode(encoding))
+      data = sep.join(byte_objects)
+      data += end
+      ttywrite(file, data)
```
houglum: Add a blank line after the end of this method definition? It currently runs together with the next code block.

gslib/utils/text_util.py
```diff
@@ -301,22 +301,37 @@ def ttyprint(*objects, **kwargs):
...
-        byte_objects.append(object)
-      elif isinstance(object, str):
-        byte_objects.append(object.encode(pref_enc))
-      else:
-        byte_objects.append(str(object).encode(pref_enc))
-    data = sep.join(byte_objects)
-    data += end
-    ttywrite(file, data)
+    def Py3Print(encoding, sep=sep, end=end):
+      """ Passing in encoding allows for a try-except block to attempt the
```
houglum: Nit: The summary line of a docstring should be only one line.

```
"""Some short summary that cuts off before 80 chars.

Some longer summary text here.
"""
```

gslib/utils/text_util.py
```diff
@@ -301,22 +301,37 @@ def ttyprint(*objects, **kwargs):
...
+      default encoding and then fall back to utf-8 on failure to encode
+
+      Args:
+        encoding: Encoding to be used for all encode operations while printing.
+        sep: Line separator to be used for printing (defined in outer scope).
+        end: End character to be used for printing (defined in outer scope).
+      Return: None
```
houglum: Need an extra line of whitespace between end of `Args` and start of `Return`.  Also, `Return` should be `Returns`, or I think the linter will whine.

gslib/utils/text_util.py
```diff
@@ -301,22 +301,37 @@ def ttyprint(*objects, **kwargs):
...
+      """
+      if isinstance(sep, str):
+        sep = sep.encode(encoding)
+      if isinstance(end, str):
+        end = end.encode(encoding)
+      byte_objects = []
+      for object in objects:
```
catleeball: Nit: `object` is a reserved word; it may be good to rename this to disambiguate from the builtin type

gslib/commands/version.py
walkerjoe: Changed this one as well as the one below it, since I was doing changes anyway. (hopefully a couple less to change in the future)

gslib/utils/text_util.py
walkerjoe: Good point, changed to `for item in objects:`

gslib/utils/text_util.py
walkerjoe: Whitespace adjusted and changed `Return` to `Returns`

PR #672, opened 2019-03-11
Added configs for child jobs

catleeball:
Added these configs so we can inject specific environment variables
into each child job, namely which Python version to use.

See: go/kokoro-env-vars

Note, these differ from the job configs in Google3 in two ways:
1. `find . -type f -name *.cfg -exec sed -i 's|https://INTERNAL_LINK.com|go/kokoro-env-vars|g' {} \;`
2. `find . -type f -name *.cfg -exec sed -i 's|allowed_env_vars|env_vars|g' {} \;`

### DIFF
--- a/test/ci/kokoro/linux/py27_json.cfg
+++ b/test/ci/kokoro/linux/py27_json.cfg
@@ -1,0 +1,23 @@
+type: SUB_JOB
+
+group_spec {
+  parent_job_name: "cloud_storage_gsutil/linux/presubmit"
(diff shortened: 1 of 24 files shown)


walkerjoe:
@catleeball - Are you adding `allowed_env_vars`? or has that already been added?

catleeball:
@walkerjoe that's already added in the job configs in google3 :)

walkerjoe:
ah, gotcha, saw it in the docs and wanted to make sure!

PR #673, opened 2019-03-12
du windows fixes - text encoding

walkerjoe:
* Previous write to `sys.stdout.write` produced an error in windows, using ttyprint corrected the error.

* Previous method for gathering Pattern Exclusion only accepted UTF8, new method allows alternate encodings to be provided.

* Substituted map-lambda method and added try/except block for encoding

### DIFF
--- a/gslib/commands/du.py
+++ b/gslib/commands/du.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
+import locale
 import sys
 
 import six
@@ -33,6 +34,7 @@ from gslib.utils import ls_helper
 from gslib.utils.constants import NO_MAX
 from gslib.utils.constants import S3_DELETE_MARKER_GUID
 from gslib.utils.constants import UTF8
+from gslib.utils.text_util import ttyprint
 from gslib.utils.unit_util import MakeHumanReadable
 from gslib.utils import text_util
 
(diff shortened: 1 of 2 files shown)


gslib/commands/du.py
```diff
@@ -214,15 +232,12 @@ def RunCommand(self):
           if a == '-':
             f = sys.stdin
           else:
-            f = open(a, 'r')
+            f = open(a, 'r', encoding=UTF8)
```
houglum: I didn't think you could supply `encoding` as a kwarg for Python 2's `open` function? How does this work on Py2?

gslib/commands/du.py
```diff
@@ -113,6 +115,21 @@
...
 
+LOCAL_ENC = locale.getpreferredencoding(False)
+
+
+def _GetPatternExclusion(file, position, encoding):
+  exclusion_patterns = []
+  file.seek(position)
+  for line in file:
+    if six.PY2:
+      line = line.strip().decode(encoding)
```
houglum: Why are we getting unicode values for PY2 (decoding to unicode) and bytes values for PY3 (encoding to some encoding)?

gslib/commands/du.py
```diff
@@ -113,6 +115,21 @@
       gsutil -o GSUtil:default_project_id=project-name du -shc
 """)
 
+LOCAL_ENC = locale.getpreferredencoding(False)
```
houglum: If we only use this RunCommand, can we just define it as local variable there?

gslib/commands/du.py
walkerjoe: That's true, I could have sworn I ran Py2 test with it coming up clean, but I guess I missed it. Changes in #674 to fix.

gslib/commands/du.py
walkerjoe: The logic was copy/pasted and I didn't dig too far into it. Upon revisiting, a simpler process using `six.ensure` can be used. Changes in #674

gslib/commands/du.py
walkerjoe: Both this and `_GetPatternExclusion` can be removed. Changes in #674 .

PR #674, opened 2019-03-12
Du comment changes

walkerjoe:
Changes in response to Matt's comments

* Py2 has no `encoding` option in the `open` command. Added if/else case to handle both versions.

* `_GetPatternExclusion` was a copy-paste for the encode/decode logic and could have been revised and simplified in-line.

* `LOCAL_ENC` no longer needed (tested with mac/linux/win both py2 and 3 with no issues).

### DIFF
--- a/gslib/commands/du.py
+++ b/gslib/commands/du.py
@@ -116,19 +116,6 @@ _DETAILED_HELP_TEXT = ("""
 """)
 
 
-def _GetPatternExclusion(file, position, encoding):
-  exclusion_patterns = []
-  file.seek(position)
-  for line in file:
-    if six.PY2:
-      line = line.strip().decode(encoding)
-    else:
-      line = line.strip().encode(encoding)
-    if line:
-      exclusion_patterns.append(line)
-  return exclusion_patterns
-
-
 class DuCommand(Command):
   """Implementation of gsutil du command."""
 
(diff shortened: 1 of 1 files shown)


gslib/commands/du.py
```diff
@@ -231,14 +216,12 @@ def RunCommand(self):
...
-            f = open(a, 'r', encoding=UTF8)
-          position = f.tell()
-          try:
-            self.exclude_patterns = _GetPatternExclusion(f, position, UTF8)
-          except:
-            self.exclude_patterns = _GetPatternExclusion(f, position, LOCAL_ENC)
-          finally:
+            f = open(a, 'r') if six.PY2 else open(a, 'r', encoding=UTF8)
+            f_close = True
+          self.exclude_patterns = [six.ensure_text(line.strip()) for line in f]
```
houglum: Q: The locale stuff has been removed in your changes -- `ensure_text` decodes with utf-8 unless otherwise specified; are there any cases where we want to use the system's preferred encoding instead of utf-8?

walkerjoe: As far as our testing goes, there doesn't appear to be any need for system-preferred encoding. I'm not sure if that would limit the types of files that could be read, and if it does, how likely one of those other types would be used.

PR #675, opened 2019-03-12
Newline added in Win/Py3 with CSV writer issue fixed.

walkerjoe:
Windows with Python 3 adds a newline character at the end of a csv write
which can cause unexpected issues. Opening with newline specified as an
empty string can fix this issue.

REF: https://stackoverflow.com/questions/3191528/csv-in-python-adding-an-extra-carriage-return-on-windows

### DIFF
--- a/gslib/utils/copy_helper.py
+++ b/gslib/utils/copy_helper.py
@@ -3694,9 +3694,13 @@ class Manifest(object):
     # Aquire a lock to prevent multiple threads writing to the same file at
     # the same time. This would cause a garbled mess in the manifest file.
     with self.lock:
-      with open(self.manifest_path, 'a', 1) as f:  # 1 == line buffered
-        writer = csv.writer(f)
-        writer.writerow(data)
+      if IS_WINDOWS and six.PY3:
+        f = open(self.manifest_path, 'a', 1, newline='')
+      else:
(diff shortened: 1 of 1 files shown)


gslib/utils/copy_helper.py
```diff
@@ -3694,9 +3694,13 @@ def _WriteRowToManifestFile(self, url):
     # Aquire a lock to prevent multiple threads writing to the same file at
     # the same time. This would cause a garbled mess in the manifest file.
     with self.lock:
-      with open(self.manifest_path, 'a', 1) as f:  # 1 == line buffered
-        writer = csv.writer(f)
-        writer.writerow(data)
+      if IS_WINDOWS and six.PY3:
+        f = open(self.manifest_path, 'a', 1, newline='')
+      else:
+        f = open(self.manifest_path, 'a', 1)  # 1 == line buffered
```
catleeball: Our of curiosity, why do we specify line buffering specifically here instead of using [open's default buffering behavior](https://docs.python.org/3/library/functions.html#open)?

houglum: I'm fine with line-buffering this... I'm assuming this is here so that we don't end up with a half-written line, if we encounter some weird error with one of the fields?

walkerjoe: I'm not completely sure why it was there, I only added the newline arg...

PR #682, opened 2019-03-15
Add support for conditions in IAM policy bindings.

houglum:
This change allows `gsutil iam get` to successfully parse and print out
IAM policies for buckets/objects that contain conditions in their policy
bindings. It also allows setting such policies via `gsutil iam set`.
Finally, it prohibits `gsutil iam ch` from interacting with any such
resources (even if the desired bindings changes contain no conditions),
and presents an error message to users that instructs them to instead use
"iam get", followed by editing the policy and then using "iam set".


houglum:
@chestercun and @mohammadnasre I already showed you most of this (everything but the tests), but please give one more look-over and let me know if this looks good to you.

chestercun:
I just gave all 5 of your commits another read-through, LGTM.

gslib/commands/config.py
```diff
@@ -250,6 +250,8 @@
       tab_completion_time_logs
       tab_completion_timeout
       task_estimation_threshold
+      test_cmd_regional_bucket_location
```
mohammad-nasr: Are these related to IAM Condition?

gslib/commands/iam.py
```diff
@@ -137,8 +142,15 @@
   The gsutil -m option may be set to handle object-level operations more
   efficiently.
 
+  Note: The ``iam ch`` command may NOT be used to the change IAM policy of a
```
mohammad-nasr: Should we also clearly mention that not only they cannot update a policy with condition, they can not add condition to a policy with no condition neither.

gslib/commands/iam.py
```diff
@@ -499,7 +538,14 @@ def _PatchIam(self):
 
     # TODO: Add an error counter for files and objects.
     if not self.everything_set_okay:
-      raise CommandException('Some IAM policies could not be patched.')
+      msg = 'Some IAM policies could not be patched.'
+      if self.tried_ch_on_resource_with_conditions:
+        msg += '\n'
+        msg += '\n'.join(textwrap.wrap(
+           'Some resources had conditions present in their IAM policy '
```
mohammad-nasr: would it make sense to have same message here and above (line 427-429)?

houglum: No -- one message is for a specific resource, while the other is a "summary" message, meaning that "one or more" resources ran into this problem.  The general pattern here (and in other places in gsutil where we support this continue-on-error mode) is that users may re-run the command without the continue-on-error flag, or run in debug mode (`-D`) to see which resources/requests are producing errors.

gslib/commands/iam.py
houglum: Sure, good call. I tweaked the wording a bit.

gslib/commands/config.py
houglum: Sort of :)

The Test environment I was using doesn't have the `us-central1` region, so I added this flag to override the default region we use.  And I'm unable to test against the usual environment because of whitelists.  So since this was necessary for testing, yes :upside_down_face:

I then also realized that we weren't documenting `test_notification_url` either, so I documented that and added it into the commit too.

mohammad-nasr: Thanks for clarification.

gslib/commands/iam.py
mohammad-nasr: Ack

PR #683, opened 2019-03-15
Metrics Test Fixes for Windows / py3

walkerjoe:
## gsutil metrics test changes for Python3 in Windows

* Changes around submodule Popen to work with Windows / Py3

* Style Guide Indentation Change

* Standardize env_vars

* Add Raw String option

NOTE: All Metrics tests remain passing in macOS & Linux with the above changes

### DIFF
--- a/gslib/metrics.py
+++ b/gslib/metrics.py
@@ -731,12 +731,17 @@ class MetricsCollector(object):
       log_file_path = '"%s"' % log_file_path
 
     reporting_code = six.ensure_str(
-        ('from gslib.metrics_reporter import ReportMetrics; '
-        'ReportMetrics("{0}", {1}, log_file_path={2})').format(
+        'from gslib.metrics_reporter import ReportMetrics; '
+        'ReportMetrics("{0}", {1}, log_file_path={2})'.format(
             temp_metrics_file.name,
             log_level,
             log_file_path))
-    execution_args = [sys.executable, '-c', reporting_code]
+    # Python 3 path in windows usually involves spaces (ex: C:\Program Files...)
+    # so it has to be wrapped in double quotes.
+    sys_exc = sys.executable
+    if system_util.IS_WINDOWS and six.PY3:
+      sys_exc = '"%s"' % sys_exc
+    execution_args = [sys_exc, '-c', reporting_code]
     exec_env = os.environ.copy()
     exec_env['PYTHONPATH'] = os.pathsep.join(sys.path)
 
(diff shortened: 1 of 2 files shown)


gslib/metrics.py
```diff
@@ -728,19 +728,25 @@ def ReportMetrics(
...
-      log_file_path = '"%s"' % log_file_path
-
-    reporting_code = ('from gslib.metrics_reporter import ReportMetrics; '
-                      'ReportMetrics("{0}", {1}, log_file_path={2})').format(
-                          temp_metrics_file.name,
-                          log_level,
-                          log_file_path).encode('utf-8').decode('unicode_escape')
+      log_file_path = 'r"%s"' % log_file_path
+
+    reporting_code = six.ensure_str(
```
houglum: Q: Why is the `ensure_str` call needed here?  Are we worried about any of the types we're plugging into the string being the `bytes` type?

Having `unicode_literals` imported means that the format string should be a unicode type, right?  In general, to avoid errors that come from type-mismatches (like adding a bytes type into a unicode format string and getting something like `my_value=b'abc'` instead of `my_value=abc`), I prefer to convert each of the arguments going into the format string to the `str` type first, rather than feeding them all to the format string and then casting the result to a string.

gslib/metrics.py
```diff
@@ -728,19 +728,25 @@ def ReportMetrics(
...
+        'from gslib.metrics_reporter import ReportMetrics; '
+        'ReportMetrics(r"{0}", {1}, log_file_path={2})'.format(
+            temp_metrics_file.name,
+            log_level,
+            log_file_path))
     execution_args = [sys.executable, '-c', reporting_code]
     exec_env = os.environ.copy()
     exec_env['PYTHONPATH'] = os.pathsep.join(sys.path)
-
+    sm_env = dict()
```
houglum: What is "sm" abbreviating here?

gslib/metrics.py
```diff
@@ -728,19 +728,25 @@ def ReportMetrics(
...
+    for k, v in six.iteritems(exec_env):
+      sm_env[six.ensure_str(k)] = six.ensure_str(v)
     try:
-      p = subprocess.Popen(execution_args, env=exec_env)
+      # In order for Popen to work correctly with Windows/Py3 and spaces in the
+      # path, shell needs to be True, in addition to
```
houglum: I think you trailed off in the middle of this thought and stopped typing :)

gslib/metrics.py
walkerjoe: sub-module_environment

gslib/metrics.py
walkerjoe: haha, yes... yes I did....

gslib/metrics.py
houglum: Might be worth typing the extra characters, then. I wouldn't have guessed that one :)

PR #688, opened 2019-03-20
Bunch of small doc fixes.

starsandskies:
Fixes are based on open issues in the output documentation found on cloud.google.com.

Most edits are minor; the most significant change is to stat.py, which [I believe] contains outdated text about the consistency of bucket listing.

### DIFF
--- a/gslib/commands/stat.py
+++ b/gslib/commands/stat.py
@@ -53,10 +53,7 @@ _DETAILED_HELP_TEXT = ("""
     gsutil ls -L gs://some-bucket/some-object
 
   but is more efficient because it avoids performing bucket listings and gets
-  the minimum necessary amount of object metadata. Moreover, because it avoids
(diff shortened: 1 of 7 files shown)


gslib/commands/stat.py
```diff
@@ -53,10 +53,7 @@
     gsutil ls -L gs://some-bucket/some-object
 
   but is more efficient because it avoids performing bucket listings and gets
-  the minimum necessary amount of object metadata. Moreover, because it avoids
-  performing bucket listings (which are eventually consistent) the gsutil stat
```
houglum: Why did you choose to remove this section?  IMO, it seems useful.  The only thing I'd change here is that bucket and object listings *may* be eventually consistent, depending on the storage provider (e.g. listing objects in an S3 bucket).

starsandskies: I hadn't considered non-Google Cloud Storage providers. Edited.

PR #691, opened 2019-03-21
Removed `allow_failures` for python 3 tests

walkerjoe:
py-six-current branch passes all unit tests in python 2 and 3 (hooray!)
so we should require python 3.x to pass all future tests.


walkerjoe:
NOTE: all tests pass, using the newest boto, aka: boto/boto-six. One failure on Py3 versions until then.

PR #692, opened 2019-03-21
dev.py: recommend using --user and virtualenv instead of sudo

remyabel:


houglum:
Sorry for being late to reply -- I think these changes were fine, but it looks like your fork is gone now.

PR #695, opened 2019-03-21
 Updated Mac and Linux CI scripts and configs to working status.

catleeball:
This PR reflects commits to `ci-testing` branch where I performed
several iterations of testing on our CI scripts and configs.
These broad refactors should allow Linux CI to run tests on
manual invokations. Mac test script now symlinks to Linux's test
script since they use identical logic now.  Also, bumped processes to 8.

This PR is a resubmission of https://github.com/GoogleCloudPlatform/gsutil/pull/693 with Mac changes. Git was giving me guff and making a new PR was honestly just easier.


test/ci/kokoro/config_generator.sh
```diff
@@ -14,22 +14,22 @@
...
 
 GSUTIL_KEY=$1
 API=$2
+OUTPUT_FILE=$3
 
-cat  << EOF
-[Credentials]
-gs_service_key_file = "$GSUTIL_KEY"
+cat > $3 <<- EOM
+	[Credentials]
```
houglum: The `<<-` ignores tabs, right? I'd rather not have to tweak my editor settings to mess with this in the future (vim changes all tabs to spaces for me, for most file extensions). Could you put this back to left-aligned text?

test/ci/kokoro/linux/run_integ_tests.sh
```diff
@@ -2,33 +2,37 @@
...
 
-# Processes to use based on default Ubuntu Kokoro specs here:
-# go/gcp-ubuntu-vm-configuration-v32i
-PROCS="4"
 
-# PYMAJOR and PYMINOR environment variables are set for each Kokoro job in:
+# PYMAJOR, PYMINOR, and BOTO_CONFIG environment variables are set per job in:
 # go/kokoro-gsutil-configs
 PYVERSION="$PYMAJOR.$PYMINOR"
+API=${BOTO_CONFIG##*_}
```
houglum: What on earth is this `##*_` stuff? Lol

test/ci/kokoro/config_generator.sh
catleeball: For sure! I'll undo the tabs and change `<<-` to `<<`.

test/ci/kokoro/linux/run_integ_tests.sh
catleeball: I'll add a comment! It uses [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) matches the string after the rightmost instance of `_` in the BOTO_CONFIG environment variable.

So `/tmpfs/src/.boto_json` yields `json`. Alternatively, we can change to variable to just be `xml` or `json`, then append a path to it in the test runner script here, albeit we'll also need to change all the configs.

Typing this out, the latter sounds more proper. Let me know if you have any preference. 👍

catleeball: Updated the environment variables in cl/239702660 and the most recent commit here.

PR #699, opened 2019-03-25
Merge master branch into py-six-current

catleeball:
Catch up our py-six-current branch with master.


gslib/tests/test_cp.py
```diff
@@ -495,6 +495,20 @@ def test_noclobber_different_size(self):
       self.assertIn('Skipping existing item: %s' % suri(f), stderr)
       self.assertEqual(f.read(), b'quux')
 
+  @SequentialAndParallelTransfer
```
houglum: I think this is one that @walkerjoe removed that we didn't also remove in the master branch.  This looks like a duplicate of the test above it, and should actually be removed.

catleeball: Thanks for catching that, removed it. :+1:

PR #702, opened 2019-03-26
Remove note about object composition limits.

starsandskies:
These limits no longer apply in Cloud Storage.


starsandskies:
Think I got 'em :)

PR #705, opened 2019-03-26
Added Windows config generator, build wrapper

catleeball:
Added a wrapper .bat script to call our powershell testing script.
Kokoro looks specifically for a .bat script to call. Additionally,
created a config generator for Windows and removed some environment
variables that are instead being calculated at runtime now.

### DIFF
--- a/test/ci/kokoro/windows/run_integ_tests.bat
+++ b/test/ci/kokoro/windows/run_integ_tests.bat
@@ -1,0 +1,35 @@
+rem Copyright 2019 Google LLC
+rem
+rem Licensed under the Apache License, Version 2.0 (the "License");
+rem you may not use this file except in compliance with the License.
+rem You may obtain a copy of the License at
+rem
+rem     http://www.apache.org/licenses/LICENSE-2.0
+rem
+rem Unless required by applicable law or agreed to in writing, software
+rem distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


test/ci/kokoro/windows/run_integ_tests.bat
```diff
@@ -0,0 +1,35 @@
...
+tree /F
+echo ""
+
+
+set GsutilRepoDir=%1
+set "PyExePath=C:\python%PYMAJOR%%PYMINOR%\python.exe"
+
+cmd config_generator.bat "C:\src\keystore\74008+gsutil_kokoro_service_key" %API% "C:\src\.boto_%API%"
+
+PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& '%run_integration_tests.ps1%' '%GsutilRepoDir%' '%PyExePath%'";
```
houglum: Q: Can these args be supplied positionally?  I *thought* they had to be passed like `-PyExe="value"`, but I haven't tried just passing the value without its param name.

catleeball: Was going off this page http://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/

It looked like we mighttt be able to pass args positionally, but I changed it to use named args anyways. :)

PR #716, opened 2019-04-02
Use compatibale six.moves.zip function

catleeball:
This function was using itertools.izip which is only available on
Python 2.x. For compatibility, we're now using six.moves.zip.

### DIFF
--- a/gslib/commands/iam.py
+++ b/gslib/commands/iam.py
@@ -644,11 +644,8 @@ class IamCommand(Command):
 
       policy_it = itertools.repeat(protojson.encode_message(policy))
       self.Apply(
-          _SetIamWrapper,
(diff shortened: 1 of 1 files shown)


gslib/commands/iam.py
```diff
@@ -645,7 +645,7 @@ def _SetIam(self):
       policy_it = itertools.repeat(protojson.encode_message(policy))
       self.Apply(
           _SetIamWrapper,
-          itertools.izip(
+          zip(
```
houglum: The args can now (probably) fit on the same line, right?

catleeball: Oh, yes! I compacted these down a little in latest commit.

PR #717, opened 2019-04-02
Fixed test that wasn't explicity looking for bytes

catleeball:


catleeball:
@houglum This PR is the replacement for https://github.com/gsutil-mirrors/boto/pull/1 now with fixed test. :)

PR #720, opened 2019-04-03
Moved: see PR 722

catleeball:
Currently in Windows, cp_streaming test fails due to having line
ending \r\n instead of \n. Edited to look for os.linesep instead.

Bug: b/129889651

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -605,11 +605,12 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
   @SequentialAndParallelTransfer
   def test_streaming(self):
     bucket_uri = self.CreateBucket()
+    expected = six.ensure_binary('bar' + os.linesep)
     stderr = self.RunGsUtil(['cp', '-', '%s' % suri(bucket_uri, 'foo')],
                             stdin='bar', return_stderr=True)
     self.assertIn('Copying from <STDIN>', stderr)
     key_uri = bucket_uri.clone_replace_name('foo')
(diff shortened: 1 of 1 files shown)


walkerjoe:
Do we care if it's `os.linesep` or does it have to return `bar\n`?

catleeball:
> Do we care if it's `os.linesep` or does it have to return `bar\n`?

@walkerjoe I think so! The contents of the bucket are created by streaming from stdin, which on windows will normally have \r\n. So it seemed reasonable to expect os.linesep back since the shell will be putting os.lineseps into the file.

houglum:
I'm not sure how the `linesep` stuff got added to `RunGsUtil()` in `gslib/tests/testcase/integration_testcase.py`, but that doesn't seem right to me.  We shouldn't be appending any bytes to what's being passed in as stdin.  Gsutil doesn't do that today, e.g.:

```
# Note that `-n` tells us not to append a trailing newline, like `echo` normally would.
houglum@host $ echo -n "123456789" | gsutil cp - gs://BUCKET/obj2
Copying from <STDIN>...
/ [1 files][    0.0 B/    0.0 B]
Operation completed over 1 objects.

houglum@host $ gsutil cat gs://BUCKET/obj2
123456789houglum@host $ 
```

catleeball:
@houglum Understood, I'll remove the added line breaks from stdin in RunGsUtil and smooth things out from there. 👍

catleeball:
There were some new commits on the head of py-six-current and making a new PR is easier than doing a merge. Moving this to https://github.com/GoogleCloudPlatform/gsutil/pull/722 CC: @houglum

PR #725, opened 2019-04-04
Broaden `MonkeyPatchHttp()` for additional versions of Python3

walkerjoe:
Some versions of python do not print values along with the keys
of headers returned. Original monkeypatch was only for 3.6.5, but
testing showed additional version of python also needed the fix.

### DIFF
--- a/gslib/utils/system_util.py
+++ b/gslib/utils/system_util.py
@@ -250,7 +250,9 @@ def IsRunningInteractively():
 
 
 def MonkeyPatchHttp():
+  # Getting python version array in the following format: [major, minor, micro]
   py_ver = [int(part) for part in platform.python_version().split('.')]
+  # Checking for and applying monkeypatch to python versions 3.4 - 3.6.7
   if (py_ver[0] == 3 and
       4 <= py_ver[1] < 6 or (py_ver[1] == 6 and py_ver[2] < 8)):
     _MonkeyPatchHttpForPython_3x()
--- a/gslib/utils/system_util.py
+++ b/gslib/utils/system_util.py
@@ -250,7 +250,7 @@ def IsRunningInteractively():
 
 def MonkeyPatchHttp():
   ver = sys.version_info
-  # Checking for and applying monkeypatch to python versions 3.4 - 3.6.6
+  # Checking for and applying monkeypatch to python versions 3.0 - 3.6.6
   if (ver.major == 3 and (ver.minor < 6 or (ver.minor == 6 and ver.micro < 7))):
     _MonkeyPatchHttpForPython_3x()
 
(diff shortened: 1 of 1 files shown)


gslib/utils/system_util.py
```diff
@@ -250,11 +250,12 @@ def IsRunningInteractively():
 
 
 def MonkeyPatchHttp():
-  if platform.python_version().startswith('3.6.5'):
-    _MonkeyPatchHttpForPython_3_6_5()
+  py_ver = [int(part) for part in platform.python_version().split('.')]
```
catleeball: Nit: instead of splitting the full version and pulling it apart into a list, you can just do:

```
# Python 3.5.6
sys.version_info.major  # -> 3
sys.version_info.minor  # -> 5
sys.version_info.micro  # -> 6
```

walkerjoe: Applied this method - good catch @catleeball !

gslib/utils/system_util.py
```diff
@@ -250,11 +249,19 @@ def IsRunningInteractively():
...
-def _MonkeyPatchHttpForPython_3_6_5():
+  # Getting python version array in the following format: [major, minor, micro]
+  py_ver = [
+      sys.version_info.major,
+      sys.version_info.minor,
+      sys.version_info.micro
+  ]
+  # Checking for and applying monkeypatch to python versions 3.4 - 3.6.7
+  if (py_ver[0] == 3 and
+      4 <= py_ver[1] < 6 or (py_ver[1] == 6 and py_ver[2] < 8)):
```
catleeball: Sorry if I was unclear! My nit before mainly applied here, where it would be easier to read by not using list indexes for comparison :)

```
import sys.version_info as ver
if ver.major == 3:
    if ver.minor < 6 or (ver.minor == 6 and ver.micro < 8):
        _MonkeyPatchHttpForPython_3x()
```

houglum: Any PY3 version below v3.6.7 should have our monkeypatch

walkerjoe: changed

gslib/utils/system_util.py
```diff
@@ -250,11 +249,13 @@ def IsRunningInteractively():
 
 
 def MonkeyPatchHttp():
-  if platform.python_version().startswith('3.6.5'):
-    _MonkeyPatchHttpForPython_3_6_5()
+  ver = sys.version_info
+  # Checking for and applying monkeypatch to python versions 3.4 - 3.6.7
+  if (ver.major == 3 and (ver.minor < 6 or (ver.minor == 6 and ver.micro < 8))):
```
houglum: Nit: `ver.micro < 8` should be `ver.micro < 7`, since 3.6.7 prints the header values.

walkerjoe: Changed!

Issue #29, opened 2013-02-22
make gsutil work on Python 3.x


jterrace:
_Original author: mfschwa...@google.com (August 19, 2010 22:33:59)_

Currently if you try to run gsutil on Python v3.x it will get syntax errors at print statements, because Python 3.x made print into a function, so the old syntax:
  print 'abcd'
doesn't work, and instead you need:
  print('abcd')

There are likely other problems beyond just this to make gsutil work on v3.x. Moreover, at present the boto library only works on Python v2.x.

At some point, when boto moves to support Python 3.x, gsutil should as well.

_Original issue: http://code.google.com/p/gsutil/issues/detail?id=29_

stefanv:
Please consider removing the "Priority-Low" tag.  Python 3 is fast becoming the default on many platforms.

l3d00m:
It's been a while since the last comment here (and over 4 years since this issue was created). Is there any status update for this issue available?

houglum:
This is still fairly low priority at this point, unfortunately.  There are a several blockers for us starting on this this work, a few of them being:

- Higher-priority tasks taking up development time
- Not all dependencies are fully compatible with Python 3 (some are partially ported, or working but without full feature parity)
- Gsutil is packaged with gcloud, which only supports Python 2 syntax currently -- users that use gsutil via gcloud would see no benefit from any Python 3 compatibility changes

ADiegoCAlonso:
IMHO, this issue is quite high priority for a lot of developers and new comers to the GCP. So, if I take a look at my crystal ball, I see two futures:

-   **Option A**: GCP develops soon an independent version of *gsutil* that is compatible with Py3. This version will be used as reference for *gcloud* migration to Py3. Happy developers will use this new *gsutil* since the beginning.
-   **Option B**:  GCP will release the *gsutil* that is compatible with Py3 after X years together with *gcloud*. In this time, new libraries will appear that are not official but will more popular that *gsutil*. Developers will not be eager to change their codebase to use the new official library. Unofficial libraries will not be totally compatible with the API. Bottom line: it will be a mess.

kotoroshinoto:
it seems like the priority should be raised either this year or next year, https://python3statement.org/
https://pythonclock.org/
python 2 is being slowly abandoned

gdahlm:
Additional data that may help with priority:

[...]

PR #726, opened 2019-04-04
Merge branch 'py-six-current' into 'py-six-onto-master-2019-04-05'

catleeball:
Merging in the 'py-six-current' feature branch which introduces
compatibility changes for Python 2.7 and 3.5+ interoperability.

For bug #29


.gitignore
```diff
@@ -9,3 +9,6 @@ gsutil_*.zip
 .settings/*
 .coverage*
 .metadata/*
+.idea/*
+.vscode/*
+venv*/*
```
catleeball: Nit: might be nice to add `.DS_Store` to gitignore

catleeball: Added a commit with this

gslib/tests/testcase/integration_testcase.py
```diff
@@ -853,35 +873,58 @@ def RunGsUtil(self, cmd, return_status=False,
...
-      cmd = [sys.executable] + cmd
+    cmd = [
+        gslib.GSUTIL_PATH,
+        '--testexceptiontraces',
+        '-o',
+        'GSUtil:default_project_id=' + PopulateProjectId()
+    ] + cmd
+    if stdin is not None:
+      if six.PY3:
+        if isinstance(stdin, bytes):
```
houglum: I don't think this first check is necessary -- "if it's bytes, ensure that it's really bytes" :P

gslib/tests/testcase/unit_testcase.py
```diff
@@ -85,8 +90,9 @@ def setUp(self):
     self.stdout_save = sys.stdout
     self.stderr_save = sys.stderr
     fd, self.stdout_file = tempfile.mkstemp()
-    sys.stdout = os.fdopen(fd, 'w+')
+    sys.stdout = os.fdopen(fd, 'wb+')
     fd, self.stderr_file = tempfile.mkstemp()
+    # do not set sys.stderr to be 'wb+' - it will blow up the logger
```
houglum: Blow up? Like it will crash, or just start printing garbage, or what?

gslib/tests/testcase/unit_testcase.py
```diff
@@ -169,7 +186,7 @@ def RunCommand(self, command_name, args=None, headers=None, debug=0,
```
houglum: Can you indicate the type of the returned things? (i.e. is stdout and stderr going to be bytes or unicode?)

gslib/tests/util.py
```diff
@@ -189,10 +217,31 @@ def _HasS3Credentials():
 def _HasGSHost():
   return boto.config.get('Credentials', 'gs_host', None) is not None
 
+
+def string_compare(a, b, encoding='utf-8'):
```
houglum: We really only use this in one place, just a few lines below.  And I feel like this could just be avoided if we coerced (via `six.ensure_foo`) the two things to be the same type before comparing them.

gslib/tests/util.py
```diff
@@ -543,19 +598,30 @@ def GetTestNames():
       names.append(m.group('name'))
   return names
 
+def is_file(obj):
+  if six.PY2:
+    return isinstance(obj, file)
+  else:
```
houglum: Nit: This `else:` is unnecessary, the code will still get executed if we fall through the if check above.

gslib/utils/boto_util.py
```diff
@@ -392,6 +400,66 @@ def _PatchedGetPluginMethod(cls, requested_capability=None):
 
   boto.plugin.get_plugin = _PatchedGetPluginMethod
 
+  #########################################################################
```
houglum: We don't need this method patch anymore since we're using our updated version of Boto.

gslib/utils/cloud_api_helper.py
```diff
@@ -67,6 +72,10 @@ def GetDownloadSerializationData(
   if user_project:
     url = AddQueryParamToUrl(url, 'userProject', user_project)
 
+  if six.PY3:
+    if isinstance(url, bytes):
+      url = url.decode('ascii')
```
houglum: Let's use utf-8 for this, in case the URL has any non-url-encoded utf-8 chars in it.

gslib/utils/copy_helper.py
```diff
@@ -1976,8 +2010,9 @@ def _UploadFileToObject(
       logger, allow_splitting, upload_url, dst_url, src_obj_size,
       gsutil_api, canned_acl=global_copy_helper_opts.canned_acl,
       kms_keyname=dst_obj_metadata.kmsKeyName)
-  non_resumable_upload = (upload_size < ResumableThreshold() or
-                          src_url.IsStream() or src_url.IsFifo())
+  non_resumable_upload = (
+    (0 if upload_size is None else upload_size) < ResumableThreshold() or
```
houglum: This first parenthetical can be simplified to `upload_size or 0` -- even if it's not None but still falsy (meaning it's zero), we'll still end up with 0.

gslib/utils/copy_helper.py
```diff
@@ -3586,18 +3621,32 @@ def _CreateManifestFile(self):
...
-                           'Destination',
-                           'Start',
-                           'End',
-                           'Md5',
-                           'UploadId',
-                           'Source Size',
-                           'Bytes Transferred',
-                           'Result',
-                           'Description'])
+        if six.PY3:
```
houglum: Can we factor this list out (exactly as it is, making sure to keep the order and all)?  We type it twice and take up lots of lines for no reason.

gslib/utils/hashing_helper.py
```diff
@@ -252,7 +262,7 @@ def CalculateMd5FromContents(fp):
 
 def Base64EncodeHash(digest_value):
   """Returns the base64-encoded version of the input hex digest value."""
```
houglum: I'd expect something that returns a base64-encoded version of something to have a return type of `bytes`.  Can you add to the docstring, noting the return type is unicode?

gslib/utils/hashing_helper.py
```diff
@@ -265,7 +275,8 @@ def Base64ToHexHash(base64_hash):
   Returns:
     Hex digest of the input argument.
   """
-  return binascii.hexlify(base64.decodestring(base64_hash.strip('\n"\'')))
+  return binascii.hexlify(
```
houglum: Ditto here, please note the return type (bytes) in the method docstring.

gslib/utils/system_util.py
```diff
@@ -244,8 +248,55 @@ def IsRunningInteractively():
...
 def StdinIterator():
   """A generator function that returns lines from stdin."""
   for line in sys.stdin:
     # Strip CRLF.
     yield line.rstrip()
+
+
+class StdinIteratorCls(six.Iterator):
+  """An iterator that returns lines from stdin.
+     This is needed because Python 3 balks at pickling the
```
houglum: This docstring formatting is oddly spaced.

```
"""It should look like this, then a blank line.

Followed by lines indent-aligned with the first triple quotes.
And stuff.
"""
```

gslib/utils/text_util.py
```diff
@@ -16,26 +16,42 @@
...
+
+
 STORAGE_CLASS_SHORTHAND_TO_FULL_NAME = {
-    # Values should remain uppercase, as required by non-gs providers.
-    'CL': 'COLDLINE',
-    'DRA': 'DURABLE_REDUCED_AVAILABILITY',
-    'NL': 'NEARLINE',
-    'S': 'STANDARD',
-    'STD': 'STANDARD'}
+  # Values should remain uppercase, as required by non-gs providers.
```
houglum: Please fix the spacing back to what it was before (4 spaces instead of 2)

gslib/utils/text_util.py
```diff
@@ -16,26 +16,42 @@
...
+  'CL': 'COLDLINE',
+  'DRA': 'DURABLE_REDUCED_AVAILABILITY',
+  'NL': 'NEARLINE',
+  'S': 'STANDARD',
+  'STD': 'STANDARD'}
 
 VERSION_MATCHER = LazyWrapper(
-    lambda: re.compile(r'^(?P<maj>\d+)(\.(?P<min>\d+)(?P<suffix>.*))?'))
+  lambda: re.compile(r'^(?P<maj>\d+)(\.(?P<min>\d+)(?P<suffix>.*))?'))
```
houglum: Please fix the spacing back to what it was before (4 spaces instead of 2)

gslib/utils/text_util.py
```diff
@@ -62,22 +78,13 @@ def AddQueryParamToUrl(url_str, param_name, param_value):
...
-  if isinstance(url_str, unicode):
-    url_str = url_str.encode('utf-8')
-  if isinstance(param_name, unicode):
-    param_name = param_name.encode('utf-8')
-  if isinstance(param_value, unicode):
-    param_value = param_value.encode('utf-8')
-  scheme, netloc, path, query_str, fragment = urlparse.urlsplit(url_str)
-
-  query_params = urlparse.parse_qsl(query_str, keep_blank_values=True)
+  scheme, netloc, path, query_str, fragment = urllib.parse.urlsplit(url_str)
```
houglum: Do we expact all arguments to be unicode now?  I'm curious if the notes in the docstring about types still apply or not.

gslib/utils/text_util.py
```diff
@@ -174,19 +181,19 @@ def FixWindowsEncodingIfNeeded(input_str):
   to Unicode.
 
   Args:
-    input_str: (str) The input string.
+    input_str: (str or bytes) The input string.
```
houglum: "unicode or bytes" instead?  The return type is listed as "unicode", so I'd like to be consistent at least within the same docstring.

gslib/utils/text_util.py
```diff
@@ -203,7 +210,27 @@ def InsistAsciiHeaderValue(header, value):
       value,
       'Invalid non-ASCII value (%s) was provided for header %s.\nOnly ASCII '
       'characters are allowed in headers other than x-goog-meta- and '
-      'x-amz-meta- headers' % (value, header))
+      'x-amz-meta- headers' % (repr(value), header))
+
+
+def InsistOnOrOff(value, message):
```
houglum: 1) So, it looks like this got added 4 extra times, for a total of 5 definitions. Not sure how that happened.

2) Please add a docstring, to whichever of them you decide to keep, saying that `value` should be of type unicode.

gslib/utils/text_util.py
```diff
@@ -245,9 +272,110 @@ def PrintableStr(input_val):
   Returns:
     (str) A UTF-8 encoded string, or None.
   """
-  return input_val.encode(UTF8) if input_val is not None else None
+  return input_val
```
houglum: This... no longer does what the docstring says. In fact, it basically does nothing now. Do we still need this method?

gslib/wildcard_iterator.py
```diff
@@ -41,16 +46,20 @@
...
 FLAT_LIST_REGEX = re.compile(r'(?P<before>.*?)\*\*(?P<after>.*)')
 
 _UNICODE_EXCEPTION_TEXT = (
-    'Invalid Unicode path encountered (%s). gsutil cannot proceed '
-    'with such files present. Please remove or rename this file and '
-    'try again. NOTE: the path printed above replaces the '
-    'problematic characters with a hex-encoded printable '
-    'representation. For more details (including how to convert to a '
-    'gsutil-compatible encoding) see `gsutil help encoding`.')
+  'Invalid Unicode path encountered (%s). gsutil cannot proceed '
```
houglum: Please fix the spacing back to what it was before (4 spaces instead of 2)

gsutil.py
```diff
@@ -16,14 +16,26 @@
...
+
 import os
 import sys
 import warnings
 
 # TODO: gsutil-beta: Distribute a pylint rc file.
 
-if not (2, 6) <= sys.version_info[:3] < (3,):
-  sys.exit('gsutil requires python 2.6 or 2.7.')
+if not ((2, 6) <= sys.version_info[:3] < (3,)
```
houglum: We only need `version_info[:2]` for both of these checks, right? Since we're only concerned with major and minor version, not patch version

gsutil.py
```diff
@@ -16,14 +16,26 @@
...
+        or (3, 4) <= sys.version_info[:3]):
+    sys.exit('gsutil requires python 2.7 or 3.4+.')
```
houglum: "3.4+" to me sounds like "3.4 or higher".   Change this to "3.5+"?

catleeball: will do 👍

gsutil.py
catleeball: sounds good, I'll clean this up maybe into a function in utils or somewhere since `__init__.py` does the same check

gslib/__init__.py
```diff
@@ -22,15 +22,15 @@
...
 
 from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
 
 import os
 import pkgutil
 import sys
 import tempfile
```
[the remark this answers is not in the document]
catleeball: Sounds good to me. Since I'm also cleaning this up in gsutil.py I'll make a single function for them both in utils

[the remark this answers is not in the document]
walkerjoe: Doesn't gsutil.py have a version check at the top? I think it closes the program if anything other than the approved Python versions try to run.

gsutil.py
walkerjoe: Changed message and logic

gsutil.py
walkerjoe: Logic revised. I can add to utils or something if that would be a better place, and then just call it at this stage and at __init__.py

gslib/wildcard_iterator.py
walkerjoe: Fixed indentation spacing.

gslib/utils/text_util.py
walkerjoe: I see no reason to keep it... there is really only one block that uses it in the same file around lines 341 - 343. My vote is to take it out.

gslib/utils/text_util.py
walkerjoe: Removed duplicates and added docstrings to a couple functions, including `InsistOnOrOff()`

gslib/commands/rsync.py
```diff
@@ -811,7 +820,10 @@ def _EncodeUrl(url_string):
   Returns:
     encoded URL.
   """
-  return urllib.quote_plus(url_string.encode(constants.UTF8))
+  if six.PY2:
+    return urllib.parse.quote_plus(url_string.encode(constants.UTF8))
+  else:
```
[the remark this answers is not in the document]
walkerjoe: Reconfigured Method

gslib/commands/rsync.py
```diff
@@ -823,7 +835,10 @@ def _DecodeUrl(enc_url_string):
   Returns:
     decoded URL.
   """
-  return urllib.unquote_plus(enc_url_string).decode(constants.UTF8)
+  if six.PY2:
+    return urllib.parse.unquote_plus(enc_url_string).decode(constants.UTF8)
+  else:
```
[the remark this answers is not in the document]
walkerjoe: Reconfigured Method

gslib/commands/ls.py
```diff
@@ -317,26 +322,26 @@ def _PrintBucketInfo(self, bucket_blr, listing_style):
...
-                  bucket.defaultObjectAcl)}
-
-    fields['versioning'] = bucket.versioning and bucket.versioning.enabled
-    fields['website_config'] = 'Present' if bucket.website else 'None'
-    fields['logging_config'] = 'Present' if bucket.logging else 'None'
-    fields['cors_config'] = 'Present' if bucket.cors else 'None'
-    fields['lifecycle_config'] = 'Present' if bucket.lifecycle else 'None'
-    fields['requester_pays'] = bucket.billing and bucket.billing.requesterPays
+    fields = {
+      'bucket': bucket_blr.url_string,
```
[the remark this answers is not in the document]
walkerjoe: Added spacing

gslib/commands/signurl.py
```diff
@@ -272,10 +277,13 @@ def _GenSignedUrl(key, client_id, method, duration,
       query_string=canonical_query_string, headers=canonical_headers,
       signed_headers=canonical_signed_headers, hashed_payload=_UNSIGNED_PAYLOAD)
 
+  if six.PY3:
+    canonical_request = canonical_request.encode()
```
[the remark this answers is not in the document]
walkerjoe: Encoding UTF8 added

gslib/commands/signurl.py
```diff
@@ -272,10 +277,13 @@ def _GenSignedUrl(key, client_id, method, duration,
...
+
   canonical_request_hasher = hashlib.sha256()
   canonical_request_hasher.update(canonical_request)
   hashed_canonical_request = base64.b16encode(
-      canonical_request_hasher.digest()).lower()
+      canonical_request_hasher.digest()).lower().decode()
```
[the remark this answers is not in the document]
walkerjoe: Encoding UTF8 added

gslib/commands/signurl.py
```diff
@@ -287,7 +295,20 @@ def _GenSignedUrl(key, client_id, method, duration,
...
+  if six.PY2:
+    digest = b'RSA-SHA256'
+  else:
+    # Your IDE may complain about this due to a bad docstring in pyOpenSsl:
+    # https://github.com/pyca/pyopenssl/issues/741
+    digest = 'RSA-SHA256'
+
+  # If string looks like u'...', strip off the u' '
+  # See: https://github.com/GoogleCloudPlatform/gsutil/pull/619/files/19fe4b93543d1423f10792c96de2edc4a5a06db9#r256283790
+  if string_to_sign[:2] == "'u" and string_to_sign[-1:] == "'":
```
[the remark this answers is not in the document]
walkerjoe: Looked up the history, it has always been like it shows now, ever since it was added.

gslib/parallel_tracker_file.py
```diff
@@ -71,6 +78,9 @@ def ReadParallelUploadTrackerFile(tracker_file_name, logger):
     tracker_data = tracker_file.read()
     tracker_json = json.loads(tracker_data)
     enc_key_sha256 = tracker_json[_CompositeUploadTrackerEntry.ENC_SHA256]
+    if six.PY3:
+      if enc_key_sha256:
+        enc_key_sha256.encode('ascii')
```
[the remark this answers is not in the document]
walkerjoe: Context: This was a change made by Ed.

gslib/tests/test_daisy_chain_wrapper.py
```diff
@@ -167,9 +171,9 @@ def testDownloadWithPartialWrite(self):
     """Tests unaligned writes to the download stream from GetObjectMedia."""
     with open(self.test_data_file, 'rb') as stream:
       chunk = stream.read(TRANSFER_BUFFER_SIZE)
-    one_byte = chunk[0]
+    one_byte = chunk[0:1]
```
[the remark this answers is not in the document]
walkerjoe: Did some debugging, without extra `:1`, `one_byte`'s value is `77`, with it, the value is `M`
(coming from: `b'MJoTM...''.

gslib/tests/test_defacl.py
```diff
@@ -34,6 +38,9 @@ class TestDefacl(case.GsUtilIntegrationTestCase):
   _defacl_get_prefix = ['defacl', 'get']
   _defacl_set_prefix = ['defacl', 'set']
 
+  def setUp(self):
+    super(TestDefacl, self).setUp()
```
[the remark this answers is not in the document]
walkerjoe: Tests perform the same with or without. I think it was either testing other things, like you mentioned, or someone being very explicit in including a setup method that calls super's setup method. A similar thing was done in the parent class, but it also added some functionality beyond just calling super class's setup().

Class method removed.

gslib/tests/test_gcs_json_media.py
```diff
@@ -21,10 +21,24 @@
...
 from gslib.gcs_json_media import UploadCallbackConnectionClassFactory
 import gslib.tests.testcase as testcase
-import mock
+import six
+
+from six import add_move, MovedModule
+add_move(MovedModule('mock', 'mock', 'unittest.mock'))
+from six.moves import mock
+
+if six.PY2:
```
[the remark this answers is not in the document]
walkerjoe: six does have `six.http_client` that could do both, but this test passes that text into the decorator `@mock.patch` which takes a string argument in the form of 'package.module.ClassName'.

gslib/tests/test_help.py
```diff
@@ -31,37 +34,37 @@ class HelpUnitTests(testcase.GsUtilUnitTestCase):
 
   def test_help_noargs(self):
     stdout = self.RunCommand('help', return_stdout=True)
-    self.assertIn('Available commands', stdout)
+    self.assertIn(b'Available commands', stdout)
```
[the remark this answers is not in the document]
walkerjoe: stdout is returned as bytes.

gslib/tests/test_naming.py
```diff
@@ -683,13 +687,13 @@ def testLsBucketNonRecursive(self):
...
     output = self.RunCommand('ls', [suri(src_bucket_uri, '*')],
                              return_stdout=True)
-    expected = set([suri(src_bucket_uri, 'foo1'),
-                    suri(src_bucket_uri, 'd1', ':'),
-                    suri(src_bucket_uri, 'd1', 'd2') + src_bucket_uri.delim,
-                    suri(src_bucket_uri, 'd0', ':'),
-                    suri(src_bucket_uri, 'd0', 'foo2')])
-    expected.add('')  # Blank line between subdir listings.
-    actual = set(output.split('\n'))
+    expected = set([suri(src_bucket_uri, 'foo1').encode('utf-8'),
```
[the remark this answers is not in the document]
walkerjoe: Good point, changed to recommendation. Also added UTF8 as an imported constant.

gslib/tests/testcase/integration_testcase.py
walkerjoe: Shortened to only encode if *not* bytes.

gslib/tests/testcase/unit_testcase.py
walkerjoe: Context: comment by Ed.

gslib/tests/testcase/unit_testcase.py
walkerjoe: Added comments to docstring for return values.

gslib/tests/util.py
walkerjoe: There are actually a LOT of these kind of methods in this block. Removed all one-time-use, single-line methods.

gslib/tests/util.py
walkerjoe: Removed `else` from method

gslib/__init__.py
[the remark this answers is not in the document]
houglum: For certain installation methods (pip installations), the gsutil.py script won't exist, and `gsutil` will be a script that simply invokes `gslib.__main__()`.  So we need the check here as well.

gslib/boto_resumable_upload.py
```diff
@@ -333,7 +343,11 @@ def _UploadFileBytes(self, conn, http_conn, fp, file_length,
     # in debug stream.
     http_conn.set_debuglevel(0)
     while buf:
-      http_conn.send(buf)
+      buf_bytes = buf
+      if six.PY3:
+        if isinstance(buf, str):
+          buf_bytes = buf.encode('utf-8')
+      http_conn.send(buf_bytes)
       total_bytes_uploaded += len(buf)
```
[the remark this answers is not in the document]
houglum: Marking as resolved; opened a comment thread on the "right" side off the diff instead - please comment there.

gslib/boto_resumable_upload.py
```diff
@@ -333,7 +343,11 @@ def _UploadFileBytes(self, conn, http_conn, fp, file_length,
     # in debug stream.
     http_conn.set_debuglevel(0)
     while buf:
-      http_conn.send(buf)
+      buf_bytes = buf
+      if six.PY3:
+        if isinstance(buf, str):
+          buf_bytes = buf.encode('utf-8')
+      http_conn.send(buf_bytes)
```
houglum: Moving my comment to the "right" side of the diff (GitHub for some reason put it on the left last time):

This should be `total_bytes_uploaded += len(buf_bytes)` instead, since `buf_bytes` could be longer than `buf`, its `str` equivalent -- calling `len()` on a unicode object counts the number of code points, not the number of bytes needed to encode the thing in utf-8 or any other codec.

gslib/commands/signurl.py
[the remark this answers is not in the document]
houglum: Well, that leads me to believe that this isn't doing what we thought it was, so we should either remove it or figure out what it's doing instead.

Maybe add a debug print statement to print `string_to_sign` both before and after this if-block?

gslib/parallel_tracker_file.py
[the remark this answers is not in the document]
houglum: We should... try printing this out and seeding what `enc_key_sha256` contains, and what type it is... and seeing if that's consistent with what's used/expected below.

As it is, this is doing nothing, so we should either remove it or see if we actually need to convert the thing from unicode to ascii-bytes.

gslib/tests/test_daisy_chain_wrapper.py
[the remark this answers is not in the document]
houglum: Whoa, you're right! That's... neat.

```
$ python3
>>> s = b'MJo'
>>> s[0]
77
>>> s[0:1]
b'M'
>>> type(s[0:1])
<class 'bytes'>
>>> type(s[0])
<class 'int'>
```

Could you leave a comment above this explaining that chunk[0] would return an `int`, but this returns a `bytes` object?

gslib/tests/test_gcs_json_media.py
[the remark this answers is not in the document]
houglum: Ahhhhh got it.

Could you add a comment explaining this?

gslib/__init__.py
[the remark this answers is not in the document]
walkerjoe: Gotcha, per my other comment, should this check maybe be put in utils and then called from `gsutil.py` and `gslib/__init__.py`?

gslib/tests/test_daisy_chain_wrapper.py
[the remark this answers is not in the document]
walkerjoe: Added comment.

gslib/tests/test_gcs_json_media.py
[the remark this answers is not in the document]
walkerjoe: Comment added, similar to my comment above, explaining why strings are used instead of a direct import or something like `six.moves`

gslib/tests/test_help.py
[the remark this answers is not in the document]
walkerjoe: Comments in RunCommand's docstring should now show return types for stdout, stderr, etc... Hope this clarifies it in the future!

gslib/utils/cloud_api_helper.py
walkerjoe: I tried using UTF8 and I started to get failures on tests that are testing utf8 characters. If we change it here, it could require further digging to get it to a useable state again.

gslib/__init__.py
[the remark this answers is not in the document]
catleeball: @walkerjoe I made a thing in utils last night, about to PR it in just a quick minute :)

gslib/utils/text_util.py
walkerjoe: Removed method and calls. Tests that use that section of code have been tested and are still working.

gslib/__init__.py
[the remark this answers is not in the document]
catleeball: @walkerjoe whoops, I was going to PR it but accidentally just pushed it to this branch, so the commit is on the head of this PR.

I went maybe a little too ham, sorry for maybe being a bit verbose in my implementation. :stuck_out_tongue: 

It looks like to import from gslib from gsutil.py we'd need to add a top level __init__.py, but I'm not sure if that's something we'd want to do given it's just supposed to wrap gslib/__main__.py. Thoughts?

gsutil.py
catleeball: See the most recent two commits and the comment on gslib/__init__.py; thoughts on adding a top level __init__.

gslib/commands/signurl.py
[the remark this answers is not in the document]
walkerjoe: As far as I can tell, it never hits the code inside the if-statement, even with the change to `u'`.

I added a generic `raise Exception` inside and it and it never triggered an error on `test signurl` tests

gslib/parallel_tracker_file.py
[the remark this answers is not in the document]
walkerjoe: If needed, the conversion must already happen elsewhere. After it's done in the if-statement, it just gets returned.

gslib/utils/version_check.py
```diff
@@ -0,0 +1,65 @@
...
+  """Exit if gsutil is being run in an incompatible version of Python.
+
+  This function compares the running version of cPython and against the list
+  of supported python versions maintained in the constants file. If the running
+  version is less than any of the supported versions, exit.
+
+  Versions of Python greater than those listed in the currently supported
+  versions are implicitly allowed.
+  """
+  supported = constants.SUPPORTED_PYTHON_VERSIONS
```
houglum: Discussed offline, the functionality in this module should be refactored a bit.

gslib/commands/signurl.py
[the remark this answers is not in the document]
houglum: Oh cool.  Well, this whole blocks sounds unnecessary, so let's just remove it.

gslib/parallel_tracker_file.py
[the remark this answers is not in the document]
houglum: Well, it sounds like we can just remove the whole if-block, then.  SGTM.

gslib/utils/version_check.py
catleeball: Committed refactor!

gslib/utils/constants.py
```diff
@@ -26,14 +26,26 @@
...
+import six
+
 from gslib.utils.unit_util import ONE_GIB
 from gslib.utils.unit_util import ONE_KIB
 from gslib.utils.unit_util import ONE_MIB
 
+
+# Key:   Int, Python major version
+# Value: Sorted tuple of ints, Python minor versions
+SUPPORTED_PYTHON_VERSIONS = {
```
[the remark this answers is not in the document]
catleeball: Done :)

PR #730, opened 2019-04-09
Refactored test_cp to decode strings

catleeball:
Previous logic was manually removing `b' '` from around strings that
weren't decoded. Refactored so that strings are decoded and don't need
to be stripped.

Bugs: b/130189680 , b/130189227

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1545,37 +1545,20 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
     with open(logpath, 'r') as f:
       lines = f.readlines()
     if six.PY3:
-      # What the hell is this?
-      # Well, it is a hack. So far as I can tell, CreateTempFile has no need
-      # for text file output except for here (where it is used to create a
-      # logfile), and the other place this 'code' appears in the manifest
-      # download test. So rather than mangle that code to support this one
-      # special case, I did this. The purpose is to get rid of strings that
-      # contain the string representation of bytes. Not all do, but even one
-      # is enough.
-      decode_lines = []
-      for line in lines:
-        if line.startswith("b'"):
-          some_strs = line.split(',')
-          line_parts = []
-          for some_str in some_strs:
-            if some_str.startswith("b'"):
-              line_parts.append(ast.literal_eval(some_str).decode('utf-8'))
-            else:
-              line_parts.append(some_str)
-          decode_lines.append(','.join(line_parts))
-        else:
-          decode_lines.append(line)
-      lines = decode_lines
+      lines = [six.ensure_str(line.encode(UTF8)) for line in lines]
     else:  # PY2
-      lines = [unicode(line, 'utf-8') for line in lines]
+      lines = [unicode(line, UTF8) for line in lines]
     self.assertEqual(len(lines), 2)
 
+    # TODO(b/130189227): Refactor into dictionary results = {header: result}
     expected_headers = ['Source', 'Destination', 'Start', 'End', 'Md5',
                         'UploadId', 'Source Size', 'Bytes Transferred',
                         'Result', 'Description']
     self.assertEqual(expected_headers, lines[0].strip().split(','))
     results = lines[1].strip().split(',')
+    for header in results:
+      if isinstance(header, (six.string_types, six.text_type)):
+        header = six.ensure_str(header.encode(UTF8))
     self.assertEqual(results[0][:7], 'file://')  # source
     self.assertEqual(results[1][:5], '%s://' %
                      self.default_provider)      # destination
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1543,8 +1543,10 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
     open(logpath, 'w').close()
     self.RunGsUtil(['cp', '-L', logpath, fpath, dsturi])
 
-    with io.open(logpath, 'r', encoding=UTF8) as f:
+    with open(logpath, 'r') as f:
       lines = f.readlines()
+    if six.PY2:
+      lines = [unicode(line, UTF8) for line in lines]
 
     self.assertEqual(len(lines), 2)
 
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1562,9 +1562,9 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
 
     results = dict(zip(expected_headers, results))
 
-    self.assertEqual(results['Source'][:7], 'file://')  # source
+    self.assertEqual(results['Source'][:7], 'file://')
     self.assertEqual(results['Destination'][:5], '%s://' %
-                     self.default_provider)      # destination
+                     self.default_provider)
 
     date_format = '%Y-%m-%dT%H:%M:%S.%fZ'
     start_date = datetime.datetime.strptime(results['Start'], date_format)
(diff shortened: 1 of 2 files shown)


catleeball:
@houglum I went back from the list comprehension used in PR #729 because it caused a regression in `test_cp_manifest_upload_unicode` in Python 2.7

gslib/tests/test_cp.py
```diff
@@ -1545,53 +1545,43 @@ def _ManifestUpload(self, file_name, object_name, manifest_name):
...
                         'Result', 'Description']
     self.assertEqual(expected_headers, lines[0].strip().split(','))
     results = lines[1].strip().split(',')
-    self.assertEqual(results[0][:7], 'file://')  # source
-    self.assertEqual(results[1][:5], '%s://' %
-                     self.default_provider)      # destination
+
+    for header in results:
+      if isinstance(header, (six.string_types, six.text_type)):
+        header = six.ensure_str(header.encode(UTF8))
```
houglum: Seems like overkill if we know the lines have been read in as unicode.

gslib/utils/copy_helper.py
```diff
@@ -3681,17 +3681,16 @@ def _WriteRowToManifestFile(self, url):
     """Writes a manifest entry to the manifest file for the url argument."""
     row_item = self.items[url]
```
houglum: What does `row_item` contain at this point -- always a unicode object, or a version-dependent type (bytes in PY2, unicode in PY3)?

It would be neat if we could ensure the former, and would prevent a lot of the crazy casting/{en,de}coding work going on below.

houglum:
Ah, I didn't realize the last PR had both `ensure_str()` and `str()` in the list comprehension.  In Python 2, if you have the bytes `'abcẻ'` (which you do in PY2 since you invoked `str(item)`), and try to call `str(item).encode('utf-8')`, it will implicitly first try to decode those bytes to unicode using the default (ascii) codec.

I think we're better off reading stuff from the file so that the contents we get are unicode (using `io.open`), since we know the file's format will be utf-8 (being that we control the manifest file).

gslib/utils/copy_helper.py
catleeball: It's a variety of types, including version-dependent strings, ints, and datetime tuples, which is why I tried `str()` on all the items when it was a list comprehension.

```
{'description': '', 'bytes': 4247, 'size': 4247, 'end_time': datetime.datetime(2019, 4, 8, 23, 39, 8, 87305), 'source_uri': 'file:///usr/local/google/home/cball/.vimrc', 'result': 'OK', 'start_time': datetime.datetime(2019, 4, 8, 23, 39, 7, 749583), 'md5': 'YSH3vG6cGl7y4+hRBf1G7w==', 'destination_uri': 'gs://do-not-delete-cball-test/.vimrc'}
```

gslib/tests/test_cp.py
catleeball: Makes sense! I'll remove the encode

gslib/tests/test_cp.py
```diff
@@ -1545,53 +1545,43 @@ def _ManifestUpload(self, file_name, object_name, manifest_name):
...
-          for some_str in some_strs:
-            if some_str.startswith("b'"):
-              line_parts.append(ast.literal_eval(some_str).decode('utf-8'))
-            else:
-              line_parts.append(some_str)
-          decode_lines.append(','.join(line_parts))
-        else:
-          decode_lines.append(line)
-      lines = decode_lines
+      lines = [six.ensure_str(line.encode(UTF8)) for line in lines]
```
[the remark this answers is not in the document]
catleeball: Got it! `io.open` sounds cleanest for compat, so I'll give that a try first. Thanks for noticing this!

gslib/utils/copy_helper.py
catleeball: > I think we're better off reading stuff from the file so that the contents we get are unicode (using io.open), since we know the file's format will be utf-8 (being that we control the manifest file).

Ok, changed it to use io.open

catleeball:
@houglum Ok! All changes requested are implemented, namely we're using io.open now in both cases and it's much cleaner. :+1:

catleeball:
@houglum Removed some of the the `io.open` changes since [Python 2.7 CSV writer](https://stackoverflow.com/a/18449496) doesn't agree with `io.open`.

gslib/utils/copy_helper.py
```diff
@@ -3682,14 +3682,14 @@ def _WriteRowToManifestFile(self, url):
...
-        row_item['destination_uri'],
+        six.ensure_str(row_item['source_uri']),
+        six.ensure_str(row_item['destination_uri']),
         '%sZ' % row_item['start_time'].isoformat(),
         '%sZ' % row_item['end_time'].isoformat(),
         row_item['md5'] if 'md5' in row_item else '',
         row_item['upload_id'] if 'upload_id' in row_item else '',
-        str(row_item['size']) if 'size' in row_item else '',
-        str(row_item['bytes']) if 'bytes' in row_item else '',
+        six.ensure_str(str(row_item['size'])) if 'size' in row_item else '',
```
houglum: (Here and potentially elsewhere)

Shouldn't `ensure_str` wrap around the entire rest of the expression?  If we have `unicode_literals` imported, the "else" case will give back a unicode object in PY2, but the "if" case will give back a `bytes` object.

gslib/utils/copy_helper.py
```diff
@@ -3591,7 +3591,7 @@ def _ParseManifest(self):
     """
     try:
       if os.path.exists(self.manifest_path):
-        with io.open(self.manifest_path, 'r', encoding=UTF8) as f:
+        with open(self.manifest_path, 'r') as f:
```
houglum: Please add a comment explaining why we don't use `io.open()` here to open the file in text/unicode mode in both Python versions (i.e. Python2's `csv.reader()` requires the file be opened in binary mode).

catleeball: Sure thing!

gslib/utils/copy_helper.py
catleeball: Ah, that makes sense!

catleeball: I ended up going back to list comprehension if that's ok :) newer comprehension tested good with Python 2.7 and 3.5 running cp tests on json / linux, doing full tests now

houglum: Also, as we discussed offline, the `str()` cast inside of `ensure_str()` is sometimes necessary -- I think `ensure_str` expects the argument to be either `unicode` or `bytes`, so you'll need to cast things like "size" (an int) to a string-like thing first before passing it to `ensure_str()`.

gslib/utils/copy_helper.py
```diff
@@ -3687,11 +3689,12 @@ def _WriteRowToManifestFile(self, url):
...
         row_item['upload_id'] if 'upload_id' in row_item else '',
-        row_item['size'] if 'size' in row_item else '',
-        row_item['bytes'] if 'bytes' in row_item else '',
+        str(row_item['size']) if 'size' in row_item else '',
+        str(row_item['bytes']) if 'bytes' in row_item else '',
         row_item['result'],
         row_item['description']]
-    data = [six.ensure_str(str(item).encode(UTF8)) for item in data]
+
+    data = [six.ensure_str(header) for header in data]
```
houglum: Super nit: I think the name "header" here is misleading.  It will only be header in the case of the first row :)

catleeball: True! maybe 'result' or 'value'? I'll patch in a rename :)

PR #732, opened 2019-04-10
Ensure unicode for rsync temp file attributes

catleeball:
For b/130244083

On the Py2/3 compat branch, rsync would crash when attempting to
recursively rsync a directory containing files with unicode in their
names. This was due to some temp file attributes not being properly
encoded.


catleeball:
Manual tests are all green for Mac / Windows using XML / JSON on versions 2.7, 3.5, 3.6, 3.7  :+1:

PR #737, opened 2019-04-11
Remove deprecated note

smalleel:
Note previously read that we don't accept GitHub PRs, this is no longer true.


smalleel:
This was just to test if the `kokoro:run` label works to run tests on external contributor code, closing PR

PR #749, opened 2019-04-18
Allow configuring compression level

jakewins:
The current gzip compression level is 9, the python stdlib default, which is rather high. Reducing it to a more reasonable value - like 6, the default used by the gzip tool - improves compression speed substantially without losing much space.

In my use case, this improves gsutil end-to-end upload performance about one order of magnitude, though I suspect this will vary depending on compressibility of the files, which is why I'm not proposing changing the default.

With current defaults, uploading a 256MiB database file:
```
$ time ./gsutil cp -Z ~/Downloads/testfile-256Mi gs://jake-compressiontest
Copying file:///home/jake/Downloads/testfile-256Mi [Content-Type=application/octet-stream]...
Compressing file:///home/jake/Downloads/testfile-256Mi (to tmp)...
/ [1 files][ 28.1 MiB/ 28.1 MiB]                                                
Operation completed over 1 objects/28.1 MiB.                                     

real	1m22.710s
user	1m17.232s
sys	0m0.404s
```

With compression set to 6:
```
$ time ./gsutil -o "GSUtil:gzip_compression_level=6" cp -Z ~/Downloads/testfile-256Mi gs://jake-compressiontest/lvl6
Copying file:///home/jake/Downloads/testfile-256Mi [Content-Type=application/octet-stream]...
Compressing file:///home/jake/Downloads/testfile-256Mi (to tmp)...
/ [1 files][ 28.3 MiB/ 28.3 MiB]                                                
Operation completed over 1 objects/28.3 MiB.                                     

real	0m12.101s
user	0m6.844s
sys	0m0.296s
```

### DIFF
--- a/gslib/commands/config.py
+++ b/gslib/commands/config.py
@@ -248,6 +248,7 @@ _DETAILED_HELP_TEXT = ("""
       sliced_object_download_threshold
       parallel_process_count
       parallel_thread_count
+      gzip_compression_level
       prefer_api
       resumable_threshold
       resumable_tracker_dir (deprecated in 4.6, use state_dir)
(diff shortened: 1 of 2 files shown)


houglum:
Thanks for the PR!

Looks good to me.  Adding @catleeball to reviewers to trigger our new integration test matrix against this PR, since we recently added PY3 support as well (manually looking over the `gzip` docs for both 2 and 3, the stuff you've added seems fine, but I'd still like to run the tests).  The Kokoro tests will likely report as a failure until we get all the rough edges of the test configs fixed, just FYI.

catleeball:
LGTM too 👍

PR #752, opened 2019-04-20
Bump gcs-oauth2-boto-plugin version to 2.3

houglum:


catleeball:
LGTM! Thank you for releasing the new version and bumping it in gsutil! Sorry I'm a doofus and never PR'd the gsutil version bump.

PR #754, opened 2019-04-23
Install Py 3.7.2 on Windows when using JSON

catleeball:
Workaround for b/131089489 where the preinstalled version of Python
3.7.0 on Kokoro behaves unexpectedly with the JSON API. The issue
doesn't occur on public Python 3.7.0 with Windows and JSON; it appears
to be isolated to Kokoro.

Unexpected behavior was in `-D` flag tests, headers were being printed in an
order inconsistent with other versions of 3.7.x or with local 3.7.0. Adding the workaround
to the Kokoro test runner seemed like a better place than in the test, since the test
doesn't seem to fail in any other environments except Kokoro Windows with Python 3.7.0
and using the JSON API.

Manual Kokoro test run pointing at `chocoPy` test branch with these changes.
go/chocoPyLogs

### DIFF
--- a/test/ci/kokoro/windows/run_integ_tests.bat
+++ b/test/ci/kokoro/windows/run_integ_tests.bat
@@ -25,6 +25,11 @@ type %BOTO_CONFIG%
 
 cd %GsutilRepoDir%
 git submodule update --init --recursive
+
+rem Workaround for b/131089489 where Kokoro's preinstalled Py 3.7.0 encounters errors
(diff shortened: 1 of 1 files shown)


houglum:
Commented on the internal bug; looks like this is another Python version where we need to apply the httplib header logging hotfix: https://github.com/GoogleCloudPlatform/gsutil/blob/83585ce2c2c45b2c1d3856e10bc5bb2e91ce8e7a/gslib/utils/system_util.py#L251

catleeball:
Closing this in favor of https://github.com/GoogleCloudPlatform/gsutil/pull/755

PR #758, opened 2019-04-23
Fix setup.py to not package test and boto bloat.

houglum:
End-users don't need stuff in the top-level test/ directory; likewise, the only
directory needed under gslib/vendored/boto/ is the boto/ directory.

Internal notes for Googlers: These changes were also included in our
tarball build script in changelist 244894227.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,9 +132,18 @@ setup(
         'Topic :: System :: Filesystems',
         'Topic :: Utilities',
     ],
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
+    # Gsutil supports Python 2.7, 3.5+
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -132,9 +132,18 @@ def make_release_tree(self, base_dir, files):
         'Topic :: System :: Filesystems',
         'Topic :: Utilities',
     ],
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, <4',
+    # Gsutil supports Python 2.7, 3.5+
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4',
```
catleeball: Super nit: would it be worth adding an explicit `>=3.5.*` here?

houglum: We already have a `>=2.7` expression; wouldn't the `>=3.5.*` intersect with that?

catleeball: Oh you're right! Whoop, that makes sense :)

PR #761, opened 2019-04-23
Get GID for -P tests from dest dir

catleeball:
For b/131088388. When running cp, mv, or rsync with -P, sometimes
tests will fail due to how GID was computed. The expected GID was
taken from the current process's GID, while the file's actual GID may differ
if the /tmp dir is owned by a different group than the user's group who
launched the tests.

To rectify this, -P tests now compute expected GID by running stat
on the destination directory, and using its GID. The file being sent
to the cloud exists only in memory until it is downloaded.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -55,11 +55,10 @@ from gslib.utils.system_util import IS_WINDOWS
 # These POSIX-specific variables aren't defined for Windows.
 # pylint: disable=g-import-not-at-top
 if not IS_WINDOWS:
+  from gslib.tests import util
   from gslib.tests.util import DEFAULT_MODE
   from gslib.tests.util import INVALID_GID
   from gslib.tests.util import INVALID_UID
-  from gslib.tests.util import NON_PRIMARY_GID
-  from gslib.tests.util import PRIMARY_GID
   from gslib.tests.util import USER_ID
 # pylint: enable=g-import-not-at-top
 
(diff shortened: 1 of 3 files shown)


gslib/tests/test_rsync.py
```diff
@@ -1460,6 +1460,7 @@ def test_bucket_to_dir_preserve_posix_no_errors(self):
     """Tests that rsync -P works properly with default file attributes."""
     bucket_uri = self.CreateBucket()
     tmpdir = self.CreateTempDir()
+    PRIMARY_GID = os.stat(tmpdir).st_gid
```
houglum: If we're overwriting `PRIMARY_GID` here, there doesn't seem to be a point in keeping it (or the method it calls, `GetPrimaryGid()`) in `gslib/tests/util.py`.  To that effect, the naming convention for this variable can be changed, from `PRIMARY_GID` to `primary_gid`.

Similarly, it looks like we only use `NON_PRIMARY_GID` in these two tests.  Thus, I'd suggest keeping the `GetNonPrimaryGid()` method in `gslib/tests/util.py`, getting rid of the `NON_PRIMARY_GID` module member there (since it's just a LazyWrapper that isn't all that useful), and just using an expression like this at the top of these two tests: 

```
non_primary_gid = GetNonPrimaryGid()
```

catleeball: Got it! Will do :+1:

catleeball: Staged changes on a test branch, running tests: gpaste/6197306502152192

catleeball: A couple more tests are failing post-refactor surrounding tracker files, going to smooth that out and test it again

catleeball: It was a silly typo I think, here's a new batch of tests targeting the `GCS/gsutil : macGidFix` branch: https://paste.googleplex.com/5510027513167872

catleeball: Those runs seemed to go much better. :) Mac and Linux both were all green (except a single flaked io.close on linux that we're aware can happen) and Windows green except resumable test failure (which we have a bug for and will work on next) :+1: 

Will merge the changes on the testing branch into this PR shortly.

catleeball: Commit 3bbdad74bce8fd6730e325f5228da4172f529609 has the refactor and is now in this PR :muscle:

catleeball: Oh wait, this refactor seems to have broken test_tab_complete and test_stat somehow. Investigating.

catleeball: I'm kind of lost how these changes would have impacted either of the above tests. I'm going to ask Kokoro to rerun tests on this to make sure it wasn't a weird test flake.

catleeball: Rerunning tests and the errors still reappear. Specifically, all Mac XML tests encounter errors on `test_tabcomplete.test_subcommands` and all Linux XML tests encounter errors on `test_stat_bucket_wildcard`.

Test logs: https://paste.googleplex.com/4826875685765120

catleeball: Oh! I re-ran against master, and the Tabcomplete and stat_bucket_wildcard issues are there too. So not related to PR which makes me feel less like I've gone nuts. :-)  I'll open bugs for those to smooth out.

catleeball:
Okay, with previous comment now actually resolved, refactor should be operational and ready for review @houglum  :)

PR #766, opened 2019-05-01
XML >1k buckets test flake fix

catleeball:
For b/131254662.
Some tests that rely on tab-complete will fail when using the XML
API and when there are more than 1,000 buckets present. This is because
XML API does not accept prefixes for bucket names, it just returns a full list
off all buckets that gsutil filters to make tab complete work.

Since the list returned it alphabetized, as a workaround to the API limitation,
we append 'aaa-' to effected tabcomplete and wildcard tests to hoist their buckets
to the top of the list returned by the API.

### DIFF
--- a/gslib/tests/test_stat.py
+++ b/gslib/tests/test_stat.py
@@ -42,7 +42,6 @@ from gslib.tests.util import TEST_ENCRYPTION_KEY2
 from gslib.tests.util import TEST_ENCRYPTION_KEY2_SHA256_B64
 from gslib.tests.util import unittest
 from gslib.utils.retry_util import Retry
-from gslib.utils.text_util import get_prefix
 
 
 class TestStat(testcase.GsUtilIntegrationTestCase):
@@ -140,7 +139,8 @@ class TestStat(testcase.GsUtilIntegrationTestCase):
     self.assertIn('%s:' % suri(bucket_uri, 'notmissing'), stdout)
 
   def test_stat_bucket_wildcard(self):
-    bucket_uri = self.CreateBucket(bucket_name_prefix=get_prefix())
+    # Prefix is a workaround for XML API limitation, see PR 766 for details
+    bucket_uri = self.CreateBucket(bucket_name_prefix='aaa-')
     self.CreateObject(bucket_uri=bucket_uri, object_name='foo', contents=b'z')
     stat_string = suri(bucket_uri)[:-1] + '?/foo'
     self.RunGsUtil(['stat', stat_string])
--- a/gslib/tests/test_stat.py
+++ b/gslib/tests/test_stat.py
@@ -42,6 +42,7 @@ from gslib.tests.util import TEST_ENCRYPTION_KEY2
 from gslib.tests.util import TEST_ENCRYPTION_KEY2_SHA256_B64
 from gslib.tests.util import unittest
 from gslib.utils.retry_util import Retry
+from gslib.utils.text_util import get_prefix
 
 
 class TestStat(testcase.GsUtilIntegrationTestCase):
@@ -139,11 +140,11 @@ class TestStat(testcase.GsUtilIntegrationTestCase):
     self.assertIn('%s:' % suri(bucket_uri, 'notmissing'), stdout)
 
   def test_stat_bucket_wildcard(self):
-    bucket_uri = self.CreateBucket()
+    bucket_uri = self.CreateBucket(bucket_name_prefix=get_prefix())
     self.CreateObject(bucket_uri=bucket_uri, object_name='foo', contents=b'z')
-    stat_string = suri(bucket_uri) + '/?oo'
+    stat_string = suri(bucket_uri)[:-1] + '?/foo'
     self.RunGsUtil(['stat', stat_string])
-    stat_string2 = suri(bucket_uri) + '/*foo'
+    stat_string2 = suri(bucket_uri)[:-1] + '*/foo'
     self.RunGsUtil(['stat', stat_string2])
 
   def test_stat_object_wildcard(self):
--- a/gslib/tests/test_stat.py
+++ b/gslib/tests/test_stat.py
@@ -141,9 +141,9 @@ class TestStat(testcase.GsUtilIntegrationTestCase):
   def test_stat_bucket_wildcard(self):
     bucket_uri = self.CreateBucket()
     self.CreateObject(bucket_uri=bucket_uri, object_name='foo', contents=b'z')
-    stat_string = suri(bucket_uri)[:-1] + '?/foo'
+    stat_string = suri(bucket_uri) + '/?oo'
     self.RunGsUtil(['stat', stat_string])
-    stat_string2 = suri(bucket_uri)[:-1] + '*/foo'
+    stat_string2 = suri(bucket_uri) + '/*foo'
     self.RunGsUtil(['stat', stat_string2])
 
   def test_stat_object_wildcard(self):
--- a/gslib/tests/test_tabcomplete.py
+++ b/gslib/tests/test_tabcomplete.py
@@ -152,7 +152,7 @@ class TestTabComplete(testcase.GsUtilIntegrationTestCase):
     bucket_name = bucket_base_name + '-suffix'
     self.CreateBucket(bucket_name)
 
-    bucket_request = '%s://%s' % (self.default_provider, bucket_base_name)
+    bucket_request = '%s://%s' % (self.default_provider, bucket_name[:-2])
     expected_bucket_result = '//%s ' % bucket_name
 
     local_file = 'a_local_file'
--- a/gslib/tests/test_tabcomplete.py
+++ b/gslib/tests/test_tabcomplete.py
@@ -46,9 +46,9 @@ class TestTabComplete(testcase.GsUtilIntegrationTestCase):
   def test_single_bucket(self):
     """Tests tab completion matching a single bucket."""
 
-    bucket_name = self.MakeTempName('bucket')
-    # Workaround for XML API limitation, see PR 766 for details
-    self.CreateBucket(bucket_name, bucket_name_prefix='')
+    # Prefix is a workaround for XML API limitation, see PR 766 for details
+    bucket_name = self.MakeTempName('bucket', prefix='aaa-')
+    self.CreateBucket(bucket_name)
 
     request = '%s://%s' % (self.default_provider, bucket_name[:-2])
     expected_result = '//%s/' % bucket_name
--- a/gslib/tests/test_tabcomplete.py
+++ b/gslib/tests/test_tabcomplete.py
@@ -31,6 +31,7 @@ from gslib.tests.util import ARGCOMPLETE_AVAILABLE
 from gslib.tests.util import SetBotoConfigForTest
 from gslib.tests.util import unittest
 from gslib.tests.util import WorkingDirectory
+from gslib.utils.text_util import get_prefix
 from gslib.utils.boto_util import GetTabCompletionCacheFilename
 
 
@@ -47,7 +48,7 @@ class TestTabComplete(testcase.GsUtilIntegrationTestCase):
     """Tests tab completion matching a single bucket."""
 
     bucket_name = self.MakeTempName('bucket')
-    self.CreateBucket(bucket_name)
+    self.CreateBucket(bucket_name, bucket_name_prefix=get_prefix())
 
     request = '%s://%s' % (self.default_provider, bucket_name[:-2])
     expected_result = '//%s/' % bucket_name
@@ -59,7 +60,7 @@ class TestTabComplete(testcase.GsUtilIntegrationTestCase):
     """Tests bucket-only tab completion matching a single bucket."""
 
     bucket_name = self.MakeTempName('bucket')
-    self.CreateBucket(bucket_name)
+    self.CreateBucket(bucket_name, bucket_name_prefix=get_prefix())
 
     request = '%s://%s' % (self.default_provider, bucket_name[:-2])
     expected_result = '//%s ' % bucket_name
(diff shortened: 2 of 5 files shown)


gslib/tests/test_stat.py
```diff
@@ -141,9 +141,9 @@ def test_stat_one_missing_wildcard(self):
   def test_stat_bucket_wildcard(self):
     bucket_uri = self.CreateBucket()
     self.CreateObject(bucket_uri=bucket_uri, object_name='foo', contents=b'z')
-    stat_string = suri(bucket_uri)[:-1] + '?/foo'
+    stat_string = suri(bucket_uri) + '/?oo'
```
houglum: This test is called `test_stat_bucket_wildcard`, but you're removing the wildcard from the bucket portion of the URI :P

It's interesting that this fails when more than 1k buckets are present in the project -- the prefix we give before the wildcard is *the full bucket name*, so I'd assume that the listing call would use the bucket name a prefix for the listing call-- oooh, wait. That's right. The XML API doesn't have the `prefix` field for bucket listing requests:
https://cloud.google.com/storage/docs/xml-api/get-service

...like the JSON API does:
https://cloud.google.com/storage/docs/json_api/v1/buckets/list

Yay for interoperability with other poorly-designed storage providers' APIs :expressionless: 

...all that being said, we're stuck with this shortcoming of the XML API... but I can think of one disgusting hack to get around it.  The bucket listing is ordered lexicographically, so you're going to get the first 1000 buckets in nicely sorted order.  So for the buckets we use for these tests, we could just start their names with an alphabetically lesser value... I think `CreateBucket()` will create you a name starting with `gsutil-test-` if you don't supply it with one. So, let's prefix these buckets' names with something like `aaa-list-sort-`.  Should be doable via:

```
bucket_name = self.MakeTempname('bucket', prefix='aaa-list-sort-')
self.CreateBucket(bucket_name)
```

This applies here and in the other tests.

gslib/tests/test_stat.py
```diff
@@ -141,9 +141,9 @@ def test_stat_one_missing_wildcard(self):
...
     self.RunGsUtil(['stat', stat_string])
-    stat_string2 = suri(bucket_uri)[:-1] + '*/foo'
+    stat_string2 = suri(bucket_uri) + '/*foo'
```
houglum: Along with my comment above, please move the wildcard characters back into the bucket portion of these URIs.

gslib/tests/test_stat.py
catleeball: Thanks for all the context, and whoops, my mistake swapping the wildcard to the object! I'll edit up this PR. :)

catleeball: Ok, applied prefixes. Let me know if naming is ok, and if this looks like an ok spot for the new function to live. 👍

gslib/tests/testcase/integration_testcase.py
```diff
@@ -511,6 +513,8 @@ def CreateBucket(self,
           True.
       bucket_policy_only: If True, set the bucket's iamConfiguration's
           bucketPolicyOnly attribute to True.
+      name_prefix: Unicode string to be prepended to bucket_name
```
houglum: bucket_name_prefix, bucket_name_suffix

Also, it's worth noting whether these do or don't get added into the name if `bucket_name` is empty/not passed in.

gslib/tests/testcase/integration_testcase.py
```diff
@@ -527,8 +531,18 @@ def CreateBucket(self,
...
+    bucket_name_prefix = six.ensure_text(bucket_name_prefix)
+    bucket_name_suffix = six.ensure_text(bucket_name_suffix)
+
     if bucket_name:
+      bucket_name = ''.join([bucket_name_prefix,
+                             bucket_name,
+                             bucket_name_suffix])
       bucket_name = util.MakeBucketNameValid(bucket_name)
+    else:
+      bucket_name = self.MakeTempName(''.join([bucket_name_prefix,
```
houglum: I don't think this is semantically correct.  The arg you're passing corresponds to this method's `kind` parameter, so the "prefix" and "suffix" actually gets crammed into the middle of the generated name.  This method (defined in gslib/tests/testcase/base.py) also has a `prefix` parameter -- we should pass the prefix using that.  Similarly, it might be worth adding `suffix` as another kwarg and passing the suffix along that way.

gslib/utils/text_util.py
```diff
@@ -408,3 +409,23 @@ def get_random_ascii_chars(size, seed=0):
   contents = six.ensure_binary(contents)
   random.seed()  # Reset the seed for any other tests.
   return contents
+
+def get_prefix():
```
houglum: Nit: For new methods, we should use the `GoogleStyleGuideNamingStyle`, `not_the_python_one` :)

Also, this method name isn't super indicative of what the method does. How do you feel about calling this something like `GetLowStringSortingPrefix`?

gslib/utils/text_util.py
```diff
@@ -35,6 +35,7 @@
 from gslib.lazy_wrapper import LazyWrapper
 from gslib.utils.constants import UTF8
 from gslib.utils.constants import WINDOWS_1252
+from gslib.tests.util import USING_JSON_API
```
houglum: Please don't add test logic to non-test modules :)

gslib/utils/text_util.py
```diff
@@ -408,3 +409,23 @@ def get_random_ascii_chars(size, seed=0):
...
+
+  This creative workaround makes sure that buckets use in tab completion and wildcard tests
+  are hoisted to the top of the list of buckets. This workaround is primarily for our own CI
+  and tests may still flake if you have more than 1,000 buckets that would be ordered before
+  'aaa-' and are using the XML API.
+
+  Returns:
+    String literal 'aaa-' if currently using XML API.
+  """
+  if USING_JSON_API:
```
houglum: I think it's fine to always return 'aaa-', even for the JSON API, in favor of not adding test logic to non-test modules.  (If you disagree, you could add this logic to the callsites in the test code instead)

gslib/tests/testcase/integration_testcase.py
catleeball: Good catch! Will fix names and also make sure that it's prepended & appended when bucket names aren't specified.

gslib/tests/testcase/integration_testcase.py
catleeball: Whoops! Sorry, I should have checked the method signature, my bad. :) Will fix

gslib/utils/text_util.py
catleeball: Right! I'll pascal case it, and `GetLowStringSortingPrefix` sounds good!

gslib/utils/text_util.py
catleeball: Whoops, that makes sense! Will remove the import and grab that value manually

gslib/utils/text_util.py
catleeball: I'd be fine removing this method and doing this for the JSON API too! Applying it to the XML API only was trying to make sure it would take longer to hit >1000 buckets with `aaa-` prepended to them, but given the small set of tests this applies to, and the specific nature of this bug and workaround, going the simpler route and doing this regardless of API probably makes sense. :)

gslib/tests/testcase/integration_testcase.py
catleeball: It looks like if a name isn't supplied, the else block on (this revision's) line 542 should still give it the base name 'bucket' like before with prefix and suffix (once I've fixed the method semantics) :)

gslib/tests/testcase/base.py
```diff
@@ -102,8 +102,8 @@ def MakeTempName(self, kind, prefix=''):
       providers (e.g. replacing "_" with "-", converting uppercase letters to
       lowercase, etc.).
     """
-    name = '{prefix}gsutil-test-{method}-{kind}'.format(
-      prefix=prefix, method=self.GetTestMethodName(), kind=kind)
+    name = '{prefix}gsutil-test-{method}-{kind}{suffix}'.format(
+      prefix=prefix, method=self.GetTestMethodName(), kind=kind, suffix=suffix)
```
houglum: With all these additions, it's possible to go over the `MAX_BUCKET_LENGTH-9` character cutoff that we do below; it's possible that this completely chops off the suffix, breaking assumptions that the test might have based on the suffix of the bucket name.

We might want to have a sanity check here that fails if suffix was passed in and part of it was cut off, e.g.:

```
    name = '{prefix}gsutil-test-{method}-{kind}{suffix}'.format(
        prefix=prefix, method=self.GetTestMethodName(), kind=kind, suffix=suffix)
    if suffix and kind == 'bucket':
      # Chop off 9 characters so we can append a random string and
      # psuedo-randomize the bucket name and not go over the length limit.
      old_name = name
      name = name[:MAX_BUCKET_LENGTH-9]
      # Fail if we would have chopped off all or part of the suffix, as
      # test logic might depend on the suffix being present.
      if len(name) > len(old_name):
        self.fail(
            'Tried to create a psuedo-random bucket name with a specific '
            'suffix, but the generated name was too long and there was not '
            'enough room for the suffix. Please use shorter strings or perform '
            'name randomization manually.\nRequested name: %s\nName truncated '
            'to make room for random string: %s' % (old_name, name))

    name = '{name}-{rand}'.format(name=name, rand=self.MakeRandomTestString())
```

gslib/tests/test_tabcomplete.py
```diff
@@ -46,11 +46,11 @@ def setUp(self):
   def test_single_bucket(self):
     """Tests tab completion matching a single bucket."""
 
-    bucket_base_name = self.MakeTempName('bucket')
-    bucket_name = bucket_base_name + '-suffix'
-    self.CreateBucket(bucket_name)
+    bucket_name = self.MakeTempName('bucket')
+    # Workaround for XML API limitation, see PR 766 for details
+    self.CreateBucket(bucket_name, bucket_name_prefix='')
```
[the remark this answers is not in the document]
catleeball: Yes! Thanks for catching that

gslib/tests/testcase/base.py
```diff
@@ -88,7 +88,7 @@ def MakeRandomTestString(self):
     """Creates a random string of hex characters 8 characters long."""
     return '%08x' % random.randrange(256**4)
 
-  def MakeTempName(self, kind, prefix=''):
+  def MakeTempName(self, kind, prefix='', suffix=''):
```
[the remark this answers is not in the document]
catleeball: ack! 👍

gslib/tests/testcase/base.py
catleeball: Got it, that makes sense to me! Would it make sense if I put this logic in the `MakeBucketNameValid` method, since that gets called for `'bucket' type objects in this logic?

Likewise, would it be reasonable to also send the prefix and suffix to `MakeBucketNameValid` and then if the name is too long, automatically truncate from the base bucket name enough characters until the name is small enough, and only throw an error if `len(prefix) + len(suffix) >= MAX_BUCKET_LENGTH-12` or something? Or if that risks loosing too much entropy and causing bucket name collisions, I'll just do the logic you have above. :)

catleeball: Actually, I realized `MakeBucketName` isn't from the testing side of the library, I'll just do this here in base.py. 🙂

catleeball: Actually I just realized that the suffix is used exactly zero times when calling this method. Maybe it would be simplest if I just remove the suffix.

gslib/tests/testcase/base.py
```diff
@@ -106,8 +107,17 @@ def MakeTempName(self, kind, prefix=''):
       prefix=prefix, method=self.GetTestMethodName(), kind=kind)
     name = name[:MAX_BUCKET_LENGTH-9]
     name = '{name}-{rand}'.format(name=name, rand=self.MakeRandomTestString())
-    # As of March 2018, S3 no longer accepts underscores or uppercase letters in
-    # bucket names.
+    totalNameLen = len(name) + len(suffix)
```
houglum: s/totalNameLen/total_name_len

gslib/tests/testcase/base.py
```diff
@@ -106,8 +107,17 @@ def MakeTempName(self, kind, prefix=''):
...
+    if suffix:
+      if kind == 'bucket' and totalNameLen > MAX_BUCKET_LENGTH:
+        self.fail(
+          'Tried to create a psuedo-random bucket name with a specific '
+          'suffix, but the generated name was too long and there was not '
+          'enough room for the suffix. Please use shorter strings or perform '
+          'name randomization manually.\nRequested name: ' + name)
+      else:
```
houglum: the `else` isn't needed here; if the `if` is hit above, `self.fail()` will break the execution (can't remember if it returns or raises an exception of some sort...).

gslib/tests/testcase/base.py
```diff
@@ -106,8 +107,17 @@ def MakeTempName(self, kind, prefix=''):
...
-    # As of March 2018, S3 no longer accepts underscores or uppercase letters in
-    # bucket names.
+    totalNameLen = len(name) + len(suffix)
+    if suffix:
+      if kind == 'bucket' and totalNameLen > MAX_BUCKET_LENGTH:
+        self.fail(
+          'Tried to create a psuedo-random bucket name with a specific '
+          'suffix, but the generated name was too long and there was not '
+          'enough room for the suffix. Please use shorter strings or perform '
+          'name randomization manually.\nRequested name: ' + name)
```
houglum: Might also be worth including the suffix in the "requested name" part of the error message.

gslib/tests/testcase/base.py
catleeball: ack, fixed!

gslib/tests/testcase/base.py
catleeball: Fixed 👍

gslib/tests/testcase/base.py
catleeball: Makes sense! Fixed 🙂

gslib/tests/testcase/base.py
```diff
@@ -106,8 +107,16 @@ def MakeTempName(self, kind, prefix=''):
...
-    # As of March 2018, S3 no longer accepts underscores or uppercase letters in
-    # bucket names.
+    total_name_len = len(name) + len(suffix)
+    if suffix:
+      if kind == 'bucket' and totalNameLen > MAX_BUCKET_LENGTH:
+        self.fail(
+          'Tried to create a psuedo-random bucket name with a specific '
+          'suffix, but the generated name was too long and there was not '
+          'enough room for the suffix. Please use shorter strings or perform '
+          'name randomization manually.\nRequested name: ' + name + prefix)
```
houglum: s/name + prefix/name + suffix/

gslib/tests/testcase/base.py
```diff
@@ -106,8 +107,16 @@ def MakeTempName(self, kind, prefix=''):
       prefix=prefix, method=self.GetTestMethodName(), kind=kind)
     name = name[:MAX_BUCKET_LENGTH-9]
     name = '{name}-{rand}'.format(name=name, rand=self.MakeRandomTestString())
-    # As of March 2018, S3 no longer accepts underscores or uppercase letters in
-    # bucket names.
+    total_name_len = len(name) + len(suffix)
+    if suffix:
+      if kind == 'bucket' and totalNameLen > MAX_BUCKET_LENGTH:
```
houglum: s/totalNameLen/total_name_len/

gslib/tests/testcase/base.py
catleeball: whoops! ack, fixing

PR #768, opened 2019-05-03
Yapf format repository

catleeball:
Ran `pyformat -r -i -s 2 --exclude third_party,vendored .` followed by `yapf --style='{based_on_style: google, indent_width: 2}' -e '[vendored|third_party]' -irp .` on the gsutil repository to make it closer to compliant with the Google Python style guide.

Local tests all pass still, going to see if Kokoro picks up any issues when it runs against this PR.

Also added `yapf` to Travis CI to lint incoming PRs. The `-d` flag included will print the linted diff to stdout if there is any, and will return exit code 1 if there's a diff, alerting us to lint nits.


gslib/commands/cp.py
```diff
@@ -765,26 +764,20 @@
...
-                                   _CHECKSUM_VALIDATION_TEXT,
-                                   _RETRY_HANDLING_TEXT,
-                                   _RESUMABLE_TRANSFERS_TEXT,
-                                   _STREAMING_TRANSFERS_TEXT,
-                                   _SLICED_OBJECT_DOWNLOADS_TEXT,
-                                   _PARALLEL_COMPOSITE_UPLOADS_TEXT,
-                                   _CHANGING_TEMP_DIRECTORIES_TEXT,
-                                   _COPYING_SPECIAL_FILES_TEXT,
-                                   _OPTIONS_TEXT])
+_DETAILED_HELP_TEXT = '\n\n'.join([
```
houglum: Boooo. I'm assuming there's a YAPF option to keep long lists split up one-item-per-line like we had before?  The former was much cleaner to read and produced more readable diffs when altering the items.

I think this option is `SPLIT_ALL_COMMA_SEPARATED_VALUES`, and should be changed to True if we want to keep the former style.

catleeball:
Moving this to https://github.com/GoogleCloudPlatform/gsutil/pull/769 so the diff isn't tangled with the initial pyformat run I did in this PR since we're going with yapf.

PR #769, opened 2019-05-03
Add yapf linting

catleeball:
Added yapf configs and a line to .travis.yaml to run yapf every PR.

Also ran `yapf -irp .` on whole repository with the yapf configs added
in this PR.


catleeball:
Yapf seems to have done something weird, some tests in Kokoro are failing now, and in some versions of Python, Travis complains that Yapf had new nits. Investigating py2/3 compatible yapf, and the test failures.

catleeball:
It looks like the integ tests were another XML issue when over 1k buckets. Noted this, going to fix the additional XML wildcard flake.

Weirdly though, Travis does still find new yapf nits that locally it doesn't, and only when running in 3.x.

catleeball:
Latest commit I rewrote the print JSON lines in iam.py and it seems to have calmed down the Yapf confusion in Travis. :slightly_smiling_face:

Issue #770, opened 2019-05-04
When bucket name starts with a number: CommandException: Incorrect public member type for binding 


bbzg:
I had a rough few hours trying to figure this out, but I think I finally understand what is going on.

I used deployment manager to create a bucket with a random UUID prefix. Unfortunately for me the UUID started with a number!

I tried running this command:

```
gsutil iam ch allUsers:objectViewer gs://058612a7-ee32-41fe-9f9a-e57cae02b2c3-thing
```

Which resulted in this error

> CommandException: Incorrect public member type for binding gs://058612a7-ee32-41fe-9f9a-e57cae02b2c3-thing

When I deleted that bucket and created a new one that started with a letter, it worked.

So I have to assume that this is a bug in `gsutil iam ch` relating to buckets starting with a number.

catleeball:
Hi @bbzg !

Thanks for reporting this! I was able to reproduce this error and it looks like the command parsing in `gsutil/gslib/utils.py` in the `BindingStringToTuple` function is a little too sensitive. I'll poke around and see if I can fix the logic there to behave a little more nicely when buckets start with a number, and I'll keep this bug updated as we deploy a fix. :slightly_smiling_face: 

Thanks,
Lee

catleeball:
Digging around with the debugger, it looks like this is actually an issue with `gslib/commands/iam.py` https://github.com/GoogleCloudPlatform/gsutil/blob/cca0c5e9b4a576e1712f0178625426fb982bb7b1/gslib/commands/iam.py#L483

The regex here doesn't properly recognize the bucket URL as a bucket if it starts with a number.

catleeball:
There is a fix in the repository's `master` branch now and it should be included in our next release of gsutil. Thank you very much for flagging this issue to us!

bbzg:
No, thank you!

PR #772, opened 2019-05-06
Fix Iam Bucket URL Regex

catleeball:
Reported in bug #770 , iam operations fail on buckets whose names don't
contain any letters. For example, bucket with a UID name that contains
only digits and dashes will throw an exception, since the iam command
will try to parse it as a binding rather than a URL.

See also b/132080434

### DIFF
--- a/gslib/commands/iam.py
+++ b/gslib/commands/iam.py
@@ -53,7 +53,6 @@ from gslib.utils.iam_helper import IsEqualBindings
 from gslib.utils.iam_helper import PatchBindings
 from gslib.utils.iam_helper import SerializeBindingsTuple
 from gslib.utils.retry_util import Retry
-from gslib.utils.text_util import IsBucketNameValid
 
 _SET_SYNOPSIS = """
   gsutil iam set [-afRr] [-e <etag>] file url ...
@@ -216,6 +215,7 @@ _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _ch_help_text = CreateHelpText(_CH_SYNOPSIS, _CH_DESCRIPTION)
 
+STORAGE_URI_REGEX = re.compile(r'[a-z]+://.+')
 
 IAM_CH_CONDITIONS_WORKAROUND_MSG = (
     'To change the IAM policy of a resource that has bindings containing '
@@ -480,7 +480,7 @@ class IamCommand(Command):
     # expecting to come across the -r, -f flags here.
     it = iter(self.args)
     for token in it:
-      if IsBucketNameValid(token):
+      if STORAGE_URI_REGEX.match(token):
         patterns.append(token)
         break
       if token == '-d':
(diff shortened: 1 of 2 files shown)


catleeball:
Note, it looks like doing `gsutil mb` to make a bucket that begins with `-` or containing capital letters gets rejected in another part of gsutil, although the docs don't seem to mention these as restrictions on bucket names: https://cloud.google.com/storage/docs/naming#objectnames  Is this intentional? @houglum

gslib/commands/iam.py
```diff
@@ -480,7 +480,7 @@ def _PatchIam(self):
     # expecting to come across the -r, -f flags here.
     it = iter(self.args)
     for token in it:
-      if STORAGE_URI_REGEX.match(token):
+      if IsBucketNameValid(token):
```
houglum: I thought about this a bit afterward, and I think we should be doing minimal validation on the client side - if it's an invalid bucket name, URI, etc., then we should pass it along to the service via the appropriate API call and let the returned error message explain that the given resource name was invalid or didn't exist.

Thus, to check if something is a valid storage URI or not, we should just check for the presence of some `[a-z]` character(s), followed by `://`, followed by at least one more character. So basically what we had before, except change the regex to be `r'[a-z]+://.+'`.

wdyt?

catleeball: That sounds pretty reasonable to me since we only really need to distinguish the bucket strings from binding strings. Assuming bindings are always `:` delimited with two to three tokens, and bucket strings are always two tokens delimited by `://` is there any case where it would be valid for any part of the binding to start with `//`? Fearful for a case of falsely identifying a binding as a bucket. 

Particularly this might be a problem for the domain tag, where the example lists `domain:www.my-domain.org:objectViewer` someone might do `domain:https://my-domain.org:objectViewer` which would ID as a bucket string. I'm not sure if the latter is supposed to be valid binding syntax, but it seems like a reasonably possible thing for sometime to do and get a weird error from.

We know that neither the bucket prefix or URI will contain colons, so we could use your regex, plus a rule "contains no more than one colon" possibly. I'm testing and struggling to make a regex that satisfies that without producing partial matches.

We could combine them with something like:
```
r = re.compile(r'[a-z]+://.+')

if r.match(token) and token.count(':') < 2:
   do bucket stuff
```

or the regex is getting simple enough that we might be able to do without a regex at all:

```
if '://' in token and token.count(':') < 2:
    do bucket stuff
```

Do either of those two snippets sound reasonable?

catleeball: (made some edits to above comment)

houglum: I think you're thinking of `re.search()`; the above command uses `re.match()`, which only returns a match if there's an exact match starting from the beginning of the string; thus the beginning of the string would have to be the `[a-z]` chars, followed by `://`.  Something with a colon before the `://`, like `domain:foo.bar:someRole` wouldn't match. :)

catleeball: Oh oh! You're right! That is much simpler then. I'll add your regex and cut the function :)

catleeball: Ok! Put the new regex in, and pulled the validation function out. 👍

catleeball:
@houglum whoops, did I merge this one without an LGTM? very sorry! I was flipping through a few PRs and got a little too clicky with the merge button! Thank you for post-submit approval, sorry!

PR #774, opened 2019-05-07
Fixed integ test for 'cp -L' flag

catleeball:
For b/130055461. While doing Python 2/3 compat changes, we discovered
that the 'cp -L' flag had integration tests, but they failed to catch when headers
were wrapped in b' ' artifacts. This was because the asserts were only checking
a partial string slice. We fixed this by matching against the full source and dest.

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1746,9 +1746,14 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
 
     results = dict(zip(expected_headers, results))
 
-    self.assertEqual(results['Source'][:7], 'file://')
-    self.assertEqual(results['Destination'][:5],
-                     '%s://' % self.default_provider)
+    self.assertEqual(
+        results['Source'],
+        'file://' + fpath,
+    )
+    self.assertEqual(
+        results['Destination'],
+        dsturi,
+    )
 
     date_format = '%Y-%m-%dT%H:%M:%S.%fZ'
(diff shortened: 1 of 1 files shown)


gslib/tests/test_cp.py
```diff
@@ -1744,10 +1744,6 @@ def _ManifestUpload(self, file_name, object_name, manifest_name):
     self.assertEqual(expected_headers, lines[0].strip().split(','))
     results = lines[1].strip().split(',')
 
-    for header in results:
```
houglum: This did nothing, so I'm up for taking it out.  Changing the copy of the thing you're iterating over (the `header` variable, here) doesn't actually change the original thing.

```
>>> l = [1, 2, 3]
>>> for item in l:
...   l = 9
...
>>> l
[1, 2, 3]
```

So, that being said, I don't think this will actually catch any new problems.

catleeball: Good catch! I'll poke around a little more and figure out how the `b'value'` thing slipped by this integ test's asserts

catleeball: Oh whoops, I see that the assert is using a substring of the field: https://github.com/GoogleCloudPlatform/gsutil/blob/e1f413d43b1e7f4f119949b29f1acb2303c37123/gslib/tests/test_cp.py#L1749

I'll make it use the full field so it would include any weird artifacts. :)

catleeball: Ok, now it's asserting that the full source and dest buckets match!

PR #781, opened 2019-05-15
Adding location type to ls command and JSON API

thomasmaclean:
Since locationType is now present in the v1 API, just adding a field for buckets in the ls command, even if it isn't populated yet.

### DIFF
--- a/gslib/commands/ls.py
+++ b/gslib/commands/ls.py
@@ -420,8 +420,7 @@ class LsCommand(Command):
 
     text_util.print_to_fd(
         ('{bucket} :\n'
-         '\tStorage class:\t\t\t{storage_class}\n' +
-         location_type_line +
+         '\tStorage class:\t\t\t{storage_class}\n' + location_type_line +
          '\tLocation constraint:\t\t{location_constraint}\n'
          '\tVersioning enabled:\t\t{versioning}\n'
          '\tLogging configuration:\t\t{logging_config}\n'
--- a/gslib/commands/ls.py
+++ b/gslib/commands/ls.py
@@ -543,6 +543,7 @@ class LsCommand(Command):
             'iamConfiguration',
             'labels',
             'location',
+            'locationType',
             'logging',
             'lifecycle',
             'metageneration',
--- a/gslib/commands/ls.py
+++ b/gslib/commands/ls.py
@@ -337,10 +337,12 @@ class LsCommand(Command):
     # We're guaranteed by the caller that the root object is populated.
     bucket = bucket_blr.root_object
     location_constraint = bucket.location
+    location_type = bucket.locationType
     storage_class = bucket.storageClass
     fields = {
         'bucket': bucket_blr.url_string,
         'storage_class': storage_class,
+        'location_type': location_type,
         'location_constraint': location_constraint,
         'acl': AclTranslation.JsonFromMessage(bucket.acl),
         'default_acl': AclTranslation.JsonFromMessage(bucket.defaultObjectAcl),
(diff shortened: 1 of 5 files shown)


gslib/commands/ls.py
```diff
@@ -416,6 +418,7 @@ def _PrintBucketInfo(self, bucket_blr, listing_style):
     text_util.print_to_fd(
         ('{bucket} :\n'
          '\tStorage class:\t\t\t{storage_class}\n'
+         '\tLocation type:\t\t\t{location_type}\n'
```
houglum: Is this field going to be present in XML API responses too? If not, let's not bother printing `Location type: None` if the XML API was used (we do the same thing for other JSON-only fields - see `time_updated_line` and other `<foo>_line` variables).

gslib/commands/ls.py
```diff
@@ -337,10 +337,12 @@ def _PrintBucketInfo(self, bucket_blr, listing_style):
     # We're guaranteed by the caller that the root object is populated.
     bucket = bucket_blr.root_object
     location_constraint = bucket.location
+    location_type = bucket.locationType
     storage_class = bucket.storageClass
     fields = {
         'bucket': bucket_blr.url_string,
         'storage_class': storage_class,
+        'location_type': location_type,
```
houglum: Can you add some logic to the long-listing tests in `gslib/tests/test_ls.py` to make sure this field shows up as expected?

thomasmaclean: I may just need to wait on this one as the [API has the field available](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest) but it isn't populating yet, apparently, so the  I'll circle back once this field starts populating.

gslib/commands/ls.py
thomasmaclean: Good call. I'm fixing that now.

thomasmaclean:
@catleeball Thanks for the note! My changes are in `gslib/third_party/storage_apitools/storage_v1_messages.py` rather than `third_party/apitools`. Sorry if I overlooked it but I couldn't find any trace of the `storage_v1_messages.py` in the `apitools` repo. Any guidance would be appreciated, thanks!

catleeball:
@thomasmaclean Oops, that's my mistake! I misread the path to the file you modified and thought it was in the submodule. Please disregard my last comment, this is a good location for your change! :sweat_smile:

gslib/commands/ls.py
houglum: Sounds good.

thomasmaclean:
@catleeball @houglum bigstore-gsutil-testing is now whitelisted for location type, so the new integration tests pass (hopefully kokoro confirms this). Let me know if you want this to go into this release or wait.

houglum:
@thomasmaclean I'd prefer not to introduce too many new things into gsutil until we can get the Py 2-and-3 release out -- we're ironing out a few compat issues we found during tests for the 4.39-prerelease build, and I'd like to stick with the stuff there, get everything working, and throw new functionality/features into a subsequent version.

thomasmaclean:
@houglum knowing there's another release coming, I believe this should be prioritized

houglum:
I went meta and filed a pull request against your pull request :upside_down_face: 

This new test was failing against the Kokoro build for the linux_py27_xml suite:
http://sponge/target?id=0b6312f8-57f1-44b1-adcb-aeec3a35831c&target=cloud_storage_gsutil%2Flinux%2Fpy27_xml&searchFor=

(I think the ACL tests that failed in that log have been fixed since you filed this PR;  I didn't see anything else in there that was failing, and when I ran the full test suite against 2.7, both against JSON and XML APIs, on my workstation, everything passed after I pulled in the changes from my PR)

Please merge that into your branch, at which point the code should also be updated in this PR.

thomasmaclean:
Thanks @houglum! Sorry I missed this on my first pass. Merged.

PR #784, opened 2019-05-22
Small doc edits

starsandskies:
* Add the -j flag to the relevant example in `gsutil cp`.
* Clarify that `-m` is a top-level flag, not specific to `gsutil cp`.
* Offer links to additional conceptual information for the Bucket Policy Only and GCS-specific service account.
* These changes address the following user-reported issues: b/75005961, b/111501119, b/112908799, and b/131860498.

### DIFF
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -97,8 +97,8 @@ _DESCRIPTION_TEXT = """
     gsutil cp -r dir gs://my-bucket
 
   If you have a large number of files to transfer you might want to use the
-  gsutil -m option, to perform a parallel (multi-threaded/multi-processing)
-  copy:
+  top-level gsutil -m option (see "gsutil help command_opts"), to perform a
+  parallel (multi-threaded/multi-processing) copy:
 
     gsutil -m cp -r dir gs://my-bucket
 
@@ -252,8 +252,8 @@ _COPY_IN_CLOUD_TEXT = """
 
     gsutil cp -A gs://bucket1/obj gs://bucket2
 
-  The gsutil -m flag is disallowed when using the cp -A flag, to ensure that
-  version ordering is preserved.
+  The top-level gsutil -m flag is disallowed when using the cp -A flag, to
+  ensure that version ordering is preserved.
 """
 
 _CHECKSUM_VALIDATION_TEXT = """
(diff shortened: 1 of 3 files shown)


gslib/commands/cp.py
```diff
@@ -605,7 +605,7 @@
 
                    gsutil -o "GSUtil:parallel_process_count=8" \\
                      -o "GSUtil:parallel_thread_count=1" \\
-                     -m cp -r /local/source/dir gs://bucket/path
+                     -m cp -j html -r /local/source/dir gs://bucket/path
```
catleeball: Not sure if this is a doc bug or an implementation bug, but with gsutil 4.38 `-j` will match the given pattern for any part of the filename, not just the extension. For example:

```
cball@cball:~/testo$ ls
a.txt  b.txt  txt.jpg

cball@cball:~/testo$ gsutil -o "GSUtil:parallel_process_count=8" -m cp -j txt -r ./ gs://do-not-delete-cball-test
Copying file://./a.txt [Content-Type=text/plain]...
Copying file://./b.txt [Content-Type=text/plain]...
Copying file://./txt.jpg [Content-Type=image/jpeg]...
/ [3/3 files][    0.0 B/    0.0 B]

cball@cball:~/testo$ gsutil ls gs://do-not-delete-cball-test
gs://do-not-delete-cball-test/a.txt
gs://do-not-delete-cball-test/b.txt
gs://do-not-delete-cball-test/txt.jpg
```

Should I open a bug for this and update the `-j` implementation, or is the doc inaccurate?

CC @houglum

starsandskies: This is definitely something Matt should weigh in on, but how do you know that gzip transport encoding was applied to txt.jpg in your example?

catleeball: @starsandskies oops! I was confusing things, for some reason I was thinking that non-matches from `-j` would be excluded entirely from the copy. You're right that gzip encoding isn't applied to `txt.jpg` when running with `-D`. https://paste.googleplex.com/5962524765716480

Nevermind then! This PR looks good to me. 🙂

starsandskies: Awesome! Glad we're good to go :D

gslib/commands/cp.py
```diff
@@ -97,8 +97,8 @@
     gsutil cp -r dir gs://my-bucket
 
   If you have a large number of files to transfer you might want to use the
-  gsutil -m option, to perform a parallel (multi-threaded/multi-processing)
-  copy:
+  top-level gsutil -m option (see "gsutil help command_opts"), to perform a
```
houglum: This doesn't work; should be `gsutil help options`.

houglum: @starsandskies

starsandskies: Thanks Matt, I'll put the fix in!

PR #786, opened 2019-05-25
Improve logic for detecting PKCS 12 keys

catleeball:
When attempting to load a PKCS 12 key in Python 3, `io.open` would
throw an error since it tried to apply UTF-8 text encoding to the
key file, which was not a compatible text file.

Changes tested using .p12 key from our pre-release tests.

### DIFF
--- a/gslib/gcs_json_credentials.py
+++ b/gslib/gcs_json_credentials.py
@@ -25,6 +25,7 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import base64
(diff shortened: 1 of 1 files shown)


catleeball:
Very similar changes were also added to https://github.com/GoogleCloudPlatform/gcs-oauth2-boto-plugin/pull/45

Once the plugin change above is merged, I'll update the submodule in gsutil to point to the new head

PR #796, opened 2019-06-06
Test `help` command for all commands.

houglum:
This will also be run against new commands added in the future, since
the list of commands is obtained by a call to `Command.__subclasses__()`.

Context: I found that the new hmac command failed when I tried to run
the help command against it.


catleeball:
Similar to #781 , should we squeeze this into our upcoming bugfix release?

PR #797, opened 2019-06-06
Revert PR #792

catleeball:
This reverts the change in PR #792 

This seems to cause tabcomplete tests to fail locally and on Kokoro. With the dependency compatibility updates, tabcomplete tests locally and in prerelease pass when this line is removed. Previously pre-release tests were unable to find the local interpreter in some circumstances; with the dependency updates, this doesn't seem to occur anymore.


houglum:
Let's cancel this PR, as we've merged in #798 instead.

PR #798, opened 2019-06-06
Add debugging mechanisms and fix tabcomplete tests.

houglum:
We started running tabcomplete tests by specifying the interpreter used
to invoke gsutil; this ended up causing argcomplete to provide no
matches, as it was trying to find a Completer for something like "python
gsutil ls" (which didn't exist) rather than "gsutil ls" (which existed).

This is the 4th time I've had to do "clever" things to see what
argcomplete was printing out for me in order to debug test failures, so
I added in a boolean in the test code to toggle the add'l debug prints.

### DIFF
--- a/gslib/tests/testcase/integration_testcase.py
+++ b/gslib/tests/testcase/integration_testcase.py
@@ -1005,21 +1005,49 @@ class GsUtilIntegrationTestCase(base.GsUtilTestCase):
       expected_results: The expected tab completion results for the given input.
     """
     cmd = [gslib.GSUTIL_PATH] + ['--testexceptiontraces'] + cmd
-    cmd = [str(sys.executable)] + cmd if not InvokedFromParFile() else cmd
+    if InvokedFromParFile():
+      argcomplete_start_idx = 1
+    else:
+      argcomplete_start_idx = 2
+      # Prepend the interpreter path; ensures we use the same interpreter that
+      # was used to invoke the integration tests. In practice, this only differs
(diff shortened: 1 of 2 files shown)


gslib/tests/testcase/integration_testcase.py
```diff
@@ -1005,21 +1005,49 @@ def RunGsUtilTabCompletion(self, cmd, expected_results=None):
...
+      # see any output you can get. I've had to do this so many times that I'm
+      # just going to leave this in the code for convenience ¯\_(ツ)_/¯
+      #
+      # If set, this will print out extra info from the argcomplete subprocess.
+      # You'll probably want to find one test that's failing and run it
+      # individually, e.g.:
+      #   python3 ./gsutil test tabcomplete.TestTabComplete.test_single_object
+      # so that only one subprocess is run, thus routing the output to your
+      # local terminal rather than swallowing it.
+      _hacky_debugging = False
```
catleeball: Given the struggle of figuring this out, I appreciate that you left in the debugging logic! Could we gate it with the `-D` flag rather than a static boolean?

catleeball: (If wiring it to `-D` is a hassle, then no worries, it's honestly nbd)

catleeball: nevermind, we chatted offline, this wouldn't make sense for `-D` because it will make the test fail

PR #803, opened 2019-06-12
Enable gsutil to specify a request reason either from http header or environment variable, tracking bug: b/131747209

jingwgg:
…environment variable

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -15,8 +15,8 @@
 """Integration tests for cp command."""
 
 from __future__ import absolute_import
-from __future__ import print_function
 from __future__ import division
+from __future__ import print_function
 from __future__ import unicode_literals
 
 import base64
@@ -25,6 +25,7 @@ import datetime
 import gslib
 import gzip
 from hashlib import md5
+import io
 import logging
 import os
 import pickle
@@ -34,7 +35,6 @@ import re
 import string
 import sys
 import threading
-import io
 
 import six
 from six.moves import http_client
@@ -1001,6 +1001,42 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
       self.assertNotRegex(stdout, r'image/gif,\s*image/gif')
 
     _Check2()
+
+  @SkipForXML('No compressed transport encoding support for the XML API.')
+  @SequentialAndParallelTransfer
+  def test_request_reason_header(self):
+    """Tests that x-goog-request-reason is set successfully on copy."""
+    bucket_uri = self.CreateBucket()
+    dst_uri = suri(bucket_uri, 'foo')
+    fpath = self._get_test_file('test.gif')
+
+    self.RunGsUtil(['-h', 'X-Goog-Request-Reason:b/this_is_reason', 'cp',
+                    fpath, dst_uri])
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_reason')
+    dst_uri2 = suri(bucket_uri, 'bar')
+    self.RunGsUtil(['cp', dst_uri, dst_uri2])
+    # Ensure metadata was preserved across copy.
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri2], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_reason')
+
+  @SkipForXML('No compressed transport encoding support for the XML API.')
+  @SequentialAndParallelTransfer
+  def test_request_reason_set_by_environment_var(self):
+    """Tests that request reason can be set successfully using the environment
+    variable CLOUDSDK_CORE_REQUEST_REASON."""
+    os.environ['CLOUDSDK_CORE_REQUEST_REASON'] = 'b/this_is_env_reason'
+    bucket_uri = self.CreateBucket()
+    dst_uri = suri(bucket_uri, 'foo')
+    fpath = self._get_test_file('test.gif')
+    self.RunGsUtil(['cp', fpath, dst_uri])
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
+    dst_uri2 = suri(bucket_uri, 'bar')
+    self.RunGsUtil(['cp', dst_uri, dst_uri2])
+    # Ensure metadata was preserved across copy.
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri2], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
 
   @SequentialAndParallelTransfer
   def test_other_headers(self):
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1033,10 +1033,13 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
     stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
     self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
     dst_uri2 = suri(bucket_uri, 'bar')
-    self.RunGsUtil(['cp', dst_uri, dst_uri2])
+    # test when both environment variable and http header is set,
+    # use the reason specified in http header
+    self.RunGsUtil(['-h', 'X-Goog-Request-Reason:b/this_is_reason', 'cp',
+                    dst_uri, dst_uri2])
     # Ensure metadata was preserved across copy.
     stdout = self.RunGsUtil(['ls', '-L', dst_uri2], return_stdout=True)
-    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_reason')
 
   @SequentialAndParallelTransfer
   def test_other_headers(self):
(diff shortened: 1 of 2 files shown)


gslib/tests/test_cp.py
```diff
@@ -1002,6 +1002,42 @@ def _Check2():
...
+    variable CLOUDSDK_CORE_REQUEST_REASON."""
+    os.environ['CLOUDSDK_CORE_REQUEST_REASON'] = 'b/this_is_env_reason'
+    bucket_uri = self.CreateBucket()
+    dst_uri = suri(bucket_uri, 'foo')
+    fpath = self._get_test_file('test.gif')
+    self.RunGsUtil(['cp', fpath, dst_uri])
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
+    dst_uri2 = suri(bucket_uri, 'bar')
+    self.RunGsUtil(['cp', dst_uri, dst_uri2])
```
houglum: I don't think we need this second part if we're already testing it in `test_request_reason_header()`.

gslib/tests/test_cp.py
```diff
@@ -1002,6 +1002,42 @@ def _Check2():
...
+    # Ensure metadata was preserved across copy.
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri2], return_stdout=True)
+    self.assertRegex(stdout, r'X-Goog-Request-Reason:b/this_is_env_reason')
+
```
houglum: I think you'll also want a test for the case where both (-h arg, and via env var) are supplied.  Either a unit test (if you factor out your code in the other file into a function) or another integration test

catleeball:
Thank you very much for this PR!

Tiny nit: If there are any relevant buganizer bugs, would you mind listing the bug numbers in the PR? Just as a nice way to track between bugs and PRs. 🙂

gslib/tests/test_cp.py
jingwgg: Thanks for the suggestion. I change the second part to test when both the environment variable and http header is set, the request reason comes from the http header

gslib/tests/test_cp.py
jingwgg: Done.

jingwgg:
Done

jingwgg:
> Thank you very much for this PR!
> 
> Tiny nit: If there are any relevant buganizer bugs, would you mind listing the bug numbers in the PR description? Just as a nice way to track between bugs and PRs. 🙂

Done. Thanks for the suggestion.

houglum:
Looks good.  We've already done the prerelease tests for the current release though, so I'm going to wait until after the release to merge this PR.

jingwgg:
> Looks good. We've already done the prerelease tests for the current release though, so I'm going to wait until after the release to merge this PR.

SG. Thank you!

gslib/tests/test_cp.py
```diff
@@ -1002,6 +1002,44 @@ def _Check2():
 
     _Check2()
 
+  @SkipForXML('No compressed transport encoding support for the XML API.')
```
houglum: Actually, I have one more questions before I merge this -- it looks like the skip explanation message here is for tests that use transport encoding... why are we actually skipping this for the XML API?   Did you intend to skip this for the XML API and just forget to change the message, or did you copy/paste this and not forget to remove the skip decorator entirely?

houglum: FYI, if I remove this decorator, the test fails when using the XML API.  Looks like you might just need to tweak the regex.  The metadata shows up like this when doing an `ls -L` using the JSON API:

```
    Metadata:               
        X-Goog-Request-Reason: some-reason-here
```

but shows up like this when using the XML API:

```
    Metadata:               
        x-goog-request-reason:some-reason-here
```

jingwgg: Thank you for noticing that. I add that decorator, because as you know, the test fails when using the XML API. 
Actually I don't find the output of metadata when using the XML API

gslib/tests/test_cp.py
```diff
@@ -1002,6 +1002,42 @@ def _Check2():
...
+  def test_request_reason_set_by_environment_var(self):
+    """Tests that request reason can be set successfully using the environment
+    variable CLOUDSDK_CORE_REQUEST_REASON."""
+    os.environ['CLOUDSDK_CORE_REQUEST_REASON'] = 'b/this_is_env_reason'
+    bucket_uri = self.CreateBucket()
+    dst_uri = suri(bucket_uri, 'foo')
+    fpath = self._get_test_file('test.gif')
+    self.RunGsUtil(['cp', fpath, dst_uri])
+    stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
+    self.assertRegex(stdout.lower(), r'x-goog-request-reason:b/this_is_env_reason')
```
houglum: Note that the output has a space between the `:` and the `b` in the JSON API output, but not the XML API output.  This regex should work without needing to convert all of `stdout` to lowercase:

```
self.assertRegex(stdout, r'[Xx]-[Gg]oog-[Rr]equest-[Rr]eason:( )?b/this_is_env_reason')
```

gslib/tests/test_cp.py
```diff
@@ -1002,6 +1002,42 @@ def _Check2():
...
+
+  @SequentialAndParallelTransfer
+  def test_request_reason_set_by_environment_var(self):
+    """Tests that request reason can be set successfully using the environment
+    variable CLOUDSDK_CORE_REQUEST_REASON."""
+    os.environ['CLOUDSDK_CORE_REQUEST_REASON'] = 'b/this_is_env_reason'
+    bucket_uri = self.CreateBucket()
+    dst_uri = suri(bucket_uri, 'foo')
+    fpath = self._get_test_file('test.gif')
+    self.RunGsUtil(['cp', fpath, dst_uri])
```
houglum: It turns out that the `cp` operation doesn't work under the XML API, which is why this test fails when run using XML.  (Note that the `ls` command still works, but the output is slightly different, as I mentioned below)

When I run this command with debug output enabled, it looks like gsutil is sending the header:

```
$ $ ./gsutil -D -o 'GSUtil:prefer_api=xml' -h 'x-goog-request-reason: b/1' cp <(echo testingcontents) gs://do-not-delete-lklk/testreqreasonxml
[...]
send: u'PUT /do-not-delete-lklk/testreqreasonxml HTTP/1.1\r\nHost: storage.googleapis.com\r\nAccept-Encoding: identity\r\nX-Goog-Request-Reason:  b/1\r\nExpect: 100-Continue\r\ncontent-language: en\r\nTransfer-Encoding: chunked\r\nx-goog-api-version: 1\r\nContent-Type: application/octet-stream\r\nAuthorization: <REDACTED>
[...]
```

Do we want to hold off on merging this until the XML API accepts and correctly stores this header?  If so, please point me to an internal tracking bug number for this.

jingwgg: > It turns out that the `cp` operation doesn't work under the XML API, which is why this test fails when run using XML. (Note that the `ls` command still works, but the output is slightly different, as I mentioned below)
> 
> When I run this command with debug output enabled, it looks like gsutil is sending the header:
> 
> ```
> $ $ ./gsutil -D -o 'GSUtil:prefer_api=xml' -h 'x-goog-request-reason: b/1' cp <(echo testingcontents) gs://do-not-delete-lklk/testreqreasonxml
> [...]
> send: u'PUT /do-not-delete-lklk/testreqreasonxml HTTP/1.1\r\nHost: storage.googleapis.com\r\nAccept-Encoding: identity\r\nX-Goog-Request-Reason:  b/1\r\nExpect: 100-Continue\r\ncontent-language: en\r\nTransfer-Encoding: chunked\r\nx-goog-api-version: 1\r\nContent-Type: application/octet-stream\r\nAuthorization: <REDACTED>
> [...]
> ```
> 
> Do we want to hold off on merging this until the XML API accepts and correctly stores this header? If so, please point me to an internal tracking bug number for this.

We don't have another bug for just XML API, I guess I should fix it in this pull request. Actually, can you point me to the code where it is related to XML API, I have hard time to identify where to change for XML API. Thanks!

houglum: Per our offline thread, we determined that this should **not** be setting the request reason as custom metadata on the object.  This should be used only for auth'z purposes, and not preserved in the metadata.  We should instead be testing that:

1) If this header is supplied for a `cp` command, the resulting object's custom metadata should **not** contain the request reason
2) The header should be included along with the request.  We can probably check for this by running the command in debug mode, e.g. `gsutil -D cp src dst` and checking for the header in the output of stderr.

houglum: I think we probably want to change this PR to:

- Only accept the request reason from an environment variable
- [JSON API] Change gslib/gcs_json_api.py such that all when all requests are made, they're wrapped in a  `_ApitoolsRequestHeaders()` context manager, and that context manager method is the place where we should be checking the environment variable for a request reason and adding it to the headers if it's present.
- [XML API] Change gslib/boto_translation.py to do the same thing; the method that creates headers for this API client is `_CreateBaseHeaders()`.

houglum: I think we probably want to change this PR to:

- Only accept the request reason from an environment variable
- [JSON API] Change gslib/gcs_json_api.py such that when we construct an api_client, the `additional_http_headers` object checks for the request reason and adds it to that dict if present
- [XML API] Change gslib/boto_translation.py to do something similar; before we send each request, we construct the base headers -- the method that creates headers for this API client is `_CreateBaseHeaders()`.  In that method, we should check for the request reason env var, and set it in the headers dict if present.

houglum: I started a new branch with these things done; I'll leave the test writing to you :)

I'll go ahead and close this PR; to build on top of the changes I checked in, check out the `req-reason-header` branch and make your changes on top of that; push them to a branch in your fork, and send a PR to merge your changes into the `req-reason-header` branch.  Then we can merge that whole branch into master when everything is finished.

jingwgg: Thank you very much, will do!

PR #804, opened 2019-06-13
4.39 release setup

houglum:
Hey @catleeball - here are the commits I want to add before doing the 4.39 release.  As discussed, we're not going to get Py3 support into this version, so I've added a quick commit (which I'll undo after the release) to prevent users from accidentally trying to run this under Py3.  Please let me know if I missed anything.


catleeball:
That looks like all of the blockers we took out iirc! Should be good to block folks from accidentally using Py3 yet. :)

LGTM ship it! ⛵️

houglum:
Thanks for the quick review!

Note that the Travis tests failed, but only on the Py3 envs due to them trying to run the thing under Python 3 :)

PR #805, opened 2019-06-16
Update rsync.py

BobiGilburd:
Fixed a bug in rsync when the file name contains spaces (would throw an exception if encounter such a file). The fix is to split only the rest of the arguments (changed split to rsplit with exact number of arguments).

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -1046,7 +1046,7 @@ class _DiffIterator(object):
                      crc32c, md5)
     """
     (encoded_url, size, time_created, atime, mtime, mode, uid, gid, crc32c,
-     md5) = line.split()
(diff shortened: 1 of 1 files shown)


BobiGilburd:
I signed it!

catleeball:
Note to internal folks, Kokoro test logs on 2.7 platforms listed here, waiting for completion: https://paste.googleplex.com/6396482808709120

catleeball:
2.7 Kokoro tests came back good, going to squash and merge. Thank you again @BobiGilburd !

houglum:
FYI, the bigger issue was the removal of a `quote_plus()` call in the synchronization stage of the rsync command.  This was fixed in #835

houglum:
(rsync works properly again in v4.40)

Issue #807, opened 2019-06-17
New integration test: rsync filenames with spaces


catleeball:
A bug reached our production release where rsync failed on filenames with spaces due to a rogue `str.split()`. We should add a testcase that includes filenames with spaces.

Refs:
PR: https://github.com/GoogleCloudPlatform/gsutil/pull/805
Bug: https://github.com/GoogleCloudPlatform/gsutil/issues/806

PR #810, opened 2019-06-17
Add file with whitespace in name to rsync tests

catleeball:
For bugs:
https://github.com/GoogleCloudPlatform/gsutil/issues/807
b/135474885

Added a test case whith a file containing whitespace. A regression was
found that rsync tests didn't catch where split was called and
improperly parsed arguments when a file with whitespace was present.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -2835,7 +2835,6 @@ class TestRsync(testcase.GsUtilIntegrationTestCase):
 
     _Check()
 
-
   def test_rsync_files_with_whitespace(self):
     """Test to ensure filenames with whitespace can be rsynced"""
     filename = 'foo bar baz.txt'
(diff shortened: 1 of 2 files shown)


gslib/tests/test_rsync.py
```diff
@@ -2835,6 +2835,26 @@ def _Check():
...
+    filename = 'foo bar baz.txt'
+    local_uris = []
+    bucket_uri = self.CreateBucket()
+    tmpdir = self.CreateTempDir()
+    contents = 'File from rsync test: test_rsync_files_with_whitespace'
+    local_file = self.CreateTempFile(tmpdir, contents, filename)
+
+    expected_list_results = frozenset(['/foo bar baz.txt'])
+
+    """Tests rsync works as expected."""
```
houglum: Make this a regular pound-sign-style comment please :)

catleeball: Whoops! Fixed. :)

Issue #809, opened 2019-06-17
Update command is broken for v4.39


houglum:
Originally reported in https://issuetracker.google.com/issues/135470907.

We added new yapf config files since the last release.  When users update to v4.39, it pulls in those new files; subsequent runs of the update command will fail because gsutil is detecting what it thinks is "user data" in the top-level gsutil directory.

houglum:
The resulting error looks like:

```
$ ./gsutil update -f -n
CommandException: A file (/path/to/gsutil/gsutil/.style.yapf) that is
not distributed with gsutil was found in the gsutil directory. The
update command cannot run with user data in the gsutil directory.
```

houglum:
I fixed this with https://github.com/GoogleCloudPlatform/gsutil/pull/808, which will be included in the v4.40 release.

The downside here is that anyone who's using a standalone tarball/zip-file install of gsutil, and has updated to v4.39, will be stuck with a broken `gsutil update` command until they delete `.style.yapf` and `.yapfignore` from the root of their gsutil directory.

houglum:
We just released 4.40, which included a fix to ignore those new top-level files.

PR #811, opened 2019-06-17
Include all directory files in `update` test

catleeball:
The integration test for the update command copied only expected release
files when testing. This can miss regressions like in issue #809 .
Updated test logic to include everything in the gsutil directory except
the .git directory.

### DIFF
--- a/gslib/tests/test_update.py
+++ b/gslib/tests/test_update.py
@@ -84,13 +84,12 @@ class UpdateTest(testcase.GsUtilIntegrationTestCase):
     # working files left in top-level directory by gsutil developers (like tags,
     # .git*, etc.)
     os.makedirs(gsutil_dst)
-    for comp in ('CHANGES.md', 'CHECKSUM', 'gslib', 'gsutil', 'gsutil.py',
-                 'LICENSE', 'MANIFEST.in', 'README.md', 'setup.py',
-                 'third_party', 'VERSION'):
-      cp_src_path = os.path.join(GSUTIL_DIR, comp)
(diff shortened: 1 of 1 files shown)


gslib/tests/test_update.py
```diff
@@ -84,13 +84,12 @@ def test_update(self):
...
     # .git*, etc.)
     os.makedirs(gsutil_dst)
-    for comp in ('CHANGES.md', 'CHECKSUM', 'gslib', 'gsutil', 'gsutil.py',
-                 'LICENSE', 'MANIFEST.in', 'README.md', 'setup.py',
-                 'third_party', 'VERSION'):
-      cp_src_path = os.path.join(GSUTIL_DIR, comp)
-      cp_dst_path = os.path.join(gsutil_dst, comp)
-      func = shutil.copytree if os.path.isdir(cp_src_path) else shutil.copyfile
-      func(cp_src_path, cp_dst_path)
+    for comp in os.listdir(GSUTIL_DIR):
```
houglum: It seems like we should add something to gslib/utils/update_util.py to get the list of expected files for an update (as parsed from the manifest), along with extra entries for the additional files that we know to be okay (e.g. .travis.yml).  And both the update command and this test should use that method.

catleeball: Understood! Looking at update_util.py I can see how that additional function makes sense there, plus we could unit test the function directly, in addition to this integ test.

catleeball:
After talking with matt, we're tabling the integ test for the moment until a possibly later date when we have our manifest and ignore lists in a more consolidated state.

PR #814, opened 2019-06-19
Replace sys.stderr.write with print_to_fd

catleeball:
For Python 2/3 compatibility, print_to_fd offers us more flexibility
when trying to write various string types to an arbitrary file
descriptor.

Also added get_utf8able_str function to try to ensure strings being
printed are in the right encoding, which cropped up a few times in
recent pre-release test runs where the deubg flag is included.

### DIFF
--- a/gslib/utils/text_util.py
+++ b/gslib/utils/text_util.py
@@ -404,3 +404,35 @@ def get_random_ascii_chars(size, seed=0):
   contents = six.ensure_binary(contents)
   random.seed()  # Reset the seed for any other tests.
   return contents
+
+
+def get_utf8able_str(s, errors='strict'):
+    """Returns a UTF8-encodable string in PY3, UTF8 bytes in PY2.
+
+    This method is similar to six's `ensure_str()`, except it also
+    makes sure that any bytes passed in can be decoded using the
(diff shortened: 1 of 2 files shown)


gslib/utils/text_util.py
```diff
@@ -404,3 +404,35 @@ def get_random_ascii_chars(size, seed=0):
   contents = six.ensure_binary(contents)
   random.seed()  # Reset the seed for any other tests.
   return contents
+
+
+def get_utf8able_str(s, errors='strict'):
```
houglum: At the call sites, can we just use the boto version of this method? It's already defined in gslib/vendored/boto/boto/utils.py, and we vendor that library now, so we control that code. No sense in defining it twice, assuming the places that need this method are okay with also importing boto.  (If using that at any call sites would result in a dependency cycle, I'm fine with duplicating it here as long as we leave a comment saying that this method should be kept in sync with the one in gslib/vendored/boto/boto/utils.py.)

catleeball: Pretty sure that won't create any dependency cycles! I'll just call it from boto, good idea. :)

catleeball: Okay, pushed the change to this PR

PR #816, opened 2019-06-20
Skip a test that's currently failing for...

houglum:
...some of our test projects. Details, for Google devs, are in
https://issuetracker.google.com/issues/135700569.


houglum:
@thomasmaclean for FYI. Assigning the internal bug to you; for now, we're just going to skip this test.

PR #822, opened 2019-06-22
Made resumable upload test use bigger file

catleeball:
Sometimes the resumable upload test would fail because it couldn't
artifically halt the test before finishing uploading the entire test
file. Made the test file much larger with more entropy to help avoid
this race condition.

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -3807,7 +3807,8 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
 
   def start_over_error_test_helper(self, http_error_num):
     bucket_uri = self.CreateBucket()
-    fpath = self.CreateTempFile(contents=b'a' * 2 * ONE_KIB)
+    fpath = self.CreateTempFile(contents=get_random_ascii_chars(
+        size=(ONE_MIB*4)))
     boto_config_for_test = ('GSUtil', 'resumable_threshold', str(ONE_KIB))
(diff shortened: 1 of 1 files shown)


gslib/tests/test_cp.py
```diff
@@ -3807,7 +3807,8 @@ def test_cp_resumable_upload_start_over_http_error(self):
 
   def start_over_error_test_helper(self, http_error_num):
     bucket_uri = self.CreateBucket()
-    fpath = self.CreateTempFile(contents=b'a' * 2 * ONE_KIB)
+    rand_chars = get_random_ascii_chars(size=(ONE_MIB * 4))
```
houglum: Could you leave a comment here explaining that the size has to be fairly large so that the callback can halt the download before it finishes?

catleeball: Sure thing!

gslib/tests/test_cp.py
```diff
@@ -3807,7 +3807,10 @@ def test_cp_resumable_upload_start_over_http_error(self):
 
   def start_over_error_test_helper(self, http_error_num):
     bucket_uri = self.CreateBucket()
-    fpath = self.CreateTempFile(contents=b'a' * 2 * ONE_KIB)
+    # The bucket contents need to be fairly large to avoid the race condition
```
houglum: s/bucket contents/object contents/

catleeball: ack! fixed

PR #823, opened 2019-06-24
Adding support for requestedPolicyVersion to storage.bucket.getIamPolicy

thomasmaclean:
b/135208774
go/iam-policy-version-integration

### DIFF
--- a/gslib/gcs_json_api.py
+++ b/gslib/gcs_json_api.py
@@ -333,7 +333,8 @@ class GcsJsonApi(CloudApi):
 
   def GetBucketIamPolicy(self, bucket_name, provider=None, fields=None):
     apitools_request = apitools_messages.StorageBucketsGetIamPolicyRequest(
-        bucket=bucket_name, userProject=self.user_project)
+        bucket=bucket_name, userProject=self.user_project,
+        requestedPolicyVersion=3)
     global_params = apitools_messages.StandardQueryParameters()
     if fields:
       global_params.fields = ','.join(set(fields))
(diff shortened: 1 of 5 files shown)


catleeball:
I added the bug number and a go link to the project doc to the description for context, hope that's okay. :slightly_smiling_face:

gslib/gcs_json_api.py
```diff
@@ -326,7 +326,8 @@ def _ValidateEncryptionFieldsForMetadataGetRequest(self, fields=None):
 
   def GetBucketIamPolicy(self, bucket_name, provider=None, fields=None):
     apitools_request = apitools_messages.StorageBucketsGetIamPolicyRequest(
-        bucket=bucket_name, userProject=self.user_project)
+        bucket=bucket_name, userProject=self.user_project,
+        requestedPolicyVersion=3)
```
catleeball: Should this value be configurable now or in the future? I saw the doc mentioned the value being "at least 3", but I don't have a lot of background on this, so I wasn't sure if a user would ever actually need to set this greater than 3.

thomasmaclean:
@catleeball I think adding configuration here would only pollute the configuration file as this will likely only change once in a blue moon. If this situation ever changes we can obviously revisit this.

gslib/gcs_json_api.py
thomasmaclean: I think adding configuration here would only pollute the configuration file as this will likely only change once in a blue moon. If this situation ever changes we can obviously revisit this.

thomasmaclean: Resolved this as I just noticed there's a note saying this doesn't need to be configurable in the doc you linked to earlier.

thomasmaclean:
Just a quick side note: this is now running fine against all environments, i.e. this is a non-breaking change in environments that don't support the policy version field.

PR #825, opened 2019-06-24
Added retry decorators to more tests

catleeball:
Per b/135780661 and b/29278122, we sometimes see multiprocess testing on
nonexistent buckets hang. Timeouts have been added to the threads
running these tests, and now retry decorators are being added to the
tests themselves so they can attempt the test again after the previous
thread dies.


catleeball:
> LGTM, with the nit that we should probably also leave a comment above each of these saying to remove them once we figure out why b/135780661 happens.

Got it, added comments :slightly_smiling_face:

PR #832, opened 2019-06-27
Updated hex conversions

catleeball:
b/135635815

### DIFF
--- a/gslib/utils/text_util.py
+++ b/gslib/utils/text_util.py
@@ -149,7 +149,8 @@ def DecodeLongAsString(long_to_convert):
   Returns:
     String decoded from the input long.
   """
-  return str(long_to_convert)
+  unhexed = binascii.unhexlify(hex(long_to_convert)[2:].rstrip('L'))
+  return six.ensure_str(unhexed)
 
 
 def EncodeStringAsLong(string_to_convert):
@@ -164,9 +165,12 @@ def EncodeStringAsLong(string_to_convert):
   Returns:
     Long that represents the input string.
   """
-  if six.PY2:
-    return long(codecs.encode(string_to_convert, 'hex_codec'))
-  return int(codecs.encode(string_to_convert, 'hex_codec'))
+  hex_bytestr = codecs.encode(six.ensure_binary(string_to_convert), 'hex_codec')
+  if six.PY3:
+    long = int
+  # Note that `long`/`int` accepts either `bytes` or `unicode` as the
+  # first arg in both py2 and py3:
+  return long(hex_bytestr, 16)
 
 
 def FixWindowsEncodingIfNeeded(input_str):
--- a/gslib/utils/text_util.py
+++ b/gslib/utils/text_util.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
+import codecs
 import os
 import sys
 import io
@@ -148,10 +149,7 @@ def DecodeLongAsString(long_to_convert):
   Returns:
     String decoded from the input long.
   """
-  if isinstance(long_to_convert, basestring):
-    # Already converted.
-    return long_to_convert
-  return hex(long_to_convert)[2:-1].decode('hex')
+  return str(long_to_convert)
 
 
 def EncodeStringAsLong(string_to_convert):
@@ -166,7 +164,9 @@ def EncodeStringAsLong(string_to_convert):
   Returns:
     Long that represents the input string.
   """
-  return long(string_to_convert.encode('hex'), 16)
+  if six.PY2:
+    return long(codecs.encode(string_to_convert, 'hex_codec'))
+  return int(codecs.encode(string_to_convert, 'hex_codec'))
 
 
 def FixWindowsEncodingIfNeeded(input_str):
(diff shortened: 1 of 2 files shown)


gslib/utils/text_util.py
```diff
@@ -148,10 +149,7 @@ def DecodeLongAsString(long_to_convert):
   Returns:
     String decoded from the input long.
   """
-  if isinstance(long_to_convert, basestring):
-    # Already converted.
-    return long_to_convert
-  return hex(long_to_convert)[2:-1].decode('hex')
+  return str(long_to_convert)
```
houglum: This doesn't give us the same result we had before.  Let's take a real x-amz-version-id value and test it:

'9PpsRjBGjBh90IvIS96dgRc_UL6NyGqD'
(encoded long: 25923956239092482442895228561437790190304192615858167521375267910356975448388L)

(Note that all the `long`s we're dealing with were originally strings from header values, and thus should be convertible back to utf-8 -- probably ascii if AWS was strict about those things, but I'm not sure how much I trust the strict ascii-ness of their values)

Py2:
```
>>> import binascii
>>> var = 25923956239092482442895228561437790190304192615858167521375267910356975448388L
>>> hex(var)
'0x39507073526a42476a42683930497649533936646752635f554c364e79477144L'

# Just strips off the "0x" and trailing "L". Note that the L isn't there in Python 3 though...
>>> hex(var)[2:-1]
'39507073526a42476a42683930497649533936646752635f554c364e79477144'

>>> hex(var)[2:-1].decode('hex')
'9PpsRjBGjBh90IvIS96dgRc_UL6NyGqD'

# Py2 just uses this under the hood for decode('hex'):
>>> binascii.unhexlify(hex(var)[2:-1])
'9PpsRjBGjBh90IvIS96dgRc_UL6NyGqD'
```

But this function gives us something different:
```
>>> str(var)
'25923956239092482442895228561437790190304192615858167521375267910356975448388'
```

And in Py3:
```
>>> import binascii
# Note there's no trailing L here:
>>> var = 25923956239092482442895228561437790190304192615858167521375267910356975448388

# No trailing L on this string:
>>> hex(var)
'0x39507073526a42476a42683930497649533936646752635f554c364e79477144'

# Close, but we get bytes...
>>> binascii.unhexlify(hex(var)[2:])
b'9PpsRjBGjBh90IvIS96dgRc_UL6NyGqD'

# There we go, a string representing the encoded long:
>>> import six
>>> six.ensure_str(binascii.unhexlify(hex(var)[2:]))
'9PpsRjBGjBh90IvIS96dgRc_UL6NyGqD'
```

So, I think a Py2-and-3 compatible solution might look like this:

```
unhexed = binascii.unhexlify(hex(long_to_convert)[2:].rstrip('L'))
return six.ensure_str(unhexed)
```

gslib/utils/text_util.py
```diff
@@ -166,7 +164,9 @@ def EncodeStringAsLong(string_to_convert):
   Returns:
     Long that represents the input string.
   """
-  return long(string_to_convert.encode('hex'), 16)
+  if six.PY2:
+    return long(codecs.encode(string_to_convert, 'hex_codec'))
+  return int(codecs.encode(string_to_convert, 'hex_codec'))
```
houglum: Similar comment to the one above.  I think this might work:

```

hex_bytestr = codecs.encode(six.ensure_binary(string_to_convert), 'hex_codec')
if six.PY3:
  long = int
# Note that `long` accepts either `bytes` or `unicode` as the
# first arg in both py2 and py3:
return long(hex_bytestr, 16)
```

gslib/utils/text_util.py
catleeball: Woah, thank you for the info and a really nice solution!! I'll add that. 🙂 (also I think I forgot to finish editing this function and just left it in the debug print lol)

gslib/utils/text_util.py
catleeball: Whoops dang, you're right, my logic is a bit doofed! Thanks for being smart and catching that 🙂

gslib/utils/text_util.py
```diff
@@ -148,10 +149,8 @@ def DecodeLongAsString(long_to_convert):
   Returns:
     String decoded from the input long.
   """
-  if isinstance(long_to_convert, basestring):
-    # Already converted.
-    return long_to_convert
-  return hex(long_to_convert)[2:-1].decode('hex')
+  unhexed = binascii.unhexlify(hex(long_to_convert)[2:].rstrip('L'))
```
houglum: Don't forget to import binascii :)

catleeball: Whoops! Added that in. 👍 

What's a good way to test these code paths btw? Integ tests with XML api preferred and AWS credentials still seemed to miss my bugs, and I was able to successfully cp and cat s3 buckets when I cut the PR. Maybe my boto is misconfigured though if it's missing these?

houglum:
Sent https://github.com/catleeball/gsutil/pull/6. Please merge that, then this should be good to go.  I ran that unit test in both py2 and py3 (needed to temporarily comment out the gsutil.py version restriction to run it in py3).

gslib/utils/text_util.py
houglum: You'll need to run `test` with the `-s` argument to test against S3, e.g. `BOTO_CONFIG=/path/to/boto.cfg ./gsutil test -s -p 30`

catleeball: Oh, I didn't know about the `-s` flag for test! Thanks! 😄

catleeball:
Thanks for PRing a test in!! Just left a comment on there @houglum 🙂

PR #836, opened 2019-06-28
Add mode attribute to {up,down}load wrapper classes

houglum:
This was causing a test to fail under Python 3. The underlying stream
was being treated as opened in text mode, rather than in binary mode,
because write_to_fd() could not determine its mode (and assumed text).
This caused sliced downloads to fail in
cp.TestCp.test_cp_resumable_download_gzip. Internal devs can see the
full failure stack trace and more details in this Issue Tracker issue:
https://issuetracker.google.com/issues/122963518#comment5

### DIFF
--- a/gslib/gcs_json_api.py
+++ b/gslib/gcs_json_api.py
@@ -1190,12 +1190,6 @@ class GcsJsonApi(CloudApi):
                                      end_byte=end_byte,
                                      serialization_data=serialization_data,
                                      decryption_tuple=decryption_tuple)
-      # If you are fighting a redacted exception spew in multiprocess/multithread
-      # calls, add your exception to HTTP_TRANSFER_EXCEPTIONS and put
-      # something like this immediately after the following except statement:
-      # import sys, traceback; sys.stderr.write('\n{}\n'.format(
-      #     traceback.format_exc())); sys.stderr.flush()
-      # This may hang, but you should get a stack trace spew after Ctrl-C.
       except HTTP_TRANSFER_EXCEPTIONS as e:
         self._ValidateHttpAccessTokenRefreshError(e)
         start_byte = download_stream.tell()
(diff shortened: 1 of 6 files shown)


gslib/gcs_json_api.py
```diff
@@ -1206,6 +1200,24 @@ def _PerformResumableDownload(self,
           retries = 0
         retries += 1
         if retries > self.num_retries:
+          ##### DEBUG
```
catleeball: Thanks for including some debug code in here, especially with multiprocessing it can get pretty hairy trying to wrestle useful info out of whichever thread has gone wild

gslib/utils/copy_helper.py
```diff
@@ -2534,6 +2534,11 @@ def __init__(self, fp, tracker_file_name, src_obj_metadata, start_byte,
     self._start_byte = start_byte
     self._end_byte = end_byte
 
+  @property
+  def mode(self):
+    """Returns the mode of the underlying file descriptor, or None."""
+    return getattr(self._orig_fp, 'mode', None)
```
catleeball: In what circumstance would the original file descriptor's mode be none? Only if we construct a `SlicedDownloadFileWrapper` object without a valid `fp`?

Asking mostly out of curiosity if this can happen in the wild, and if we should test for that circumstance, and if we're relying on the calling function to handle the error.

houglum: I don't know of such a scenario, but `write_to_fd()` has an else clause (which was being hit in the scenario I'm fixing) that's hit when the `fd` doesn't have a `mode` attribute. Maybe this was added solely because of this `SlicedDownloadFileWrapper`, or maybe there are other wrapper classes like this too?

Come to think of it, I should probably add this property to all of our Upload/Download Wrapper classes.  To answer your question, I don't think this should ever occur in practice, but I don't know that for sure.  We should probably add a statement in `write_to_fd()` to print a debug statement whenever it writes to a file descriptor that doesn't have a `mode` attribute (although this could get pretty spammy real quick if you're writing lots of blocks to a large file...)

houglum: Added the `mode` property to other wrapper classes as well, just pushed a commit. PTAL.

houglum: (I think I got them all; the command I used to find these classes was `grep "class [A-Z][A-Za-z_0-9]\+Wrapper" . -r`).

catleeball: Understood! New commit LGTM! :boat:

PR #843, opened 2019-07-03
Fixes nits for .par bundles.

houglum:
Where appropriate, we now transform bytes from pkgutil.get_data into unicode
strings in Python 3 (this fixes issues where checksum and version strings
weren't the expected types - in the .par build, these were bytes (as the
documentation seems to indicate `get_data()` will return), but when run
locally under Python 3, they were unicode; these casts only happen when
the data should be converted to unicode and isn't already unicode).

Also, we now skip the metrics tests when running from a par file, since
the interpreter is nested within the archive (but it's needed to spawn
the metrics subprocess).

### DIFF
--- a/gslib/__init__.py
+++ b/gslib/__init__.py
@@ -126,7 +126,7 @@ def _GetFileContents(filename):
 
   Returns:
     A tuple containing the absolute path to the requested file and the file's
-    contents. If the file is not actually on disk, the file path will be None.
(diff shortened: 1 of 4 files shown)


houglum:
Merging these since they were approved in the internal changelist already (see cl/256419965 and cl/256282588), and the Kokoro tests are all green.

PR #847, opened 2019-07-10
Fixes bug where copying files > 100MiB from GCS to S3 hangs

mkalmanson:
https://issuetracker.google.com/issues/135959331

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -44,6 +44,7 @@ from boto.storage_uri import BucketStorageUri
 from gslib.cloud_api import ResumableUploadStartOverException
 from gslib.commands.config import DEFAULT_SLICED_OBJECT_DOWNLOAD_THRESHOLD
 from gslib.cs_api_map import ApiSelector
+from gslib.daisy_chain_wrapper import _DEFAULT_DOWNLOAD_CHUNK_SIZE
 from gslib.discard_messages_queue import DiscardMessagesQueue
 from gslib.gcs_json_api import GcsJsonApi
 from gslib.parallel_tracker_file import ObjectFromTracker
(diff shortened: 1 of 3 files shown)


gslib/tests/testcase/integration_testcase.py
```diff
@@ -958,17 +960,33 @@ def RunGsUtil(self,
...
     if six.PY3:
       # TODO(b/135936279): Make this number configurable in .boto
       comm_kwargs['timeout'] = 180
+    else:
+      timer = Timer(180, kill)
+      timer.start()
+
     c_out = p.communicate(**comm_kwargs)
+
+    if timer is not None:
```
thomasmaclean: A little rusty on python here, but will Python 3 throw an error because timer is undefined? Otherwise everything looks great, this looks like a pretty good approach!

mkalmanson: It did.  And I fixed it 5875 :-)

thomasmaclean: :+1:

thomasmaclean:
You're getting a failure here:
https://sponge.corp.google.com/target?id=671e6af0-c5fd-4755-8114-17e59323091b&target=cloud_storage_gsutil%2Fmacos%2Fpy27_xml&searchFor=

gslib/tests/test_cp.py
```diff
@@ -1382,6 +1371,20 @@ def test_cross_provider_large_cp(self):
       # Ensure copy also works across json upload chunk boundaries.
       self.RunGsUtil(['cp', suri(s3_key), suri(gs_bucket)])
 
+  @unittest.skipUnless(HAS_S3_CREDS, 'Test requires both S3 and GS credentials')
+  def test_gs_to_s3_multipart_cp(self):
+    """Ensure daisy_chain works for an object that is downloaded in 2 parts."""
+    s3_bucket = self.CreateBucket(provider='s3')
+    gs_bucket = self.CreateBucket(provider='gs', prefer_json_api=True)
+    gs_key = self.CreateObject(bucket_uri=gs_bucket,
+                               contents=b'b' * 101 * 1024 * 1024,
```
houglum: Can we maybe define this in relation to `_DEFAULT_DOWNLOAD_CHUNK_SIZE` in `gslib/daisy_chain_wrapper.py`?  E.g. `num_bytes = int(_DEFAULT_DOWNLOAD_CHUNK_SIZE * 1.1)`.  Without doing so, if we ever change that constant, it would be easy to forget to change this test too.

houglum:
Looks like this fixes the issue in https://github.com/GoogleCloudPlatform/gsutil/pull/638.  Feel free to close that once this is submitted :)

gslib/tests/test_cp.py
mkalmanson: Done

PR #858, opened 2019-07-19
Add .pylintrc to check a subset of warnings.

houglum:
Also adds a shell script that selects which files should be passed to
the linter.

Fixes https://issuetracker.google.com/issues/13656871.

TODOs:
- We still need to gradually go through the rest of the warnings that are
commented out and see which ones can safely be uncommented, which ones
are revealing actual issues, and which ones are just noise (and thus
should stay disabled) due to some of the odd imports and monkeypatching we do.
- We should also run this script as part of the TravisCI presubmits...
probably under both Python 2 and 3? I haven't tried running it under Py3
yet.
- All of the TODOs listed in the pylintrc file.

### DIFF
--- a/test/.pylintrc_limited
+++ b/test/.pylintrc_limited
@@ -1,0 +1,473 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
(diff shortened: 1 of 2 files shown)


thomasmaclean:
created a bug to track the docs/travis updates: 138375199

PR #859, opened 2019-07-19
Only try calling os.setsid if that function exists

houglum:
It does not exist on Windows, so the integration tests were failing.
Not sure why Kokoro was reporting successes; maybe it was only looking
up the status of the first tests to finish (which are usually the Linux
tests)?  Regardless, this fixes the issue of trying to call os.setsid on
Windows.


houglum:
Going to go ahead and submit this without review so that we can un-block other PRs.

PR #866, opened 2019-07-29
Make TravisCI run pylint.

houglum:
Fixes https://issuetracker.google.com/issues/138375199.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ before_install:
 # command to install dependencies
 install:
   - pip install .
+  - pip install pylint
   - pip install yapf
 
 # command to run tests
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -30,11 +30,13 @@ before_install:
 # command to install dependencies
 install:
   - pip install .
+  - pip install pylint
   - pip install yapf
```
thomasmaclean: Should we remove yapf from the install/script portion, or is this not replacing yapf?

houglum: Nope; I left a comment in the bug as well, but the TL;DR is that these tools do mostly different things (both of which are important).

PR #869, opened 2019-08-01
Make SkipForXML skip if running tests against S3.

houglum:
This was causing some perfdiag tests to fail, since we were trying to gzip encode a portion of the payload in an API request that didn't support it.


houglum:
Going to go ahead and merge this, as I've run the prerelease tests using a release candidate built at this commit, and they all passed (see https://issuetracker.google.com/issues/137876039).

PR #871, opened 2019-08-02
Update helptexts before release.

houglum:
This fixes the formatting so that options tables are displayed correctly
(without leading and trailing rows) and indents commands so that they're
rendered as `<pre>` divs in the generated docs.


houglum:
Merging these changes, as the generated docs were approved in an internal code review (cr/261282114).

Issue #877, opened 2019-08-22
gsutil --help crashed in python3 when trying to use a pager


feileacan:
Repro Steps: `python3 gsutil.py --help` on a sufficiently small terminal (`tput lines` == 53)
Error:
```
Traceback (most recent call last):
  File "gsutil.py", line 128, in <module>
    RunMain()
  File "gsutil.py", line 124, in RunMain
    sys.exit(gslib.__main__.main())
  File "/home/davidmunro/proj/gsutil/gslib/__main__.py", line 315, in main
    command_runner.RunNamedCommand('help')
  File "/home/davidmunro/proj/gsutil/gslib/command_runner.py", line 411, in RunNamedCommand
    return_code = command_inst.RunCommand()
  File "/home/davidmunro/proj/gsutil/gslib/commands/help.py", line 173, in RunCommand
    self._OutputHelp(''.join(output))
  File "/home/davidmunro/proj/gsutil/gslib/commands/help.py", line 201, in _OutputHelp
    Popen(pager, stdin=PIPE).communicate(input=help_str)
  File "/usr/lib/python3.6/subprocess.py", line 828, in communicate
    self._stdin_write(input)
  File "/usr/lib/python3.6/subprocess.py", line 781, in _stdin_write
    self.stdin.write(input)
TypeError: a bytes-like object is required, not 'str'
```

The same command using python2 works fine.

PR #878, opened 2019-08-22
Fix crash when showing help in a pager on Python 3. (#877)

feileacan:
Testing done: gsutil test -u


feileacan:
Forgot to mention, also tested by running `python --help` and `python3 --help`. Both showed help with this changed, without this change `python --help` worked but `python3 --help` crashes.

PR #893, opened 2019-09-18
Add the uniform bucket-level access command

HyphnKnight:
Adds the ubla command to the gsutil api, currently this still sets the bucketPolicyOnly property on a bucket but as the two fields are interchangeable this is functionally equivalent. Once the api has been released allowing the use of uniform bucket-level access command a pull request will be made updating gsutil to use the new property.

### DIFF
--- a/gslib/tests/test_ubla.py
+++ b/gslib/tests/test_ubla.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


gslib/tests/test_ubla.py
```diff
@@ -0,0 +1,102 @@
...
+    if self.test_api == ApiSelector.XML:
+      return unittest.skip(
+          'XML API has no concept of Uniform bucket-level access')
+    bucket_uri = self.CreateBucket()
+    self._AssertEnabled(bucket_uri, False)
+
+  def test_turning_off_on_enabled_buckets(self):
+    if self.test_api == ApiSelector.XML:
+      return unittest.skip(
+          'XML API has no concept of Uniform bucket-level access')
```
thomasmaclean: You don't have to fix this since I know this was a port of the original BPO test logic, but there are decorators to avoid tests that should only use certain APIs. See https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/tests/test_kms.py#L38

PR #899, opened 2019-09-19
add support for deleted users syntax

habahut:
add's support for zombie principals of the form "deleted::<>".

Python 2 and 3 tests pass as per: https://g3doc.corp.google.com/cloud/storage/g3doc/internal/gsutil/gsutil-dev.md?cl=head#running-tests

### DIFF
--- a/gslib/tests/test_iam.py
+++ b/gslib/tests/test_iam.py
@@ -310,6 +310,8 @@ class TestIamHelpers(testcase.GsUtilUnitTestCase):
       bstt(True, 'user:foo@bar.com')
     with self.assertRaises(CommandException):
       bstt(True, 'user:foo@bar.com:')
+    with self.assertRaises(CommandException):
+      bstt(True, 'deleted:user:foo@bar.com?uid=1234:')
 
   def test_remove_all_roles(self):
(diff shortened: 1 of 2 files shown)


habahut:
It doesn't look like these Kokoro failures were due to my changes, is it possible to re-run these?

thomasmaclean:
There isn't a way to re-run them through the interface, although I've occasionally just force-pushed a commit message tweak to get them to run. That said, we've seen a little bit of flakiness recently, so I won't hold it against you :)

habahut:
makes sense, I force pushed a change to the commit message and the tests are passing now

thomasmaclean:
Changes all look good... Just fix the yapf errors:
`python3 -m yapf -irp .  `

habahut:
I learned about a new complication, we can't just split on the '?" character since there can actually be email addresses in the system which have a "?" in them. So updated the logic and added new tests to handle this case.

habahut:
actually, I spoke to Josh and it turns out we don't want to strip the deleted: prefix or ?uid suffix. I'm going to just close this PR for the moment, and reopen once I have made these changes to prevent confusion.

PR #916, opened 2019-10-10
Change object versioning terminology

starsandskies:
Instead of "archived" versions, refer to them as "noncurrent" versions. This will avoid terminology overload and matches equivalent changes being made to the documentation found in cloud.google.com/storage/docs

b/140083251 tracks the request to make this change.


thomasmaclean:
LGTM

PR #923, opened 2019-10-23
Add userproject support to signurl

thoutenbos:
Add support to signurl to provide an optional projectname that will be billed, useful for generating presigned links for buckets that are set to requester pays.

### DIFF
--- a/gslib/commands/signurl.py
+++ b/gslib/commands/signurl.py
@@ -79,7 +79,7 @@ _MAX_EXPIRATION_TIME = timedelta(days=7)
 
 _SYNOPSIS = """
   gsutil signurl [-c <content_type>] [-d <duration>] [-m <http_method>] \\
-      [-p <password>] [-r <region>] keystore-file url...
+      [-p <password>] [-r <region>] [-u userproject] keystore-file url...
 """
 
 _DETAILED_HELP_TEXT = ("""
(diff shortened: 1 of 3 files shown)


thoutenbos:
@thomasmaclean what would be needed to get this merged?

thomasmaclean:
Apologies I was waiting until some work around the `signurl` was complete before merging this PR in, and then forgot to check back on this one. In any case, the `-u` option is now taken up by a new option. Can this be switched to `-b` with the idea that you're "billing" the given project?

thoutenbos:
The `-u` option is used for this purpose everywhere in `gsutil`, it would be very unexpected to use `-b` for the same feature only in `signurl`

dilipped:
The `-u` option is being used as a top level options in gsutil everywhere. https://cloud.google.com/storage/docs/gsutil/addlhelp/TopLevelCommandLineOptions
Is there a reason why we can't keep it as a top level option?

thomasmaclean:
Sounds good. I think it completely makes sense to use the global `-u` option and just document the billing behavior as it relates to requester pays.

thomasmaclean:
@thoutenbos Out of curiosity, what was the behavior of the global `-u` option when you tested it with `signurl` on a requester pays bucket?

thoutenbos:
It has been a while to be honest, but I'm pretty sure it didn't do anything, otherwise I would not have gone through the trouble of implementing it myself

pauldwolfe:
Hey all, I work for the client originally requesting this feature from @thoutenbos. he'll no longer be working on this but we could really use this feature (actually feels a bit more like a bug). 

How can we get it prioritized somewhere?

thomasmaclean:
Closing in favor of #1068

PR #928, opened 2019-11-04
Clarify bucketpolicyonly name change to UBLA

karenarialin:
--Change name from BPO to uniform bucket-level access
--Clarify that BPO is supported but UBLA is preferred

Parallel to changes made in cl/279111918


karenarialin:
@googlebot I signed it!

PR #938, opened 2019-12-10
Removing -d from the default usage example

laylasells:
Separate the usage of syncing from deleting objects to limit accidental deletion. Internal: 72370795

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -119,11 +119,11 @@ _DETAILED_HELP_TEXT = ("""
   subdirectory. For example, to make gs://mybucket/data match the contents of
   the local directory "data" you could do:
 
-    gsutil rsync -d data gs://mybucket/data
+    gsutil rsync data gs://mybucket/data
 
   To recurse into directories use the -r option:
 
-    gsutil rsync -d -r data gs://mybucket/data
+    gsutil rsync -r data gs://mybucket/data
 
   To copy only new/changed files without deleting extra files from
   gs://mybucket/data leave off the -d option:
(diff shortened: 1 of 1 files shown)


starsandskies:
I think there will need to be a bit more to this edit than just removing instances of -d:

1) It seems like -d is still an important flag which should have positive examples provided (not just examples warning about its dangers)

2) The descriptions associated with examples that have -d removed don't appear to accurately capture what the examples do. For instance, the description for the example on line 121 says the example will "match" the bucket to the source; however, without the -d flag, that won't be true: the bucket *will* contain the same files that are in the source, but it could also have other files, and so it won't strictly match. I believe other example descriptions are similarly affected.

laylasells:
1) agreed -d is important. the second half covers it as well. Would it be helpful to change the heading “be careful when using -d option” to first say “to delete extra files use -d” and then continue with the warning.

2) I did see one section where it discussed the deletion in the description and pulled that part out as well. I see where the description section also states it (my read was just saying if you used it it would delete it. Which I thought was good to discuss even if not shown.)



Sent from my phone

On Dec 10, 2019, at 2:02 PM, starsandskies <notifications@github.com> wrote:

﻿

I think there will need to be a bit more to this edit than just removing instances of -d:

  1.  It seems like -d is still an important flag which should have positive examples provided (not just examples warning about its dangers)

  2.  The descriptions associated with examples that have -d removed don't appear to accurately capture what the examples do. For instance, the description for the example on line 121 says the example will "match" the bucket to the source; however, without the -d flag, that won't be true: the bucket will contain the same files that are in the source, but it could also have other files, and so it won't strictly match. I believe other example descriptions are similarly affected.

PR #940, opened 2019-12-13
Updating rsync to separate copy and mirror uses

laylasells:
Removing the -d delete from the top section and moving it to it's own mirroring section.  The goal is to reduce confusion when people want to sync from source without deleting files and avoid accidental deletion. Internal CL approved: 284841364


laylasells:
cc @mkalmanson and @thomasmaclean
this makes the discussed changes from PR #938. PTAL

laylasells:
The Travis error doesn't seem related to this change.

PR #945, opened 2020-01-02
Update wildcard_iterator.py

donatas-luciunas:
With these changes I was able to upload utf-8 filenames to a bucket using Windows.

> os.walk() will not attempt to decode
I guess this assumption is wrong. os.walk() works well once you provide Unicode string as an argument.

This is not a solution / completed pull request, just wanted to share results of my research


donatas-luciunas:
@googlebot I signed it!

dilipped:
Thanks for the PR. Closing this since the issue was fixed by another https://github.com/GoogleCloudPlatform/gsutil/pull/926

PR #948, opened 2020-01-09
Documentation updates while reviewing 4.47 release

thomasmaclean:


gslib/commands/rsync.py
```diff
@@ -151,15 +151,15 @@
   workstation.
 
 
-<B>Using -d Option (with caution!) to mirror source and destination.</B>
+<B>Using ``-d`` Option (with caution!) to mirror source and destination.</B>
```
starsandskies: How sure are you that this will prevent the capitalization issue? If "very" then great! If not, I've fixed the issue on the post-processing side, so you can revert this if you want.

PR #967, opened 2020-02-21
Add examples for cmd.exe and PowerShell

grahamdunn-work:
Add examples for windows cmd.exe and powershell -x regex patterns with platform-specific escape characters

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -513,8 +513,14 @@ _DETAILED_HELP_TEXT = ("""
 
                  will skip all .txt and .jpg files in dir.
 
-                 NOTE: When using this on the Windows command line, use ^ as an
+                 NOTE: When using this in the Windows cmd.exe command line interpreter, use ^ as an
                  escape character instead of \ and escape the | character.
+                 
+                   gsutil rsync -x ".*\.txt$^|.*\.jpg$"
+                   
+                 NOTE: Windows Powershell requires single-quoted regex
+                 
+                   gsutil rsync -x '.*\.txt$^|.*\.jpg$'
 """)
 # pylint: enable=anomalous-backslash-in-string
 


grahamdunn-work:
@googlebot I signed it!

gslib/commands/rsync.py
```diff
@@ -513,8 +513,14 @@
 
                  will skip all .txt and .jpg files in dir.
 
-                 NOTE: When using this on the Windows command line, use ^ as an
+                 NOTE: When using this in the Windows cmd.exe command line interpreter, use ^ as an
                  escape character instead of \ and escape the | character.
+                 
+                   gsutil rsync -x ".*\.txt$^|.*\.jpg$"
```
starsandskies, line 519: Sorry that I missed reviewing this way back when :-/

My regex is pretty weak, so I could be way off, but should this be ".*^.txt$^|.*^.jpg$" ?

gslib/commands/rsync.py
```diff
@@ -513,8 +513,14 @@
...
+                   
+                 NOTE: Windows Powershell requires single-quoted regex
+                 
+                   gsutil rsync -x '.*\.txt$^|.*\.jpg$'
```
starsandskies, line 523: Does Powershell use the same ^ escape character as cmd.exe? If so, my above comment may apply here as well, and it may be worth saying something like: "Windows Powershell requires single-quoted regex and, similar to using the cmd.exe command line interpreter, uses ^ as an escape character"

gslib/commands/rsync.py
NickGoog: We probably don't need to list the example again, so this is fine if we remove lines 518 and 519.

NickGoog:
Closing b/c of no response to @starsandskies's question from 5/26/2020 and merge conflict.
PR author feel free to reopen with answer to question and merge conflict resolved.

PR #974, opened 2020-03-13
 Use print instead of text_util.print_to_fd in kms set encryption

dannyqiu:
This addresses the ordering of the printed messages when the PatchBucket api call fails, since `text_util.print_to_fd` buffers in `stdout` until after the exception is raised to the user, which is printed to `stderr`.

This change depends on https://github.com/GoogleCloudPlatform/gsutil/pull/979

### DIFF
--- a/gslib/commands/kms.py
+++ b/gslib/commands/kms.py
@@ -316,8 +316,8 @@ class KmsCommand(Command):
 
     bucket_metadata.encryption = apitools_messages.Bucket.EncryptionValue(
         defaultKmsKeyName=self.kms_key)
-    text_util.print_to_fd('Setting default KMS key for bucket %s...' %
-                          str(bucket_url).rstrip('/'))
+    print('Setting default KMS key for bucket %s...' %
+          str(bucket_url).rstrip('/'))
     self.gsutil_api.PatchBucket(bucket_url.bucket_name,
                                 bucket_metadata,
                                 fields=['encryption'],
--- a/gslib/tests/test_kms.py
+++ b/gslib/tests/test_kms.py
@@ -157,7 +157,7 @@ class TestKmsSubcommandsFailWhenXmlForced(testcase.GsUtilIntegrationTestCase):
       ('Credentials', 'gs_access_key_id', 'dummykey'),
       ('Credentials', 'gs_secret_access_key', 'dummysecret'),
   ]
-  dummy_keyname = ('projects/my-project/locations/global/'
+  dummy_keyname = ('projects/my-project/locations/us-central1/'
                    'keyRings/my-keyring/cryptoKeys/my-key')
 
   def DoTestSubcommandFailsWhenXmlForcedFromHmacInBotoConfig(self, subcommand):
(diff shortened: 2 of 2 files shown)


gslib/commands/kms.py
```diff
@@ -316,8 +316,8 @@ def _EncryptionSetKey(self, bucket_metadata, bucket_url,
 
     bucket_metadata.encryption = apitools_messages.Bucket.EncryptionValue(
         defaultKmsKeyName=self.kms_key)
-    text_util.print_to_fd('Setting default KMS key for bucket %s...' %
-                          str(bucket_url).rstrip('/'))
+    print('Setting default KMS key for bucket %s...' %
```
dilipped, line 319: Based on the PR description I understand that the main motivation for this change is that `print_to_fd` buffers the text and we want to avoid that. Is it guaranteed that `print` will not do the same? Wouldn't `print` buffer the text as well?
If we are sure that print does work as expected then perhaps we should add a test case here that can re-create the fail case. i.e something that fails for `print_to_fd` and works fine for `print`.

Another thing we might want to consider is that `print_to_fd` is being used at a lot of places in the code. So if this is a problem for other commands as well then perhaps we should fix the `print_to_fd` function itself by calling `flush` etc.

dannyqiu: I'm not quite sure what is going on under the hood with `print`. It seems that `print` may flush with a newline because it of the line-buffering rules of stdout, whereas `print_to_fd` writes to the underlying buffer in https://github.com/GoogleCloudPlatform/gsutil/blob/b60ce18a656d5f86e10b7677250958d5392f70da/gslib/utils/text_util.py#L380. Even with a newline in the call to `write`, it seems to be buffered by the underlying system.

A simple test demonstrates this behavior:
```python
$ cat test.py 
import sys
sys.stdout.buffer.write(b'buffer write to stdout\n')
sys.stderr.buffer.write(b'buffer write to stderr\n')
$ python3 test.py 
buffer write to stderr
buffer write to stdout
```

It's difficult to test this behavior because of how it shows up to the user. The information message is printed to `stdout`, while the error message is printed to `stderr`. We would have to test the "racing" behavior of these two streams, which is an underlying system implementation.

I wanted to address this single case before making a change to `print_to_fd` to call `flush`, since I'm not too familiar with all of the other use cases of `print_to_fd` (as in does it rely on the buffering?)

I also noticed that the other information messages in this file are already using `print` instead of `print_to_fd`: 
- https://github.com/GoogleCloudPlatform/gsutil/blob/b60ce18a656d5f86e10b7677250958d5392f70da/gslib/commands/kms.py#L271
- https://github.com/GoogleCloudPlatform/gsutil/blob/b60ce18a656d5f86e10b7677250958d5392f70da/gslib/commands/kms.py#L286

dilipped: Thanks for the explanation.  I doubt if there are any use cases that are dependent on buffering in `print_to_fd`, but even I am not fully aware of the code base so I can't say for sure. For now, I think your change should be safe enough.

PR #981, opened 2020-04-02
Use V4 signature as default for S3

dilipped:
Amazon S3 recommends using V4 signature for all the requests. The current implementation of our fork of [boto](https://github.com/gsutil-mirrors/boto) lib uses V2 signature by default for regions which support both V2 and V4 signature.

We have made changes to our boto fork to use V4 signature by default for all regions https://github.com/gsutil-mirrors/boto/pull/11

Using V4 signature makes some of the Integration tests in gsutil to break for S3, because boto lib does not play well with objects starting with slashes, among other things. Hence changes have been made in test cases to skip certain tests for S3 specifically.  The boto lib itself does not support objects starting with slashes, so it is safe to skip these tests for S3.

### DIFF
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1678,6 +1678,42 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
 
     _CopyAndCheck()
 
+  @unittest.skipUnless(HAS_S3_CREDS, 'Test requires both S3 and GS credentials')
+  def test_copy_object_to_dir_s3_v4(self):
+    """Tests copying object from s3 to local dir with v4 signature.
+
+    Regions like us-east2 accept only V4 signature, hence we will create
+    the bucket in us-east2 region to enforce testing with V4 signature.
+    """
+    src_bucket_uri = self.CreateBucket(provider='s3')
+    dst_dir = self.CreateTempDir()
+    self.CreateObject(bucket_uri=src_bucket_uri,
+                      object_name='obj0',
+                      contents=b'abc')
+    self.CreateObject(bucket_uri=src_bucket_uri,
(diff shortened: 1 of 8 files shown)


gslib/tests/test_parallel_cp.py
```diff
@@ -124,6 +125,8 @@ def testCopyingDirContainingOneFileToBucket(self):
     lines = self.AssertNObjectsInBucket(dst_bucket_uri, 1)
     self.assertEqual(suri(dst_bucket_uri, 'dir1', 'foo'), lines[0])
 
+  @SkipForS3('Boto lib required for S3 does not handle paths '
+             'starting with slash.')
```
thomasmaclean, line 129: For posterity you should note the same thing you do in test_rm.py, that V4 slashes preclude paths that start with slash.

dilipped: Fixed

gslib/tests/test_cp.py
```diff
@@ -1678,7 +1678,45 @@ def _CopyAndCheck():
...
 
-  def test_recursive_download_with_leftover_dir_placeholder(self):
+  @unittest.skipUnless(HAS_S3_CREDS, 'Test requires both S3 and GS credentials')
+  def test_copy_object_to_dir_s3_v4(self):
+    """Tests copying object from s3 to local dir with v4 signature.
+
+    Regions like us-east2 accept only V4 signature, hence we will create
+    the bucket in us-east2 region to enforce testing with V4 signature.
+    """
+    src_bucket_uri = self.CreateBucket(provider='s3')
```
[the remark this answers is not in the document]
dilipped: Good catch! Fixed it.

PR #982, opened 2020-04-07
Bring signurl help text in sync with html documentation

dilipped:
The [signurl html documentation](https://cloud.google.com/storage/docs/gsutil/commands/signurl) is generated off of the help text present in the code. Recently the help text for updated directly. This fix brings the help text in sync with the html documentation.


thomasmaclean:
This looks good but I'd like Nate to take a look since he's maintaining the docs and the HTML transform.

starsandskies:
LGTM

No clue if it will generate correctly into the cloud.google.com docs, but if not, we can tinker with it!

PR #986, opened 2020-04-14
Enhancing gsutil User Agent string to match patterns in gcloud

thomasmaclean:
Gsutil's user agent string is missing details implemented in Gcloud. This PR adds those fields and refactors the logic a bit so that there's less duplication across the two API implementations.


gslib/third_party/storage_apitools/storage_v1_client.py
```diff
@@ -35,16 +35,8 @@ class StorageV1(base_api.BaseApiClient):
...
-  if system_util.InvokedViaCloudSdk():
-    _USER_AGENT += ' google-cloud-sdk'
-    if system_util.CloudSdkVersion():
-      _USER_AGENT += '/%s' % system_util.CloudSdkVersion()
-  if MetricsCollector.IsDisabled():
-    _USER_AGENT += ' analytics/disabled'
-  else:
-    _USER_AGENT += ' analytics/enabled'
+  _USER_AGENT = 'apitools Python/%s' % platform.python_version()
+  _USER_AGENT += GetUserAgent(MetricsCollector.IsDisabled())
```
dilipped, line 39: Since the UserAgent is going to remain the same throughout the execution of the program, does it make sense to create a generic global constant which stores the result of GetUserAgent when it is first called in `gslib.__main__.py`. The global constant then can be simply appended here as well as appended to the boto.UserAgent above. This way we can avoid parsing the args twice, although the operation is trivial.

gslib/utils/user_agent_helper.py
```diff
@@ -0,0 +1,49 @@
...
+def GetUserAgent(metrics_off=True):
+  """Using the command arguments return a suffix for the UserAgent string.
+
+  Args:
+    metrics_off: boolean, whether the MetricsCollector is disabled.
+
+  Returns:
+    str, A string value that can be appended to an existing UserAgent.
+  """
+  _, opts = GetArgumentsAndOptions()
```
dilipped, line 32: If we create a module level constant to hold the UserAgent info then perhaps this call will not be necessary. You would be able to pass the args. e.g on line 57 of `gslib.__main__.py`
```
opts, args = GetArgumentsAndOptions()
gslib.USER_AGENT = GetUserAgent(metrics.MetricsCollector.IsDisabled())
boto.UserAgent += gslib.USER_AGENT
```

gslib/utils/user_agent_helper.py
```diff
@@ -0,0 +1,49 @@
...
+
+  user_agent = ' gsutil/%s' % gslib.VERSION
+  user_agent += ' (%s)' % sys.platform
+  user_agent += ' analytics/%s ' % ('disabled' if metrics_off else 'enabled')
+  user_agent += ' interactive/%s' % sys.stdin.isatty()
+
+  if len(opts) > 0:
+    user_agent += ' command/%s' % opts[0]
+    if len([segment for segment in opts if '://' in segment]) > 1:
+      user_agent += '-CloudToCloud'
```
dilipped, line 42: For a cp command, the source list can be a list of local file paths as well as cloud paths. In that case, won't we end up setting this as `CloudToCloud` for local file upload/download as well?
Also, it doesn't cover cases where a file path follows the uri `file:///...` which might be a file on a network but not necessarily on gcs or s3.

gslib/__main__.py
```diff
@@ -408,6 +398,9 @@ def main():
     else:
       command_name = args[0]
 
+    # Now that we have the command name, update the Boto user agent.
+    boto.UserAgent += ' command/%s' % command_name
```
dilipped, line 402: Is the command_name getting repeated?. We are already adding this info via `GetUserAgent`?

gslib/utils/user_agent_helper.py
thomasmaclean: This is a "cheap" way of figuring out if the operations are cloud to cloud with some significant limitations. Actually calculating the real destinations here seemed like it would be out of reach, given the complexity of that operation and the fact that it would need to happen right after bootstrap. I hadn't honestly considered the prospect of `file:///` prefixes, so I'll narrow the scope to only gs:// and s3:// prefixes.

gslib/__main__.py
thomasmaclean: Sorry I think you and I caught this around the same time.

gslib/utils/user_agent_helper.py
thomasmaclean: I like it! I had originally placed the GetArgumentsAndOptions call there but then just figured it made more sense to make it a helper and run it twice, but storing the user agent there makes a lot of sense.

gslib/utils/user_agent_helper.py
dilipped: Yeah, I can see that figuring out the CloudToCloud flag for each src destn pair might become complicated and given that the number of cases where users mix local path with cloud path in the source url list for the cp command might be low, this should be fine.

However, there is another case where all the source urls are cloud urls and the destination is local path. In that case, it will tag the operation as CloudToCloud. If we pass the args as mentioned here https://github.com/GoogleCloudPlatform/gsutil/pull/986#discussion_r409175426 then perhaps we can check if the the last item `args[-1]` is a cloud path and at the same time there is a cloud item in the source list `args[:-1]`.
Another trivial optimization would be to only check the first and last element in args, as we are making the assumption that user is not going to mix local and cloud path in the src list.

gslib/utils/user_agent_helper.py
```diff
@@ -0,0 +1,57 @@
...
+  """
+  user_agent = ' gsutil/%s' % gslib.VERSION
+  user_agent += ' (%s)' % sys.platform
+  user_agent += ' analytics/%s ' % ('disabled' if metrics_off else 'enabled')
+  user_agent += ' interactive/%s' % sys.stdin.isatty()
+
+  if len(args) > 0:
+    user_agent += ' command/%s' % args[0]
+
+    if args[0] in ['cp', 'rsync']:
```
dilipped, line 41: Do you also want to include `mv`?

gslib/utils/user_agent_helper.py
```diff
@@ -0,0 +1,57 @@
...
+      # Any cp or rsync commands that have both a source and destination in the
+      # cloud should be noted as that represents a unique use case that may be
+      # better served by the transfer service.
+      cloud_uri_pattern = '^(gs|s3)\://'
```
dilipped, line 45: Oops, I missed this, looks like unit tests are failing because of this line. Maybe change this to 
`cloud_uri_pattern = r'^(gs|s3)\://'`

dilipped: Even better
```
cloud_uri_pattern = re.compile(r'^(gs|s3)\://')
cloud_uris = [arg for arg in args if cloud_uri_pattern.search(arg)]
cloud_uri_dst = cloud_uri_pattern.search(args[-1])
```

gslib/utils/exception_helper.py
```diff
@@ -0,0 +1,63 @@
+import os
```
dilipped, line 1: Is this module still needed?

gslib/__main__.py
```diff
@@ -47,20 +53,32 @@
 import boto
 import gslib
 from gslib.utils import system_util
+from gslib.utils import text_util
```
dilipped, line 56: nit-pick: You can combine line 55 and 56 as 
```
from gslib.utils import system_util, text_util
```

PR #993, opened 2020-04-25
Fix incorrect rendered number of objects when using perfdiag -i

aubm:
Because the value passed to -n wasn't stored in the output file (when
-o is provided), running `perfdiag -i` with that file would
render inconsistent data for throughput tests.

Like in the example below, where 5 is default value for -n.

```
Copied 5 20 MiB file(s) for a total transfer size of 1.95 GiB.
```

For the record, when provided both -n and -i, perfdiag would render
something like in the example below, where 999 is value passed to -n.

```
Copied 999 20 MiB file(s) for a total transfer size of 1.95 GiB.
```

Now the number passed to -n is stored in the results, and used to
render the proper output. The previous behavior is preserved for
backward compatibility reasons, in case of running perfdiag -i with an
old file.

### DIFF
--- a/gslib/cloud_api_delegator.py
+++ b/gslib/cloud_api_delegator.py
@@ -159,8 +159,8 @@ class CloudApiDelegator(CloudApi):
       raise ArgumentException('No provider selected for CloudApi')
 
     if (selected_provider not in self.api_map[ApiMapConstants.DEFAULT_MAP] or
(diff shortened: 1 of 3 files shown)


NickGoog:
Can you add an example of a command that reproduces the issue?

NickGoog:
Closing because of inactivity

Issue #994, opened 2020-04-28
OAuth2 access token stored in world-readable file


mohamedhafez:
When I do the command `gcloud auth activate-service-account --key-file <KEYFILE>`, and then do a gsutil command like `gsutil ls gs://mybucket/`, it generates the `~/.gsutil/credstore2` file with the OAuth2 access token in it, but the file is world-readable:

```
sabu@myhost:~$ ls -l .gsutil/
total 8
-rw-rw-r-- 1 sabu sabu 4442 Apr 28 11:07 credstore2
-rw-rw-r-- 1 sabu sabu    0 Apr 28 11:07 credstore2.lock
```

mohamedhafez:
I believe the line causing the issue is [this one](https://github.com/googleapis/oauth2client/blob/50d20532a748f18e53f7d24ccbe6647132c979a9/oauth2client/contrib/multiprocess_file_storage.py#L110) in the oauth2client library creating the credential file with default permissions. 

I'd make an issue report there or even try to take a stab at fixing it (I don't know any python though), but the repo is now read only, and says its now deprecated and unmaintained...

mohamedhafez:
It looks like the code sets up a `MultiprocessFileStorage` instance [here](https://github.com/GoogleCloudPlatform/gsutil/blob/9af326229303d8636dcb6743b7d8c49f59f50d16/gslib/gcs_json_credentials.py#L152), and taking a look at the source for that in [the oauth2client library](https://github.com/googleapis/oauth2client), it looks like in the `acquire_lock` method it ends up calling `_create_file_if_needed`, which creates the credstore2 file with the default permissions [on this line](https://github.com/googleapis/oauth2client/blob/50d20532a748f18e53f7d24ccbe6647132c979a9/oauth2client/contrib/multiprocess_file_storage.py#L110). 

So I'd say this is actually a bug in the oauth2client library, and I would submit an issue there, but it says the project is read-only and is deprecated and no longer maintained...

mohamedhafez:
The tracker files that get generated for large uploads in `~/.gsutil/tracker-files` are world readable as well. I'm not sure if they contain sensitive info, it looks like there's an `upload_id` token in there though which might be

thomasmaclean:
I think umask would be a good straightforward solution.

mohamedhafez:
[...]

PR #1002, opened 2020-05-06
Fix oauth credfile permissions

dilipped:
Fixes https://github.com/GoogleCloudPlatform/gsutil/issues/994

Depends on https://github.com/gsutil-mirrors/oauth2client/pull/1

Since https://github.com/googleapis/oauth2client is now archived and in read-only mode, we have decided to move the library code to the `gslib/vendored` section.

### DIFF
--- a/gslib/wildcard_iterator.py
+++ b/gslib/wildcard_iterator.py
@@ -238,8 +238,8 @@ class CloudWildcardIterator(WildcardIterator):
             if obj_or_prefix.datatype == CloudApi.CsObjectOrPrefixType.OBJECT:
               gcs_object = obj_or_prefix.data
               if prog.match(gcs_object.name):
-                if not suffix_wildcard or (StripOneSlash(gcs_object.name)
-                                           == suffix_wildcard):
+                if not suffix_wildcard or (StripOneSlash(
+                    gcs_object.name) == suffix_wildcard):
                   if not single_version_request or (self._SingleVersionMatches(
                       gcs_object.generation)):
                     yield self._GetObjectRef(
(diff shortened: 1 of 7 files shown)


gslib/wildcard_iterator.py
```diff
@@ -238,8 +238,8 @@ def __iter__(self,
             if obj_or_prefix.datatype == CloudApi.CsObjectOrPrefixType.OBJECT:
               gcs_object = obj_or_prefix.data
               if prog.match(gcs_object.name):
-                if not suffix_wildcard or (StripOneSlash(gcs_object.name)
-                                           == suffix_wildcard):
+                if not suffix_wildcard or (StripOneSlash(
+                    gcs_object.name) == suffix_wildcard):
```
thomasmaclean, line 242: Assuming this isn't doing anything you can just remove this.

gsutil.py
```diff
@@ -74,7 +74,6 @@ def OutputAndExit(message):
     ('funcsigs', ''),  # mock dependency
     ('google-reauth-python', ''),  # Package name: google_reauth
     ('pyu2f', ''),  # google_reauth dependency
-    ('oauth2client', ''),
```
thomasmaclean, line 77: Why did you move this out of `third_party` and move it into vendored? You could just change the location of the original submodule and keep everything else set up in the same way.

gslib/wildcard_iterator.py
thomasmaclean: Assuming this isn't doing anything you can just remove this.

dilipped: Yep, these changes were made because I ran `python3 -m yapf -irp .`
But Travis CI failed anyway. Turns out, yapf behaves differently for python3.6 vs python3.7. Running in python3.6 removed this change so we should be good now.

gsutil.py
dilipped: This was done for PyPI. Because we are not maintaining our own version of oauth2client and not relying on the PyPI version, it was moved from here to `vendored_lib_dirs` in  `gslib.__init__.py` When gsutil gets installed via PyPI and invoked directly using `gslib.__main__.py` we might end up including the incorrect version.  The note on Line 91 explains this to some extent.

dilipped: To give some context, it is the same reason why boto is not included here, and is present under the `vendored_lib_dirs` section.

PR #1005, opened 2020-05-08
Fix file permissions for tracker files 

dilipped:
Fixes https://github.com/GoogleCloudPlatform/gsutil/issues/994

Tracker files were getting written out as world readable. This change sets the file mode to 600

### DIFF
--- a/gslib/tracker_file.py
+++ b/gslib/tracker_file.py
@@ -554,8 +554,7 @@ def WriteDownloadComponentTrackerFile(tracker_file_name, src_obj_metadata,
 def _WriteTrackerFile(tracker_file_name, data):
   """Creates a tracker file, storing the input data."""
   try:
-    fd = os.open(tracker_file_name,
-                 os.O_WRONLY | os.O_CREAT | os.O_TRUNC,
+    fd = os.open(tracker_file_name, os.O_WRONLY | os.O_CREAT | os.O_TRUNC,
                  0o600)
     with os.fdopen(fd, 'w') as tf:
       tf.write(data)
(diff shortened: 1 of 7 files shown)


gslib/tracker_file.py
```diff
@@ -554,14 +554,28 @@ def WriteDownloadComponentTrackerFile(tracker_file_name, src_obj_metadata,
 def _WriteTrackerFile(tracker_file_name, data):
   """Creates a tracker file, storing the input data."""
   try:
-    with os.fdopen(os.open(tracker_file_name, os.O_WRONLY | os.O_CREAT, 0o600),
-                   'w') as tf:
+    fd = os.open(tracker_file_name, os.O_WRONLY | os.O_CREAT | os.O_TRUNC,
+                 0o600)
```
thomasmaclean, line 558: so the major change here is that you're using `os.O_TRUNC` to empty the file, yes?

gslib/tracker_file.py
```diff
@@ -554,14 +554,28 @@ def WriteDownloadComponentTrackerFile(tracker_file_name, src_obj_metadata,
...
+  """Create a tracker file and write json data to it.
+
+  Raises:
+    TypeError: If the data is not JSON serializable
+  """
+  try:
+    json_str = json.dumps(data)
+  except TypeError as err:
+    raise RaiseUnwritableTrackerFileException(tracker_file_name, err.strerror)
+  _WriteTrackerFile(tracker_file_name, json_str)
```
thomasmaclean, line 576: I believe this should be inside the above `try:` right?

gslib/tracker_file.py
dilipped: The major change for the entire PR is making `_UploadTrackerCallback` call this function instead of trying to write the tracker file directly.

For this particular change specifically, you are correct, the os.)_TRUNC is the major change. This was added because I saw some unpredictable behavior if the file is not emptied.

gslib/tracker_file.py
dilipped: I intentionally kept _WriteTrackerFile outside because it has it's own try/except block. I wanted to isolate the errors that could possibly generate from `json.dumps`.

PR #1006, opened 2020-05-08
Cleanup and formatting for cp command 

karenarialin:
Also add clarification for -I option


gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
-  If you want to copy an entire directory tree you need to use the -r option.
-  For example, to upload the directory tree "dir":
+  Use the ``-r`` option to copy an entire directory tree.
+  For example, to upload the directory tree ``dir``:
 
     gsutil cp -r dir gs://my-bucket
 
-  If you have a large number of files to transfer you might want to use the
-  top-level gsutil -m option (see "gsutil help options"), to perform a
-  parallel (multi-threaded/multi-processing) copy:
```
starsandskies, line 102: I recommend keeping this - it was added because of b/75005961, and I've seen other comments in the past from people confused about -m being top-level and not command-specific

starsandskies: also, this renders as a link in our documentation, in case you weren't aware of that behavior

gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
-  NOTE: Shells (like bash, zsh) sometimes attempt to expand wildcards in ways
-  that can be surprising. Also, attempting to copy files whose names contain
-  wildcard characters can result in problems. For more details about these
-  issues see the section "POTENTIALLY SURPRISING BEHAVIOR WHEN USING WILDCARDS"
-  under "gsutil help wildcards".
+  NOTE: Shells like ``bash`` and ``zsh`` sometimes attempt to expand
+  wildcards in ways that can be surprising. You may also encounter issues when
+  attempting to copy files whose names contain wildcard characters. For more
+  details about these issues, see "Potentially Surprising Behavior When Using Wildcards"
+  under ``gsutil help wildcards``.
```
starsandskies, line 123: I believe keeping these in quotations (instead of ``) will render the contents as a link?

gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
 """
 
 _NAME_CONSTRUCTION_TEXT = """
 <B>HOW NAMES ARE CONSTRUCTED</B>
-  The gsutil cp command strives to name objects in a way consistent with how
-  Linux cp works, which causes names to be constructed in varying ways depending
+  The ``gsutil cp`` command names objects in ways that are consistent with the
```
starsandskies, line 128: I'm guessing we should include the qualifier "attempts to name" since it probably is not always able to do so.

gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
+  ``gs://my-bucket/c``:
 
     gsutil cp dir1/dir2/** gs://my-bucket
 
-  will create the object gs://my-bucket/c.
-
-  The same rules apply for downloads: recursive copies of buckets and
+  Note that in the above example, the '**' wildcard matches all names
+  anywhere under ``dir``. The wildcard '*' matches names just one level deep. For
+  more details, see ``gsutil help wildcards``.
```
starsandskies, line 150: Similar to the above comment, I think we want " instead of `` so that a link renders in c.g.c documentation

gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
-  This will cause everything nested under gs://my-bucket/data to be downloaded
-  into dir, resulting in files with names like dir/data/a/b/c.
+  This causes everything nested under ``gs://my-bucket/data`` to be downloaded
+  into ``dir``, resulting in files with names like ``dir/data/a/b/c``.
 
   Copying subdirectories is useful if you want to add data to an existing
   bucket directory structure over time. It's also useful if you want
   to parallelize uploads and downloads across multiple machines (potentially
-  reducing overall transfer time compared with simply running gsutil -m
+  reducing overall transfer time compared with running ``gsutil -m``
```
starsandskies, line 195: I think the closing `` comes too early, and that cp (on the next line) should be part of the code block

gslib/commands/cp.py
```diff
@@ -81,194 +81,188 @@
...
-  corrupted object and print a warning message. This very rarely happens, but
-  if it does, please contact gs-team@google.com.
-
-  If you know the MD5 of a file before uploading you can specify it in the
-  Content-MD5 header, which will cause the cloud storage service to reject the
+  At the end of every upload or download, the ``gsutil cp`` command validates that
+  the checksum it computes for the source file matches the checksum that
+  the service computes. If the checksums do not match, gsutil deletes the
+  corrupted object and prints a warning message. If this happens, contact
+  ``gs-team@google.com``.
```
starsandskies, line 262: Not a code block?

gslib/commands/cp.py
```diff
@@ -282,82 +276,79 @@
     Copying file://obj [Content-Type=text/plain]...
     Uploading   gs://your-bucket/obj:                                182 b/182 B
 
-    If the checksum didn't match the service would instead reject the upload and
-    gsutil would print a message like:
+    If the checksums don't match, the service rejects the upload and
+    gsutil prints the following message:
```
starsandskies, line 280: Recommend keeping "like" or "similar to" since the message the user receives will not be identical to the one we give here

gslib/commands/cp.py
```diff
@@ -282,82 +276,79 @@
...
 
     BadRequestException: 400 Provided MD5 hash "VgyllJgiiaRAbyUUIqDMmw=="
     doesn't match calculated MD5 hash "7gyllJgiiaRAbyUUIqDMmw==".
 
-  Even if you don't do this gsutil will delete the object if the computed
```
starsandskies, line 291: Do you think this automatic behavior doesn't warrant noting?

karenarialin: I found it redundant, since the very first paragraph in the section says "If the checksums do not match, gsutil deletes the corrupted object and prints a warning message." In the following discussion, I didn't find any indication that specifying Content-MD5 contradicts that statement. In fact, the bullet points after this are variations on the assumptions that "gsutil deletes objects with mismatched checksums." So I didn't feel the need to reiterate that point here. If you feel that there's something missing here, let me know.

gslib/commands/cp.py
```diff
@@ -282,82 +276,79 @@
...
-  checksum mismatches, but specifying the Content-MD5 header has several
-  advantages:
+  Specifying the Content-MD5 header has several advantages:
 
-  1. It prevents the corrupted object from becoming visible at all, whereas
-     otherwise it would be visible for 1-3 seconds before gsutil deletes it.
+  1. It prevents the corrupted object from becoming visible. Otherwise, the
```
starsandskies, line 287: Suggestion: Replace "otherwise" with "If you don't specify the header"

gslib/commands/cp.py
```diff
@@ -282,82 +276,79 @@
...
+     Content-MD5 header prevents the existing object from being replaced.
+     Otherwise, the existing object is replaced by the corrupted object and
      deleted a few seconds later.
 
-  3. It will definitively prevent the corrupted object from being left in
-     the cloud, whereas the gsutil approach of deleting after the upload
-     completes could fail if (for example) the gsutil process gets ^C'd
-     between upload and deletion request.
+  3. If the gsutil process is interrupted or fails before it can delete the
+     corrupted object after uploading it, specifying the Content-MD5 header
```
starsandskies, line 296: Suggestion: If you don't specify the Content-MD5 header, it's possible for the gsutil process to complete the upload but then be interrupted or fail before it can delete the corrupted object, leaving the corrupted object in the cloud.

gslib/commands/cp.py
```diff
@@ -373,43 +364,41 @@
   or service problems.
 
   Streaming transfers using the XML API do not support resumable
-  uploads/downloads. If you have a large amount of data to upload (say, more
-  than 100 MiB) it is recommended that you write the data to a local file and
-  then copy that file to the cloud rather than streaming it (and similarly for
-  large downloads).
+  uploads or downloads. If you have a large amount of data to upload or download,
+  over 100 MiB for example, we recommend that you write the data to a file and
```
starsandskies, line 368: It feels like including "local" in "local file" probably is important here.

gslib/commands/cp.py
```diff
@@ -451,176 +445,166 @@
...
-  make sense to enable parallel composite uploads on your machine (there are
-  efficient CRC32C implementations available in Java).
-
-  If a parallel composite upload fails prior to composition, re-running the
-  gsutil command will take advantage of resumable uploads for the components
-  that failed, and the component objects will be deleted after the first
+  upload complete. If you want to enable parallel composite
+  uploads for all of your future uploads, you can uncomment and set the
+  "parallel_composite_upload_threshold" config value in your ``.boto`` configuration
+  file to 150M.
```
starsandskies, line 457: "file to 150M, or your desired value"?

gslib/commands/cp.py
```diff
@@ -451,176 +445,166 @@
...
-  points apply to the cp command.
+  While that section refers to the ``rsync`` command, analogous
+  points apply to the ``cp`` command.
 """
 
 _OPTIONS_TEXT = """
 <B>OPTIONS</B>
-  -a canned_acl  Sets named canned_acl when uploaded objects created. See
-                 "gsutil help acls" for further details.
+  -a canned_acl  Sets the specific ``canned_acl`` when the uploaded objects are
```
starsandskies, line 532: Suggestion: "Applies the specific ``canned_acl`` to uploaded objects"

gslib/commands/cp.py
```diff
@@ -451,176 +445,166 @@
...
+                 created. See "gsutil help acls" for further details.
 
-  -A             Copy all source versions from a source buckets/folders.
+  -A             Copy all source versions from a source buckets or folder.
```
starsandskies, line 535: buckets --> bucket

gslib/commands/cp.py
```diff
@@ -451,176 +445,166 @@
...
                  composite object into a non-composite object. For example:
 
                      gsutil cp -D -p gs://bucket/obj gs://bucket/obj_tmp
                      gsutil mv -p gs://bucket/obj_tmp gs://bucket/obj
 
-                 NOTE: Daisy chain mode is automatically used when copying
-                 between providers (e.g., to copy data from Google Cloud Storage
-                 to another provider).
+                 NOTE: "Daisy chain" mode is automatically used when copying
+                 between providers: for example, when copying data from Google Cloud Storage
```
starsandskies, line 566: Nit: Cloud Storage

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
                  - Source size.
                  - Bytes transferred.
                  - MD5 hash.
-                 - UTC date and time transfer was started in ISO 8601 format.
-                 - UTC date and time transfer was completed in ISO 8601 format.
+                 - UTC date and transfer start time in ISO 8601 format.
```
starsandskies, line 618: I think you'll want to keep the "date" and "time" portion of these lines grouped together. Read literally, the proposed wording could "The current UTC date, as well as the time the transfer was started/completed"

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
...
-                 The -c option will cause copying to continue after failures
-                 occur, and the -L option will allow gsutil to pick up where it
-                 left off without duplicating work. The loop will continue
-                 running as long as gsutil exits with a non-zero status (such a
-                 status indicates there was at least one failure during the
-                 gsutil run).
+                 The -c option enables copying to continue after failures
+                 occur, and the -L option allows gsutil to pick up where it
+                 left off without duplicating work. The loop continues
+                 running as long as gsutil exits with a non-zero status, which
```
starsandskies, line 641: Recommend making this along the lines of "The loop continues running as long as gsutil exits with a non-zero status. A non-zero status indicates...:

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
...
-                 only works if you have OWNER access to all of the objects that
-                 are copied.
-
-                 You can avoid the additional performance and cost of using
-                 cp -p if you want all objects in the destination bucket to end
-                 up with the same ACL by setting a default object ACL on that
-                 bucket instead of using cp -p. See "gsutil help defacl".
-
-                 Note that it's not valid to specify both the -a and -p options
+  -p             Enables ACLs to be preserved when copying in the cloud. Note
```
starsandskies, line 657: "Preserves ACLs when copying in the cloud"

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
...
+                 that this option has performance and cost implications only when
+                 using the XML API, as it requires separate HTTP calls for
```
starsandskies, line 659: Recommend "it" --> "the XML API"

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
...
+                 performance issues by setting a default object ACL on that bucket
+                 instead of using ``cp -p``. See "gsutil help defacl".
+
+                 Note that it's not valid to specify both the ``-a`` and ``-p`` options
                  together.
 
-  -P             Causes POSIX attributes to be preserved when objects are
-                 copied. With this feature enabled, gsutil cp will copy fields
-                 provided by stat. These are the user ID of the owner, the group
-                 ID of the owning group, the mode (permissions) of the file, and
```
starsandskies, line 694: Recommend retaining "mode" here, or removing reference to it in the next paragraph

gslib/commands/cp.py
```diff
@@ -631,147 +615,141 @@
...
 
   -z <ext,...>   Applies gzip content-encoding to any file upload whose
-                 extension matches the -z extension list. This is useful when
-                 uploading files with compressible content (such as .js, .css,
-                 or .html files) because it saves network bandwidth and space
-                 in Google Cloud Storage, which in turn reduces storage costs.
+                 extension matches the ``-z`` extension list. This is useful when
+                 uploading files with compressible content such as .js, .css,
+                 or .html files, because it reduces network bandwidth and storage
+                 costs.
```
starsandskies, line 708: "because it reduces network bandwidth and storage sizes. This can improve both performance and reduce costs."

gslib/commands/cp.py
starsandskies: Reviewing the text again, I agree with you!

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Done.

gslib/commands/cp.py
karenarialin: Noted. I was thinking Google Cloud Storage vs other cloud storage providers.

starsandskies:
LGTM

Wow! Great improvements :)

PR #1008, opened 2020-05-09
Log GCS service-account key used to sign urls.

mhandugan:
When --use-service-account (-u) is used to sign URLs, allow debug
logging to record the Key ID used. This can then be looked up with
'gcloud iam service-accounts keys list' to find its expiration date.

b/154569245


dilipped:
@mhandugan Thanks for the PR. Looks like Tavis CI check is failing because of a linting issue. Can you apply the fix that has been suggested here https://travis-ci.org/github/GoogleCloudPlatform/gsutil/jobs/684930109

dilipped:
FYI, I moved this debug line a level up in the method call hierarchy so that we are able to log the message for non-impersonated service account as well. https://github.com/GoogleCloudPlatform/gsutil/pull/1010

PR #1011, opened 2020-05-13
Note that rewrite -O option should be used for ubla buckets

### LINKED ISSUE #980: Can't gsutil rewrite storage class with uniform bucket-level access: "No OWNER permission found"
Hey there, I'm trying to move some objects from `STANDARD` to `NEARLINE` storage. `gsutil rewrite` should support this, but I get this error:

```
$ gsutil rewrite -s NEARLINE gs://nz-wireless-map/prism.csv/\*\*
CommandException: No OWNER permission found for object gs://nz-wireless-map/prism.csv/2020-02-02T22:48:03Z. OWNER permission is required for rewriting objects, (otherwise their ACLs would be reset).
```

I think the problem is in [rewrite.py](https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/commands/rewrite.py#L390), which errors if we don't have a `src_metadata.acl`:

```
    elif not src_metadata.acl:
      raise CommandException(
          'No OWNER permission found for object %s. OWNER permission is '
          'required for rewriting objects, (otherwise their ACLs would be '
          'reset).' % transform_url)
```

I think my bucket doesn't have an ACL (or at least a legacy ACL): 

```
$ gsutil acl get gs://nz-wireless-map/prism.csv/2020-02-02T22:48:03Z
BadRequestException: 400 Cannot get legacy ACL for an object when uniform bucket-level access is enabled. Read more at https://cloud.google.com/storage/docs/uniform-bucket-level-access.
```

My request is to be able to run `gsutil rewrite` on my bucket with uniform bucket-level access. I think this will affect all new buckets - it seems like the ACL is legacy, so it seems likely that `gsutil rewrite` isn't working for any new buckets.

thomasmaclean:
fixes #980

### DIFF
--- a/gslib/commands/rewrite.py
+++ b/gslib/commands/rewrite.py
@@ -172,10 +172,12 @@ _DETAILED_HELP_TEXT = ("""
                 or CMEK). See 'gsutil help encryption' for details on encryption
                 configuration.
 
-  -O            Rewrite objects with the bucket's default object ACL instead of
-                the existing object ACL. This is needed if you do not have
-                OWNER permission on the object or if your bucket has uniform
-                bucket-level access (UBLA) enabled.
+  -O            When a bucket has uniform bucket-level access (UBLA) enabled,
+                the -O flag is required and will skip all ACL checks. When a
+                bucket has UBLA disabled, the -O flag rewrites objects with the
+                bucket's default object ACL instead of the existing object ACL.
+                This is needed if you do not have OWNER permission on the
+                object.
 
   -R, -r        The -R and -r options are synonymous. Causes bucket or bucket
                 subdirectory contents to be rewritten recursively.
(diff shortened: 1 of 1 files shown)


gslib/commands/rewrite.py
```diff
@@ -174,7 +174,8 @@
 
   -O            Rewrite objects with the bucket's default object ACL instead of
                 the existing object ACL. This is needed if you do not have
-                OWNER permission on the object.
+                OWNER permission on the object or if your bucket has uniform
+                bucket-level access (UBLA) enabled.
```
starsandskies, line 178: In the case of UBLA, I'm assuming -O does something besides rewrite using the bucket's default object ACL, since my understanding is default object ACLs doin't exist in UBLA buckets? I'd recommend having two parts to this flag, something along the lines of:

"When a bucket has uniform bucket-level access (UBLA) enabled, the -O flag is required and [FILL IN THE BLANK REGARDING BEHAVIOR]. When a bucket has UBLA disabled, the -O flag rewrites objects with the bucket's default object ACL instead of the existing object ACL. This is needed if you do not have OWNER permission on the object."

gslib/commands/rewrite.py
```diff
@@ -391,7 +392,8 @@ def RewriteFunc(self, name_expansion_result, thread_state=None):
       raise CommandException(
           'No OWNER permission found for object %s. OWNER permission is '
           'required for rewriting objects, (otherwise their ACLs would be '
```
starsandskies, line 394: Suggestion: "No OWNER permission found for object %s. If your bucket has uniform bucket-level access (UBLA) enabled, include the -O option in your command to avoid this error. If your bucket does not use UBLA, you can use the -O option to apply the bucket's default object ACL when rewriting."

PR #1014, opened 2020-05-13
Signurl: Restrict duration to 12 hours when -u flag is used

dilipped:
When -u flag is used, the SignBlob API is called, which guarantees the key's lifespan for up to 12 hours. This fix adds a restriction on the max duration allowed when the -u flag is used.

More info b/156160482

### DIFF
--- a/gslib/commands/signurl.py
+++ b/gslib/commands/signurl.py
@@ -157,8 +157,8 @@ _DETAILED_HELP_TEXT = ("""
   -u           Use service account credentials instead of a private key file
                to sign the url.
 
-               You can also use the --use-service-account option,
-               which is equivalent to -u.
+               You can also use the ``--use-service-account`` option,
+               which is equivalent to ``-u``.
                Note that both options have a maximum allowed duration of
                12 hours for a valid link.
 
--- a/gslib/commands/signurl.py
+++ b/gslib/commands/signurl.py
@@ -129,12 +129,12 @@ _DETAILED_HELP_TEXT = ("""
                the duration the link remains valid is the sum of all the
                duration options.
 
-               The max duration allowed is 7d when private-key-file is used.
+               The max duration allowed is 7 days when ``private-key-file``
+               is used.
 
                The max duration allowed is 12 hours when -u option is used.
-               This limitation exists because the system managed key used for
-               signing the url is guaranteed to remain valid for
-               at least 12 hours.
+               This limitation exists because the system managed key used to
+               sign the url is guaranteed to remain valid for at least 12 hours.
 
   -c           Specifies the content type for which the signed url is
                valid for.
@@ -158,8 +158,8 @@ _DETAILED_HELP_TEXT = ("""
                to sign the url.
 
                You can equivalently use --use-service-account instead of -u.
-               Note that the max alloweod duration is 12 hours if this option
-               is used.
+               The max alloweod duration for the valid link is 12 hours
+               if you use --use-service-account or -u option.
 
 <B>USAGE</B>
   Create a signed url for downloading an object valid for 10 minutes:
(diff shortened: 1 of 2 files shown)


gslib/commands/signurl.py
```diff
@@ -151,7 +157,9 @@
   -u           Use service account credentials instead of a private key file
                to sign the url.
 
-               You can equivalently use --use-service-account instead of -u
+               You can equivalently use --use-service-account instead of -u.
+               Note that the max alloweod duration is 12 hours if this option
+               is used.
```
thomasmaclean, line 162: Might want to have @starsandskies weigh in on this one as there may be a better way to say this.

karenarialin: "You can use --use-service-account instead of -u to [do this]." where [do this] is the benefit of using --use-service-account. If there's no difference at all between the two commands, maybe it would be more efficient to say "-u, --use-service-account" in the first column.

Is the max duration 12h for both -u and --use-service-account? If so, I suggest "The maximum allowed duration for the valid link is 12 hours if you use the --use-service-account or -u option."

gslib/commands/signurl.py
```diff
@@ -128,7 +129,12 @@
                the duration the link remains valid is the sum of all the
                duration options.
 
-               The max duration allowed is 7d.
+               The max duration allowed is 7d when private-key-file is used.
```
karenarialin, line 132: Expand to 7 days

gslib/commands/signurl.py
```diff
@@ -128,7 +129,12 @@
...
+
+               The max duration allowed is 12 hours when -u option is used.
+               This limitation exists because the system managed key used for
```
karenarialin, line 135: "system-managed key used to sign the url"

gslib/commands/signurl.py
karenarialin: code format private-key-file.

gslib/commands/signurl.py
dilipped: Regarding "You can equivalently use --use-service-account instead of -u", yes, there is no difference between the two options. We tried to fit it in the first column last time (cl/303730086) but it messes up with the formatting because of the length of --use-service-account.

gslib/commands/signurl.py
dilipped: Done

gslib/commands/signurl.py
dilipped: Done

gslib/commands/signurl.py
karenarialin: I see, thanks for the context. How about "You can also use the --use-service-account option, which is equivalent to -u. Note that both options have a maximum allowed duration of 12 hours for a valid link."

And it would be helpful to code format the options.

gslib/commands/signurl.py
```diff
@@ -128,7 +129,12 @@
                the duration the link remains valid is the sum of all the
                duration options.
 
-               The max duration allowed is 7d.
+               The max duration allowed is 7 days when ``private-key-file``
+               is used.
+
+               The max duration allowed is 12 hours when -u option is used.
+               This limitation exists because the system managed key used to
```
starsandskies, line 136: Nit: I believe it should be "system-managed"

gslib/commands/signurl.py
```diff
@@ -128,7 +129,12 @@
...
+               sign the url is guaranteed to remain valid for at least 12 hours.
```
starsandskies, line 137: Should this be "at most 12 hours"?

dilipped: It should be "at least". Basically, the key is guaranteed to remain valid for at least 12 hours. After 12 hours, the key may or may not be valid. Happy to rephrase this if it sounded confusing and if there are other suggestion :)

starsandskies: Possible wordsmithing: "This limitation exists because the system-managed key used to sign the url is rotated regularly, with 12 hours being the lower limit of time that the system-managed key is valid"

dilipped: Talked to @starsandskies offline and we decided on an alternate wording for this.

PR #1016, opened 2020-05-18
rsync now tries to patch before overwriting, rather than checking ACL

thomasmaclean:
This fixes a bug with `rsync` where it was previously checking ACL's to determine whether a patch or rewrite should be applied. This didn't work at all in most cases and especially didn't work for UBLA buckets where object ACL's don't exist. Now we'll simply attempt a patch first before falling back to rewrites.

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -39,6 +39,7 @@ from boto import config
 import crcmod
 from gslib.bucket_listing_ref import BucketListingObject
 from gslib.cloud_api import NotFoundException
+from gslib.cloud_api import ServiceException
 from gslib.command import Command
 from gslib.command import DummyArgChecker
 from gslib.command_argument import CommandArgument
@@ -1471,7 +1472,8 @@ def _RsyncFunc(cls, diff_to_apply, thread_state=None):
                                          obj_metadata,
                                          provider=dst_url.scheme,
                                          generation=dst_url.generation)
-        except:
+        except ServiceException as err:
+          cls.logger.debug('Error while trying to patch: %s', err)
           # We don't have permission to patch apparently, so it must be copied.
           cls.logger.info(
               'Copying whole file/object for %s instead of patching'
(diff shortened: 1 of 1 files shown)


gslib/commands/rsync.py
```diff
@@ -1460,24 +1464,18 @@ def _RsyncFunc(cls, diff_to_apply, thread_state=None):
...
+        try:
+          # Assume we have permission, and can patch the object.
           gsutil_api.PatchObjectMetadata(dst_url.bucket_name,
                                          dst_url.object_name,
                                          obj_metadata,
                                          provider=dst_url.scheme,
                                          generation=dst_url.generation)
-        else:
-          # We don't have object ownership, so it must be copied.
+        except:
```
dilipped, line 1474: Blanket exceptions can be risky. Looks like we should expect this to raise ServiceException so we can try catching that and add the exception message in debug log. https://github.com/GoogleCloudPlatform/gsutil/blob/11b9f1f4cc160180b85227ccf1df40b0797dab0e/gslib/cloud_api.py#L418
```suggestion
        except ServiceException as err:
          cls.logger.debug('Error while trying to patch: %s', err)
```

gslib/commands/rsync.py
```diff
@@ -1512,18 +1510,18 @@ def _RsyncFunc(cls, diff_to_apply, thread_state=None):
...
+        try:
+          # Assume we have ownership, and can patch the object.
           gsutil_api.PatchObjectMetadata(dst_url.bucket_name,
                                          dst_url.object_name,
                                          obj_metadata,
                                          provider=dst_url.scheme,
                                          generation=dst_url.generation)
-        else:
-          # We don't have object ownership, so it must be copied.
+        except:
```
dilipped, line 1520: same as above

gslib/commands/rsync.py
```diff
@@ -1734,6 +1733,8 @@ def _ParseOpts(self):
           gzip_arg_all = GZIP_ALL_FILES
         elif o == '-n':
           self.dryrun = True
+        elif o == '-O':
+          self.assert_ownership = True
```
dilipped, line 1737: Where are we using this flag?

gslib/commands/rsync.py
```diff
@@ -446,6 +447,10 @@
                  what would be copied or deleted without actually doing any
                  copying/deleting.
 
+  -O             Assert ownership of objects. This is useful if uniform bucket-
```
dilipped, line 450: I see that parsing for this option was removed in the last commit so we should be removing this help text as well?

gslib/commands/rsync.py
```diff
@@ -1690,6 +1691,7 @@ def _ParseOpts(self):
     # state rather than CopyHelperOpts.
     self.continue_on_error = False
     self.delete_extras = False
+    self.assert_ownership = False
```
dilipped, line 1694: Should be removed?

PR #1018, opened 2020-05-18
Small doc edits.

starsandskies:
This backfills changes initially proposed in cl/310599519:

* Improved linking.
* Documenting HTTP 408 as retryable.
* Clarifying part of the `gsutil acl` synopsis.

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -126,8 +126,8 @@ _DETAILED_HELP_TEXT = ("""
     gsutil rsync -r data gs://mybucket/data
 
   If you have a large number of objects to synchronize you might want to use the
-  gsutil -m option, to perform parallel (multi-threaded/multi-processing)
-  synchronization:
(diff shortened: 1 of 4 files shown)


gslib/commands/rsync.py
```diff
@@ -126,8 +126,8 @@
     gsutil rsync -r data gs://mybucket/data
 
   If you have a large number of objects to synchronize you might want to use the
-  gsutil -m option, to perform parallel (multi-threaded/multi-processing)
-  synchronization:
+  gsutil -m option (see "gsutil help options"), to perform parallel
```
dilipped, line 129: The CL shows this as a link, do we want to add the link here?

starsandskies: We'll have to see how the generation process goes, but my understanding is there's logic in the generator that converts this into a link. If not, I'll definitely come back to this to make sure it does.

dilipped: Ah I see. Sounds good.

PR #1029, opened 2020-06-05
Added options section to the Kms command

piotrci:
This addresses issue: https://issuetracker.google.com/issues/151814248

### DIFF
--- a/gslib/commands/kms.py
+++ b/gslib/commands/kms.py
@@ -108,8 +108,9 @@ _ENCRYPTION_DESCRIPTION = """
 
 <B>OPTIONS</B>
   -k <key>      Set the default KMS key for my-bucket using the
-                full path to the key <code>projects/[project-id]/locations/[location]/keyRings/[key-
-                ring]/cryptoKeys/[my-key]</code>
+                full path to the key `projects/[project-id]/locations/[location]/keyRings/[key-
+                ring]/cryptoKeys/[my-key]`
+                
   -w            (used with -k key) Display a warning rather than 
                 failing if gsutil is unable to verify that
                 the specified key contains the correct IAM bindings 
--- a/gslib/commands/kms.py
+++ b/gslib/commands/kms.py
@@ -105,6 +105,21 @@ _ENCRYPTION_DESCRIPTION = """
   Clear the default KMS key so newly-written objects will not be encrypted:
 
     gsutil kms encryption -d gs://my-bucket
+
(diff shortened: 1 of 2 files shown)


gslib/commands/kms.py
```diff
@@ -105,6 +105,21 @@
   Clear the default KMS key so newly-written objects will not be encrypted:
 
     gsutil kms encryption -d gs://my-bucket
+
+<B>OPTIONS</B>
+  -k <key>      Set the default KMS key for my-bucket using the
+                full path to the key <code>projects/[project-id]/locations/[location]/keyRings/[key-
+                ring]/cryptoKeys/[my-key]</code>
+  -w            (used with -k key) Display a warning rather than
```
starsandskies, line 113: I'm not sure if it matters, but to match formatting of other pages, I think there should be a newline between the -k entry and the -w entry

gslib/commands/kms.py
```diff
@@ -105,6 +105,21 @@
   Clear the default KMS key so newly-written objects will not be encrypted:
 
     gsutil kms encryption -d gs://my-bucket
+
+<B>OPTIONS</B>
+  -k <key>      Set the default KMS key for my-bucket using the
+                full path to the key <code>projects/[project-id]/locations/[location]/keyRings/[key-
```
starsandskies, line 111: Replace <code> and </code> with ``

gslib/commands/kms.py
```diff
@@ -105,6 +105,21 @@
...
+                ring]/cryptoKeys/[my-key]</code>
+  -w            (used with -k key) Display a warning rather than 
+                failing if gsutil is unable to verify that
+                the specified key contains the correct IAM bindings 
+                for encryption/decryption. This is useful for
+                users that do not have getIamPolicy permission 
+                but know that the key has the correct IAM policy
+                for encryption in the user's project.
+     
+  -d            Clear the default KMS key so newly-written objects will not be encrypted.
```
starsandskies, line 121: "Clear the default KMS key from the bucket."

I'd recommend against "newly-written objects will not be encrypted" (and I'd recommended changing the wording above), because newly-written objects definitely will be encrypted, just not with the default KMS.

starsandskies:
I think this looks good to go (I'm not sure how the multi-line code block will render, but we can cross that bridge when we get to it).

LGTM, though I'd like to leave the final approval to Dilip or Thomas.

PR #1030, opened 2020-06-05
Adding clarification regarding gsutil iam and it's -r option

piotrci:
This addresses issue: https://issuetracker.google.com/issues/122078625


starsandskies:
This looks good to me, though I'd like to defer to Dilip or Thomas as the owners of the repository.

NickGoog:
Did some quick tests, and this seems correct!

PR #1032, opened 2020-06-08
Improve rsync error text

NickGoog:
http://b/148476840

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -1594,8 +1594,8 @@ class RsyncCommand(Command):
         logger=self.logger)
     if not have_existing_container:
       raise CommandException(
-          'arg (%s) does not name a directory, bucket, or bucket subdir.' %
-          url_str)
+          'arg (%s) does not name a directory, bucket, or bucket subdir.\n'
+          'Did you forget a trailing "/"?' % url_str)
     return url
 
   def RunCommand(self):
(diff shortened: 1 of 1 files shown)


gslib/commands/rsync.py
```diff
@@ -1594,8 +1594,8 @@ def _InsistContainer(self, url_str, treat_nonexistent_object_as_subdir):
         logger=self.logger)
     if not have_existing_container:
       raise CommandException(
-          'arg (%s) does not name a directory, bucket, or bucket subdir.' %
-          url_str)
+          'arg (%s) does not name a directory, bucket, or bucket subdir.\n'
+          'Did you forget a trailing "/"?' % url_str)
```
dilipped, line 1598: I feel a trailing / should not matter. Were you able to reproduce this error. I tried it locally and for me the command mentioned in the bug works fine. 

```
./gsutil rsync -r ~/temp/junk/three gs://gsutil_test_bucket-5/three
Building synchronization state...
Starting synchronization...
Copying file:///usr/local/google/home/dpednekar/temp/junk/three/a.txt [Content-Type=text/plain]...
Copying file:///usr/local/google/home/dpednekar/temp/junk/three/b.txt [Content-Type=text/plain]...
/ [2 files][    0.0 B/    0.0 B]
Operation completed over 2 objects.
```

NickGoog: Check out this public docs page. You can skip the diagram and go straight to step one: https://cloud.google.com/storage/docs/gsutil/addlhelp/HowSubdirectoriesWork

dilipped: Yeah I have seen that link. What I mean to say is that, this does not seem like a problem because of a missing slash, because the rsync command should work even if the slash is missing.  I.e if you specify a directory without a slash, rsync command should be able to treat it as a directory. There might be something else going on there and I am not able to reproduce the problem specified in the bug.

PR #1033, opened 2020-06-09
rm will continue on object 404s

thomasmaclean:


gslib/commands/rm.py
```diff
@@ -393,10 +393,15 @@ def RemoveFunc(self, name_expansion_result, thread_state=None):
...
-                            provider=exp_src_url.scheme)
+    try:
+      gsutil_api.DeleteObject(exp_src_url.bucket_name,
+                              exp_src_url.object_name,
+                              preconditions=self.preconditions,
+                              generation=exp_src_url.generation,
+                              provider=exp_src_url.scheme)
+    except NotFoundException as e:
+      # Retries during recursive object deletes can cause harmless 404s.
+      self.logger.info('Cannot find %s', exp_src_url)
```
dilipped, line 404: Curious about why we want to retry on 404. Can you please add a link to the issue this is trying to fix?

thomasmaclean: b/148433922

gslib/commands/rm.py
```diff
@@ -393,10 +393,15 @@ def RemoveFunc(self, name_expansion_result, thread_state=None):
...
+      DecrementFailureCount()
```
dilipped, line 405: Why do we have call this? Is DeleteObject call going to increment the failure count?

thomasmaclean: `command.py` increments failure count in `PerformTask`. This score keeping would be off by one when the 404 is harmless. I'll add a slightly more verbose comment to explain why this happens.

PR #1035, opened 2020-06-09
Update linter to use python3 for better forward compatibility.

NickGoog:
Makes developer environment setup a bit simpler because our yapf formatter uses "python3", but lylint was using "python" (version 2).


thomasmaclean:
You'll notice you have some Travis failures. I think you also need to make a change to `.travis.yml`. Something like:

```
  - if [[ "$TRAVIS_PYTHON_VERSION" =~ 3.6 ]]; then yapf -dr . ; fi
  - if [[ "$TRAVIS_PYTHON_VERSION" =~ 3.6 ]]; then test/run_pylint.sh
```

NickGoog:
@dilipped wanted to hold off on this change because we're still supporting Python2, so I was just going to let this PR sit for now.

PR #1043, opened 2020-06-12
Better error message for cp with empty source directory

NickGoog:
Make error message more clear for case where user tries to copy, and source files do not exist.
http://b/150820007

### DIFF
--- a/gslib/exception.py
+++ b/gslib/exception.py
@@ -31,8 +31,10 @@ from __future__ import unicode_literals
 
 import six
 
-NO_URLS_MATCHED_GENERIC = 'No URLs matched'
-NO_URLS_MATCHED_TARGET = 'No URLs matched: %s'
+NO_URLS_MATCHED_PREFIX = 'No URLs matched'
+NO_URLS_MATCHED_GENERIC = (NO_URLS_MATCHED_PREFIX +
+                           '. Do the files you\'re operating on exist?')
+NO_URLS_MATCHED_TARGET = NO_URLS_MATCHED_PREFIX + ': %s'
 
 if six.PY3:
   # StandardError was removed, so use the base exception type instead
--- a/gslib/tests/test_naming.py
+++ b/gslib/tests/test_naming.py
@@ -44,6 +44,7 @@ from gslib.cloud_api import ServiceException
 from gslib.exception import CommandException
 from gslib.exception import InvalidUrlError
 from gslib.exception import NO_URLS_MATCHED_GENERIC
+from gslib.exception import NO_URLS_MATCHED_TARGET
 from gslib.storage_url import StorageUrlFromString
 import gslib.tests.testcase as testcase
 from gslib.tests.util import ObjectToURI as suri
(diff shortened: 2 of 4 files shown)


gslib/tests/test_cp.py
```diff
@@ -2344,7 +2345,7 @@ def test_cp_minus_e(self):
         expected_status=1)
     self.assertIn('Copying file', stderr)
     self.assertIn('Skipping symbolic link', stderr)
-    self.assertIn('CommandException: No URLs matched: %s' % fpath2, stderr)
+    self.assertIn(NO_URLS_MATCHED_TARGET % fpath2, stderr)
```
dilipped, line 2348: Nit: IMO it's better to have literal strings in tests. It improves readability and makes the tests more robust. In case the value of constant changes accidentally, this test will let you know that something has changed.

gslib/tests/test_naming.py
```diff
@@ -1429,7 +1430,7 @@ def testRemoveObjsCommand(self):
       self.RunCommand('rm', [suri(dst_bucket_uri, 'non_existent')])
       self.fail('Did not get expected CommandException')
     except CommandException as e:
-      self.assertIn(NO_URLS_MATCHED_GENERIC, e.reason)
+      self.assertIn(NO_URLS_MATCHED_PREFIX, e.reason)
```
dilipped, line 1433: Why use NO_URLS_MATCHED_PREFIX here? I am guessing just NO_URLS_MATCHED_PREFIX msg will not be raised from anywhere?

gslib/tests/test_rm.py
```diff
@@ -121,7 +121,7 @@ def _RunRmCommandAndCheck():
         update_lines = True
         # Retry 404's and 409's due to eventual listing consistency, but don't
         # add the output to the set.
-        if (NO_URLS_MATCHED_GENERIC in stderr or
+        if (NO_URLS_MATCHED_PREFIX in stderr or
```
dilipped, line 124: ditto

gslib/tests/test_cp.py
NickGoog: I'm curious about the best choice, so I made a YAQs post:
http://yaqs/4676980226785280

I'll change it for now, though.

dilipped: Thanks, I would be curious to know what others think as well. YAQs post is a good idea. You might want to add a little more context there that the constant is coming from the actual code. 
I feel it is okay to use a constant that is separate for the test code, so that you don't have to repeat the literal everywhere in the test code. 
Another alternate idea would be to have a unittest to test the literal as well, then we can be sure that the literal is what we are expecting.

gslib/tests/test_rm.py
NickGoog: Made it more specific by using `NO_URLS_MATCHED_TARGET`

dilipped: Got it. In that case , this should be good.

gslib/tests/test_naming.py
NickGoog: Made the error we're testing for more specific.

gslib/exception.py
```diff
@@ -31,8 +31,10 @@
 
 import six
 
-NO_URLS_MATCHED_GENERIC = 'No URLs matched'
-NO_URLS_MATCHED_TARGET = 'No URLs matched: %s'
+NO_URLS_MATCHED_PREFIX = 'No URLs matched'
+NO_URLS_MATCHED_GENERIC = (NO_URLS_MATCHED_PREFIX +
+                           '. Do the files you\'re operating on exist?')
```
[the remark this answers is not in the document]
NickGoog: Accidentally merged before they could respond.
Will make another PR if the text doesn't work.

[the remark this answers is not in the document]
starsandskies: I think that should be fine. There's a devilishly nuanced discussion to be had around "file" vs. "object", but I think most users understand the terms to be, generally, synonymous. If you wanted, you could change the wording from "files" to "files/objects" so that people don't accidentally try to cue off of "files" to think the issue must be within their local directory?

PR #1045, opened 2020-06-15
Require force flag for expensive mv

NickGoog:
Requires "-f" for mv on nearline, coldline, or archive storage class object that will incur charge.
First part of:
http://b/30305521
Docs change in separate CL.


NickGoog:
Holding off on merging this b/c we found it may break existing user flows.

PR #1056, opened 2020-07-01
Fix tests using wrong AWS credentials.

NickGoog:
Boto falls back on other credentials files or environment variables when fields aren't present in a Boto config file.
For those with AWS CLI installed, this meant an `~/.aws/credentials` file was giving Boto credentials even when we tried to anonymize a test client.

Documentation:
http://boto.cloudhackers.com/en/latest/boto_config_tut.html

### DIFF
--- a/gslib/tests/testcase/integration_testcase.py
+++ b/gslib/tests/testcase/integration_testcase.py
@@ -1137,8 +1137,13 @@ class GsUtilIntegrationTestCase(base.GsUtilTestCase):
 
     with SetBotoConfigForTest(boto_config_for_test, use_existing_config=False):
       # Make sure to reset Developer Shell credential port so that the child
-      # gsutil process is really anonymous.
-      with SetEnvironmentForTest({'DEVSHELL_CLIENT_PORT': None}):
+      # gsutil process is really anonymous. Also, revent Boto from falling back
+      # on credentials from other files like ~/.aws/credentials or environment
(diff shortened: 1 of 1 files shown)


gslib/tests/testcase/integration_testcase.py
```diff
@@ -1135,11 +1136,23 @@ def SetAnonymousBotoCreds(self):
...
+    if AWS_SECRET_ACCESS_KEY in os.environ:
+      saved_aws_secrety_access_key = os.environ[AWS_SECRET_ACCESS_KEY]
+    os.environ[AWS_SECRET_ACCESS_KEY] = '_'
+
+    try:
+      with SetBotoConfigForTest(boto_config_for_test,
+                                use_existing_config=False):
+        # Make sure to reset Developer Shell credential port so that the child
+        # gsutil process is really anonymous.
+        with SetEnvironmentForTest({'DEVSHELL_CLIENT_PORT': None}):
```
dilipped, line 1151: Isn't this function setting the envs for you? Can we just pass the values to this function?

NickGoog: Wow, I didn't see that

dilipped:
You have a third_party module in there. Probably got committed accidentally?

NickGoog:
Was accidentally in a detached head state inside a submodule at the time of commit.

PR #1059, opened 2020-07-06
Fixes regex conflicts with CloudSDK detection

thomasmaclean:
I cases where tests are run via a version of Gsutil invoked via CloudSDK, the user agent will have a CloudSDK suffix appended, which will interfere with a `$` in regex comparisons of the resulting user agent string. I've now mocked out the CloudSDK detection completely in cases where the `$` is necessary (i.e. when checking whether we are looking at `command/cp` or `command/cp-DaisyChain`), and removed it where it was superfluous.


NickGoog:
Did you run yapf? Travis error seems related to that.

PR #1068, opened 2020-07-15
Add userProject support to signurl

thomasmaclean:
re-rolling #923 to be compatible with the new `-u` flag.


gslib/commands/signurl.py
```diff
@@ -445,10 +458,15 @@ def _ParseAndCheckSubOpts(self):
           'url arguments if the --use-service-account flag is missing. '
           'Run `gsutil help signurl` for more info')
 
-    return method, delta, content_type, passwd, region, use_service_account
+    if use_service_account and billing_project:
+      raise CommandException(
+          'Specifying both the -b and -u options together is invalid.')
```
starsandskies, line 463: I think this is ambiguous: within the signurl command, there's also an unrelated -u flag and I don't think that one would be incompatible with -b, right?

gslib/commands/signurl.py
```diff
@@ -162,6 +162,13 @@
                Note that both options have a maximum allowed duration of
                12 hours for a valid link.
 
+  -b <project> Allows you to specify a user project that will be billed for
```
starsandskies, line 165: I think generally the options are listed alphabetically for gsutil commands, though I see that's not the case currently for this command

gslib/commands/signurl.py
```diff
@@ -162,6 +162,13 @@
...
+               requests on the signed URL. This is useful for generating
```
starsandskies, line 166: Suggestion: Change "on the" to "that use"

gslib/commands/signurl.py
```diff
@@ -162,6 +162,13 @@
...
+               presigned links for buckets that use requester pays.
+
+               Note that it's not valid to specify both the ``-b`` and ``-u``
+               options together.
```
starsandskies, line 170: Why wouldn't this be valid? Wouldn't the -u flag specify the account to bill for executing the `gsutil signurl` command, and the -b flag specify the account to bill when someone uses the signed URL that `gsutil signurl` generates?

gslib/commands/signurl.py
thomasmaclean: That's exactly what I meant here... the `-u` option is actually using a totally different API where specifying a billing project would have no effect.

gslib/commands/signurl.py
thomasmaclean: I can scoot these around so that they're alphabetical.

gslib/commands/signurl.py
thomasmaclean: It's super annoying that there's a **local** `-u` and a **global** `-u` and I can VERY much understand how you could get those 2 confused. To your point, using both **global** `-u` and local `-b` _is_ a valid use case. I'm going to replace `-u` with `--use-service-account` to hopefully clear up the confusion.

gslib/commands/signurl.py
thomasmaclean: Done

gslib/commands/signurl.py
```diff
@@ -112,12 +112,15 @@
...
 
 <B>OPTIONS</B>
-  -m           Specifies the HTTP method to be authorized for use
-               with the signed url, default is GET. You may also specify
-               RESUMABLE to create a signed resumable upload start URL. When
-               using a signed URL to start a resumable upload session, you will
-               need to specify the 'x-goog-resumable:start' header in the
-               request or else signature validation will fail.
+  -b <project> Allows you to specify a user project that will be billed for
+               requests that use signed URL. This is useful for generating
```
starsandskies, line 116: Nit: "...that use the signed..."

gslib/commands/signurl.py
```diff
@@ -112,12 +112,15 @@
...
+               presigned links for buckets that use requester pays.
+
+               Note that it's not valid to specify both the ``-b`` and
+               ``--use-service-account`` options together.
```
starsandskies, line 120: I think this still confuses me: why are -b and --use-service-account mutually exclusive (perhaps I don't understand correctly what --use-service-account does)?

thomasmaclean: `--use-service-account` tells `signurl` to use a different API, so that rather than having to supply a keyfile, the command just uses the service account key to sign the URL. The phrasing of this could probably better represent this. I could also make it so that the option is just ignored rather than throwing an error. Something like "Note that the ``-b`` option will be ignored when generating keys with the ``--use-service-acount`` option." I could then print that message again when both options are used together, again rather than throwing an error.

gslib/utils/signurl_helper.py
```diff
@@ -79,6 +81,9 @@ def CreatePayload(client_id,
       'x-goog-expires': '%d' % duration.total_seconds()
   }
 
+  if (billing_project is not None):
```
dilipped, line 84: Nit: brackets are not required here.

gslib/commands/signurl.py
dilipped: I like the idea of specifying that -b will be ignored if used along with -u.

starsandskies: Wouldn't a problem with ignoring -b be that you could create signed URLs that seem valid, but then get rejected because the signed URL tries to access a Requester Pays bucket?

PR #1081, opened 2020-08-06
Replace socks library

dilipped:
b/161686363: Proxy authentication does not work on Python3 because of socksipy-branch lib. Replacing it with pysocks.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "third_party/crcmod_osx"]
 	path = third_party/crcmod_osx
 	url = https://github.com/gsutil-mirrors/crcmod-osx.git
-[submodule "third_party/socksipy-branch"]
-	path = third_party/socksipy-branch
-	url = git://github.com/gsutil-mirrors/socksipy-branch.git
 [submodule "third_party/retry-decorator"]
 	path = third_party/retry-decorator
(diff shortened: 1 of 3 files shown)


NickGoog:
Do we need to do any due diligence on this package before accepting it into the codebase?

dilipped:
@NickGoog I have seen the package being used internally for other stuff so I think we should be good. But that's a good point and being extra careful doesn't hurt. I will try to compare if the stuff in third_party has some modification or if it being used as it is.

dilipped:
Looks like kokoro tests are failing for windows. 
```
ImportError: To run PySocks on Windows you must install win_inet_pton
```
I will look into this.

dilipped:
I have a better way of addressing this problem and I think this change will no longer be needed. I will close this PR once I confirm the alternate solution works.

thomasmaclean:
What's the alternative solution?

dilipped:
@thomasmaclean  Remove socksipy, but do not replace it with PySocks. Instead rely on the `socks.py` module present in httplib2 https://github.com/httplib2/httplib2/blob/master/python3/httplib2/socks.py

dilipped:
Closing this PR in favor of https://github.com/GoogleCloudPlatform/gsutil/pull/1084 which provides a better alternative to this problem and does not require additional dependency for Windows.

Issue #1065, opened 2020-07-09
407 Proxy Authentication Required error when running gsutil on Cloud SDK 296.0.0 or above using a proxy


turribeach:
Running "gsutil" on a Windows PC configured to use a proxy returns the "407 Proxy Authentication Required" despite all proxy settings being correctly set. I traced the problem to an update on Cloud SDK v296.0.0:

[https://cloud.google.com/sdk/docs/release-notes#29600_2020-06-09](https://cloud.google.com/sdk/docs/release-notes#29600_2020-06-09)

It seems to be the "Updated bundled Python executable to 3.7.5." line. In version 295.0.0 the bundled Python is v2.7.13 so gsutil using a proxy seems to break on Python 3.7.5.

To verify this issue:

1. Run "gcloud --version" to verify your Google Cloud SDK is 295.0.0 or earlier
2. Then confirm "gsutil ls" runs fine using a proxy
3. Then run "gcloud components update" to update your Google Cloud SDK to the latest version (likely to be 300.0.0 or above)
4. Now run "gsutil ls" again and you should get a "407 Proxy Authentication" error despite proxy settings being correctly set

You can rollback to version 295.0.0 using this command: "gcloud components update --version 295.0.0"

You can also reproduce the problem setting the CLOUDSDK_GSUTIL_PYTHON variable to use a python3 interpreter like this:

`set CLOUDSDK_GSUTIL_PYTHON=c:\[path to your python3 installation\python`

As a workaround I was able to update the Google Cloud SDK to the current latest version (300.0.0) and set CLOUDSDK_GSUTIL_PYTHON to point to the embedded Python 2 interpreter which allows gsutil to work fine using a proxy:

`set CLOUDSDK_GSUTIL_PYTHON=C:\Dev\Programs\google-cloud-sdk\platform\bundledpython2\python`

However Google claims on the [Google Cloud SDK release notes](https://cloud.google.com/sdk/docs/release-notes) that support for Python 2 will be removed on July 31st 2020 so presumably this work around will stop working when they remove Python 2 from the Google Cloud SDK.

turribeach:
We are seeing more people in our organisation find this problem, as they update their Google Cloud SDK. A couple of points worth mentioning:

[...]

PR #1084, opened 2020-08-13
Remove socksipy dependency to fix proxy authentication

dilipped:
Proxy authentication is not working for python3.  b/161686363, https://github.com/GoogleCloudPlatform/gsutil/issues/1065 
The problem is because of socksipy-branch.

This PR remove the socksipy dependency and instead relies on the socks.py lib present in httplib2

Dependent on: https://github.com/GoogleCloudPlatform/gcs-oauth2-boto-plugin/pull/52
Requires httplib2>=0.18

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "third_party/crcmod_osx"]
 	path = third_party/crcmod_osx
 	url = https://github.com/gsutil-mirrors/crcmod-osx.git
-[submodule "third_party/socksipy-branch"]
-	path = third_party/socksipy-branch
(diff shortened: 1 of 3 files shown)


dilipped:
Yeah the change is dependent on a change in gcs-oauth2-boto-plugin>

NickGoog:
Will Travis be fixed after the change is made?

dilipped:
Yes, it is complaining because gcs-oauth2-boto-plugin==2.7 is not yet available on PyPI. It will be made available once GoogleCloudPlatform/gcs-oauth2-boto-plugin#52 gets submitted.

PR #1087, opened 2020-08-19
Skip leftover dir placeholder test for S3

dilipped:
With V4 signature, objects starting with a slash do not work. So skipping this test for S3.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -2172,8 +2172,8 @@ class TestRsync(testcase.GsUtilIntegrationTestCase):
     _Check1()
 
   @unittest.skipUnless(UsingCrcmodExtension(), 'Test requires fast crcmod.')
-  @SkipForS3('The boto lib used for S3 does not handle objects '
-             'starting with slashes if we use V4 signature')
+  @SkipForS3('The boto lib used for S3 does not handle objects'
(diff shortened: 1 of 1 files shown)


gslib/tests/test_rsync.py
```diff
@@ -2172,6 +2172,8 @@ def _Check1():
     _Check1()
 
   @unittest.skipUnless(UsingCrcmodExtension(), 'Test requires fast crcmod.')
+  @SkipForS3('The boto lib used for S3 does not handle objects '
+             'starting with slashes if we use V4 signature')
```
NickGoog, line 2176: This is a super nit, but don't spaces between words normally go after line breaks?
'objects'
' starting'

Also, please add a period.

dilipped: I think the space is required because Python will just concat the two strings
```
>>> s = ('some string'
...      'value')
>>> s
'some stringvalue'
```

dilipped: ah, nevermind, I think  you meant move the space to the next line

dilipped: Done

dilipped:
I marked this PR as WIP since I have already started with the release process and do not want to include this PR in the current release. This does not affect anything from usability purpose, so it should be okay to skip this PR in the current WIP release.

PR #1091, opened 2020-08-20
Headers can now be removed

thomasmaclean:
Previously the JSON API would not delete the headers `cache-control`, `content-disposition`, `content-encoding`, and `content-language`.


NickGoog:
All failures due to timeouts

NickGoog (approved):

PR #1097, opened 2020-09-02
Fix broken signurl error message.

NickGoog:
Fixes http://b/167584551

### DIFF
--- a/gslib/boto_translation.py
+++ b/gslib/boto_translation.py
@@ -219,6 +219,10 @@ class BotoTranslation(CloudApi):
     self.api_version = boto.config.get_value('GSUtil', 'default_api_version',
                                              '1')
 
+  def GetServiceAccountId(self):
+    """Service account credentials unused for S3."""
+    return None
+
   def GetBucket(self, bucket_name, provider=None, fields=None):
     """See CloudApi class for function doc strings."""
     _ = provider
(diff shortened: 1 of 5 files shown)


NickGoog:
`client_email` is used later in the function, so please check the code branches to make sure I'm not introducing any unwanted behavior in URL generation (I actually hope this will fix URL generation for a separate issue).

gslib/gcs_json_credentials.py
```diff
@@ -133,6 +133,12 @@ def GetCredentialStoreKey(credentials, api_version):
   return '-'.join(key_parts)
 
 
+def GetImpersonateServiceAccount():
```
dilipped, line 136: This does not cover the service account scenario.

We should be instead using https://github.com/GoogleCloudPlatform/gsutil/blob/3b61bf0e6188f65f78c72c79ea3cb69e9c61da4b/gslib/gcs_json_api.py#L314

But that method is inside gcs_json_api, which means we will have to add an API method that exposes this method.

dilipped (changes requested):

gslib/gcs_json_credentials.py
NickGoog: Fixed!

dilipped (approved):
Make sure to run the S3 tests to ensure that nothing is breaking there.

dilipped:
Looks like Travis is failing

NickGoog:
S3 passed. Travis is disappointed I forgot to yapf haha

PR #1102, opened 2020-09-10
Warn when disabling parallel composite uploads for KMS encryption.

mhandugan:
Users can be surprised by reduced upload performance when using KMS encryption.
Currently, this is incompatible with composite uploads, and gsutil disables parallel uploads accordingly.
This PR changes gsutil to emit a warning when bypassing parallel composite uploads due to KMS encryption.


gslib/utils/copy_helper.py
```diff
@@ -1343,6 +1338,14 @@ def _ShouldDoParallelCompositeUpload(logger,
...
     return False
 
+  # TODO(KMS, Compose): Until we ensure service-side that we have efficient
+  # compose functionality over objects with distinct KMS encryption keys (CMEKs)
+  # or distinct CSEKs, don't utilize parallel composite uploads.
+  if kms_keyname:
+      logger.warning('Not using parallel composite upload for KMS-encryption. ' +
+                     'This combination is not currently supported by GCS.')
+      return False
+
```
thomasmaclean, line 1348: Why wouldn't you just keep line 1300 in tact and add the warning up there?

I might also make line 1349-1389 a function, which you can call at 1300, so that way you don't have to duplicate this message. Something like:
```
if kms_keyname or _DoBucketMetdataPCUCheck():
  logger.warning('Parallel composite uploads have been disabled. Composite uploads are not ' +
                 'supported in buckets that have a default customer-managed encryption key.')
  return False
```

thomasmaclean (changes requested):

gslib/utils/copy_helper.py
```diff
@@ -185,6 +185,12 @@
 # would be unnecessary.
 open_files_lock = parallelism_framework_util.CreateLock()
 
+# Declarations for tracking state of one-time warning that use of KMS is
+# disabling parallel composite uploads.
+global kms_compose_warning_counter, kms_compose_warning_lock
+kms_compose_warning_lock = parallelism_framework_util.CreateLock()
+kms_compose_warning = False
```
dilipped, line 192: Did you mean to initialize this to True?

dilipped (changes requested):

gslib/utils/copy_helper.py
```diff
@@ -1272,6 +1278,14 @@ def _DoParallelCompositeUpload(fp,
   return elapsed_time, composed_object
 
 
+def _WarnOnceAboutKmsCompose(logger):
```
dilipped, line 1281: Here and everywhere else, please follow two space indentation.

gslib/utils/copy_helper.py
```diff
@@ -1272,6 +1278,14 @@ def _DoParallelCompositeUpload(fp,
...
+    with kms_compose_warning_lock:
+        if kms_compose_warning:
+            kms_compose_warning = False
+            logger.warning('Not using parallel composite upload for KMS-encryption. ' +
```
dilipped, line 1285: Nit: + sign is not required here. The string will be concatenated without it as well.

gslib/utils/copy_helper.py
dilipped: I realized that I marked this as Resolved, but my comment was not posted.'

I talked to @mhandugan offline regarding this. The reason for doing this was to not log a warning if composite upload is not required (e.g when the files are small enough).

gslib/utils/copy_helper.py
mhandugan: Updated

dilipped (approved):

PR #1103, opened 2020-09-11
Doc improvements for gsutil du command

karenarialin:


gslib/commands/du.py
```diff
@@ -48,68 +48,66 @@
...
     gsutil du -ch gs://bucketname
 
-  To see a summary of the total bytes in the two given buckets:
+  To see a summary of the total number of bytes in two given buckets:
 
     gsutil du -s gs://bucket1 gs://bucket2
 
-  To list the size of all objects in a versioned bucket, including objects that
-  are not the latest:
+  To list the size of each object in a bucket with Object Versioning
```
starsandskies, line 100: The existing wording is not great, but it may be deliberate: this flag not only applies to buckets that have Object Versioning *currently* enabled, but I believe it also applies to buckets that had Object Versioning enabled *at some point in the past*.

starsandskies (approved):
Nice! LGTM, with one comment about Object Versioning wording.

PR #1105, opened 2020-09-17
Backfill several updates to reference docs

starsandskies:
These changes have all recently been made to the cloud.google.com documentation and should be backfilled into the Github source files.

### DIFF
--- a/gslib/commands/rm.py
+++ b/gslib/commands/rm.py
@@ -56,6 +56,10 @@ _DETAILED_HELP_TEXT = ("""
 
 
 <B>DESCRIPTION</B>
+  NOTE: As part of verifying the existence of objects prior to deletion,
+  ``gsutil rm`` makes ``GET`` requests to Cloud Storage for object metadata.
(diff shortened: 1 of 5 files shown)


gslib/commands/rm.py
```diff
@@ -56,6 +56,10 @@
 
 
 <B>DESCRIPTION</B>
+  NOTE: As part of verifying the existence of objects prior to deletion,
+  ``gsutil rm`` makes ``GET`` requests to Cloud Storage for object metadata.
```
dilipped, line 60: This is true for cp, rsync, mv as well.

dilipped (approved):

gslib/commands/rm.py
starsandskies: Is that to say that these other methods make two GET requests? One GET to verify an object's existence and a second GET to retrieve the object data?

If so, I'll think about if/how to add that info to those commands as well. Merging this PR in the meantime.

PR #1114, opened 2020-10-07
Backfill cl/335882034

starsandskies:
Improvements and centralization of the discussion of parallel composite uploads.

### DIFF
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -53,7 +53,6 @@ from gslib.utils.copy_helper import GetSourceFieldsNeededForCopy
 from gslib.utils.copy_helper import GZIP_ALL_FILES
 from gslib.utils.copy_helper import ItemExistsError
 from gslib.utils.copy_helper import Manifest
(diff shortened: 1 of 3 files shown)


gslib/commands/cp.py
```diff
@@ -338,9 +338,9 @@
 
   Until the upload has completed successfully, it is not visible at the destination
   object and does not supersede any existing object the upload is intended to
-  replace. However, parallel composite uploads may leave temporary component
-  objects in place during the upload process. See Parallel Composite Uploads for more
-  information.
+  replace. However, `parallel composite uploads
+  <https://cloud.google.com/storage/docs/uploads-downloads#parallel-composite-uploads>`_
```
NickGoog, line 342: Is the "_" intentional?

starsandskies: Yes - though I have it here because I'm copying the style used in existing links, such as the one on line 414 below. As to the purpose (or need) for it, I can only speculate it's somehow required as part of the github --> cloud.google.com generation process.

PR #1122, opened 2020-10-14
mTLS Part 2/3: Connect to Endpoint

NickGoog:
Configure gsutil to optionally connect to the mTLS endpoint, using a device certificate.

Sets up an E2E test to automatically confirm this authentication type continues working.

Builds on: https://github.com/GoogleCloudPlatform/gsutil/pull/1115
Next: https://github.com/GoogleCloudPlatform/gsutil/pull/1142

### DIFF
--- a/gslib/cloud_api_delegator.py
+++ b/gslib/cloud_api_delegator.py
@@ -21,6 +21,7 @@ from __future__ import unicode_literals
 
 import boto
 from boto import config
+from gslib import context_config
 from gslib.cloud_api import ArgumentException
 from gslib.cloud_api import CloudApi
 from gslib.cs_api_map import ApiMapConstants
@@ -209,6 +210,15 @@ class CloudApiDelegator(CloudApi):
     elif self.prefer_api in (
         self.api_map[ApiMapConstants.SUPPORT_MAP][selected_provider]):
       api = self.prefer_api
+
+    if (api == ApiSelector.XML and context_config.get_context_config() and
+        context_config.get_context_config().use_client_certificate):
+      raise ArgumentException(
+          'User enabled mTLS by setting "use_client_certificate", but mTLS'
+          ' is not supported for the selected XML API. Try configuring for '
+          ' the GCS JSON API or setting "use_client_certificate" to "False" in'
+          ' the Boto config.')
+
     return api
 
   def GetServiceAccountId(self, provider=None):
--- a/gslib/gcs_json_api.py
+++ b/gslib/gcs_json_api.py
@@ -41,6 +41,7 @@ from boto import config
 import httplib2
 import oauth2client
 
+from gslib import context_config
 from gslib.cloud_api import AccessDeniedException
 from gslib.cloud_api import ArgumentException
 from gslib.cloud_api import BadRequestException
@@ -169,6 +170,9 @@ _SKIP_LISTING_OBJECT = 'skip'
 _INSUFFICIENT_OAUTH2_SCOPE_MESSAGE = (
     'Insufficient OAuth2 scope to perform this operation.')
 
+DEFAULT_HOST = 'storage.googleapis.com'
+MTLS_HOST = 'storage.mtls.googleapis.com'
+
 
 class GcsJsonApi(CloudApi):
   """Google Cloud Storage JSON implementation of gsutil Cloud API."""
@@ -228,7 +232,18 @@ class GcsJsonApi(CloudApi):
 
     self.http_base = 'https://'
     gs_json_host = config.get('Credentials', 'gs_json_host', None)
-    self.host_base = gs_json_host or 'storage.googleapis.com'
+    if (context_config.get_context_config() and
+        context_config.get_context_config().use_client_certificate):
+      if gs_json_host:
+        raise ArgumentException(
+            '"use_client_certificate" is enabled, which sets gsutil to use the'
+            ' host {}. However, a custom host was set using'
+            ' "gs_json_host": {}. Please set "use_client_certificate" to'
+            ' "False" or comment out the "gs_json_host" line in the Boto'
+            ' config.'.format(MTLS_HOST, gs_json_host))
+      self.host_base = MTLS_HOST
+    else:
+      self.host_base = gs_json_host or DEFAULT_HOST
 
     if not gs_json_host:
       gs_host = config.get('Credentials', 'gs_host', None)
--- a/gslib/tests/test_boto_util.py
+++ b/gslib/tests/test_boto_util.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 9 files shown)


gslib/gcs_json_api.py
```diff
@@ -228,7 +231,15 @@ def __init__(self,
 
     self.http_base = 'https://'
     gs_json_host = config.get('Credentials', 'gs_json_host', None)
-    self.host_base = gs_json_host or 'storage.googleapis.com'
+    if context_config.get_context_config().use_client_certificate:
+      if gs_json_host:
+        raise ArgumentException(
+            '"use_client_certificate" is enabled, which sets gsutil to use the'
+            ' endpoint {}. However, a custom endpoint was also set: {}'.format(
```
rrauber, line 238: Maybe add a sentence suggesting that the user disable use_client_certificate or un-set a custom endpoint to fix the error.

rrauber (dismissed):

gslib/gcs_json_api.py
NickGoog: Done!

gslib/cloud_api_delegator.py
```diff
@@ -209,6 +210,13 @@ def GetApiSelector(self, provider=None):
     elif self.prefer_api in (
         self.api_map[ApiMapConstants.SUPPORT_MAP][selected_provider]):
       api = self.prefer_api
+
+    if (api == ApiSelector.XML and context_config.get_context_config() and
+        context_config.get_context_config().use_client_certificate):
+      self.logger.warn(
```
dilipped, line 216: I think we should raise a warning here? Since mTLS is a security feature, if a user is opting in for this feature and it is not going to work, they should be alerted to either make the proper API choice or set use_client_certificate to False.

gslib/gcs_json_api.py
```diff
@@ -228,7 +231,18 @@ def __init__(self,
 
     self.http_base = 'https://'
     gs_json_host = config.get('Credentials', 'gs_json_host', None)
-    self.host_base = gs_json_host or 'storage.googleapis.com'
+    if context_config.get_context_config(
```
dilipped, line 234: Nit: You can wrap the if condition within parenthesis to get rid of the weird formatting here.

rrauber (approved):

gslib/gcs_json_api.py
dilipped: Also, given that context_config.create_context_config is always called at the beginning, do we need to check for context_config.get_context_config? Should that object always exist?

gslib/gcs_json_api.py
```diff
@@ -228,7 +231,18 @@ def __init__(self,
...
+    ) and context_config.get_context_config().use_client_certificate:
```
dilipped, line 235: Are you missing test for this section? I think we can simulate the context config and add a test to make sure this error gets raised?

gslib/utils/boto_util.py
```diff
@@ -294,6 +295,14 @@ def GetNewHttp(http_class=httplib2.Http, **kwargs):
   http = http_class(proxy_info=proxy_info, **kwargs)
   http.disable_ssl_certificate_validation = (not config.getbool(
       'Boto', 'https_validate_certificates'))
+
+  global_context_config = context_config.get_context_config()
+  if global_context_config and global_context_config.use_client_certificate:
+    http.add_certificate(key=global_context_config.client_cert_path,
```
dilipped, line 301: Can we add a unit test to check that add certificate gets called if context config is present?

gslib/gcs_json_api.py
NickGoog: The first check is for tests that may not run gsutil's main setup.

gslib/cloud_api_delegator.py
NickGoog: Done and created test file

gslib/gcs_json_api.py
NickGoog: Wasn't sure if I should cross the threshold of creating a new test file for this untested class.
Done.

gslib/utils/boto_util.py
NickGoog: Added a new test file, same reasoning as above

gslib/cloud_api_delegator.py
```diff
@@ -209,6 +210,13 @@ def GetApiSelector(self, provider=None):
     elif self.prefer_api in (
         self.api_map[ApiMapConstants.SUPPORT_MAP][selected_provider]):
       api = self.prefer_api
+
+    if (api == ApiSelector.XML and context_config.get_context_config() and
+        context_config.get_context_config().use_client_certificate):
+      raise ArgumentException(
+          'User enabled mTLS by setting "use_client_certificate", but mTLS'
```
dilipped, line 217: Nit: You can add an actionable step here like you do for the error raised from GcsJsonApi

gslib/tests/test_boto_util.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for boto_util.py."""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import mock
```
dilipped, line 22: Do you need to use six.moves here? To make it compatible with both Python2 and Python3

gslib/tests/test_boto_util.py
```diff
@@ -0,0 +1,51 @@
...
+                                                       mock_get_context_config):
+    mock_context_config = mock.Mock()
+    mock_context_config.use_client_certificate = True
+    mock_context_config.client_cert_path = 'path'
+    mock_context_config.client_cert_password = 'password'
+    mock_get_context_config.return_value = mock_context_config
+
+    mock_http_class = mock.Mock(return_value=mock.Mock())
+    mock_http = boto_util.GetNewHttp(mock_http_class)
+    mock_http.add_certificate.assert_called_once_with(
```
dilipped, line 46: Maybe add an assertion that mock_get_context_config was called, since we do not have any other test that covers that.

gslib/tests/test_gcs_json_api.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for gcs_json_api.py."""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import mock
```
dilipped, line 22: Ditto

gslib/tests/test_boto_util.py
NickGoog: It would have to be called to get the "path" and "password" value asserted below

dilipped: Ah I see. that covers it. Looks good.

gslib/cloud_api_delegator.py
NickGoog: Done

gslib/tests/test_boto_util.py
NickGoog: I default to not using it until an issue pops up

gslib/tests/test_gcs_json_api.py
NickGoog: Ditto indeed

gslib/tests/test_boto_util.py
dilipped: An issue might not show up on your machine because mock might be already installed for your python version. But that might not be the case of the user.

dilipped (approved):

PR #1135, opened 2020-11-06
no need of sudo to pip install crcmod on macOS

Nakilon:
It works without `sudo`. Did it 10 minutes ago.


dilipped:
Thanks for the PR. You might have the correct permissions or you must be using the python installed for your user and hence sudo was not required. Still majority of cases are those that required sudo so I think the documentation in its current state makes sense for majority of cases and I don't think it's worth changing the instructions just for macOS.

Nakilon:
That documentation line is already in the macOS specific chapter.
Also macOS is not like a docker container or some messy temporary linux installation. It is a bad practice to run things with sudo on macOS.

dilipped:
I see. Thanks for the explanation.

Nakilon:
Thank you!

PR #1146, opened 2020-11-18
improve upload speed significantly when it runs on Windows

### LINKED ISSUE #1145: Uploading performance is very poor for Windows in high round trip latency environment.
When uploading a 10MB+ file from Windows OS using `gsutil`, from high round trip latency environment, the performance of upload is very poor. From the same network environment, sending a 13MB file from Linux spent about 1 or 2s, sending the same file with `curl` from Windows spent similar or less, but sending the same file with `gsutil` on the same Windows spent about 7 seconds, about 1MB/s speed.

```
C:\Users\user>gsutil cp obj gs://bucketfortest/obj
Copying file://obj [Content-Type=application/octet-stream]...
- [1 files][ 13.0 MiB/ 13.0 MiB]  995.3 KiB/s
Operation completed over 1 objects/13.0 MiB.
```

This is full sequence of file sending from Linux. As you can see, it starts slowly, then scaled the TCP window large enough, and send whole file at a few cycles.

gsutil-perf-002-gsutil--rtt50--linux

This is full sequence of file sending from Windows. In this sequence, it also did TCP window scaling but the final size of window is just 8K * 8 packet = 64KB and it needs bunch of PUSH/ACK cycles to finish the sending.

gsutil-perf-001-gsutil--rtt60

Close up!
gsutil-perf-003-gsutil--rtt60---early-stage

It seems like TCP window scaling is not working properly as expected, in some condition. Actually 60ms of RTT is not that huge latency but this behavior made it into a huge problem. While I am searching some hints and solution, I found that recent version of `curl` works fine as I already mentioned, but the old version, `curl` version 7.55 showed the same symptoms as `gsutil`.  After some more search on the code and commits of `curl`, I found this bug report and it looks like I found the same problem which basically caused by unexpected behavior of Windows's TCP subsystem.

https://curl.se/mail/lib-2018-07/0080.html

I made a quick patch for it and will make a PR soon. The patch will improve the performance by enlarge the application's send buffer size.

Condition of the issue:
* Windows OS (tested on Windows Server 2016/2019 Datacenter, and Windows Server version 2004 Datacenter Core)
* Slightly high end to end round trip latency to GCS end point. (tested with about 50~60ms)
* gsutil version 4.54 (Google Cloud SDK 317.0.0, core 2020.10.30)

sio4:
This fixes #1145 by setting application's send buffer large enough. (can be heard as weird but it works)


dilipped (commented):
What was the reason for changing this file?

EDIT: I later realized that I did not select the file. I am talking about the test_data/test.txt file. I am curious what was the problem with the original file. Is it the length of the file?

sio4:
Thank you for your reviewing.

When I run unit test with increased buffer size, I got 6 errors and one of them is like below (the others are the same pattern):

```
======================================================================
FAIL: testSeekBackAroundOneBuffer (gslib.tests.test_hashing_helper.TestHashingFileUploadWrapper)
  File "T:\src\github\src\gsutil\gslib\tests\test_hashing_helper.py", line 124, in testSeekBackAroundOneBuffer
    self._testSeekBack(initial_position, seek_back_amount)
  File "T:\src\github\src\gsutil\gslib\tests\test_hashing_helper.py", line 86, in _testSeekBack
    '(but was actually: %s)' % (tmp_file_len, initial_position))
AssertionError: initial_position must be less than test file size 33765 (but was actually: 65537)
```

The following code is the caller and it uses manipulated number with `TRANSPER_BUFFER_SIZE` as a `initial_position`,

https://github.com/GoogleCloudPlatform/gsutil/blob/529dda682f418bf2910317493568551e2cb15a88/gslib/tests/test_hashing_helper.py#L114-L124

and this is the callee and it just made a assert error because the `test.txt` is smaller then `TRANSFER_BUFFER_SIZE`, or `TRANSFER_BUFFER_SIZE * 3 + 1`.

https://github.com/GoogleCloudPlatform/gsutil/blob/529dda682f418bf2910317493568551e2cb15a88/gslib/tests/test_hashing_helper.py#L76-L86

I am not super clear about the background context of these functions, but according to the code, the file `test.txt` should be large enough (larger than `TRANSFER_BUFFER_SIZE * 3 + 1`).

I also found some different way to provide a test file like... 

https://github.com/GoogleCloudPlatform/gsutil/blob/529dda682f418bf2910317493568551e2cb15a88/gslib/tests/test_hashing_helper.py#L51

and I think this dynamic way is better than static `test.txt` approach. However, this is not in scope of this patch so I just size up the test file.

Please kindly let me know if you have another thought.

dilipped:
Thanks for the detailed response @sio4. The solution that you suggested will create a file with all characters being the same. Which means we would not be able to catch the hash differences when seek was not made at the correct position. To keep all the bytes random,  I would recommend doing something like 
```
tmp_file = self.CreateTempFile(contents=os.urandom(TRANSFER_BUFFER_SIZE * 4)) 
```

Your current solution looks good, except that the data is repeated. It would be better if the data is truly random in the file.
Maybe, you can do something like 
```
base64 /dev/urandom | head -c 10000 > test.txt
```

sio4:
Hi @dilipped,
I didn't realize the file should be random. I rebuild the file with your suggestion.
However, two travis-ci test for pypy failed while the other tests were done successfully. Could you please check it?

dilipped:
Thanks @sio4. The travis failure is fine as the Gsutil CI passed for pypy. There is release in progress currently, so I will hold on to merging this PR. Will approve and merge it once the release process gets completed to avoid confusion.

PR #1154, opened 2020-11-24
Delete the projects.py help topic

starsandskies:
The information in this topic is pretty much all outdated and redundant. Removing the file will reduce sending readers in the wrong direction. Parallel change has been made to the c.g.c in cl/344094124.


gslib/commands/ls.py
```diff
@@ -58,13 +58,11 @@
 
 <B>LISTING PROVIDERS, BUCKETS, SUBDIRECTORIES, AND OBJECTS</B>
   If you run gsutil ls without URLs, it lists all of the Google Cloud Storage
-  buckets under your default project ID:
+  buckets under your default project ID (or all of the {{gcs_name}} buckets
```
NickGoog, line 61: Do we actually do text substitution in these? {{}}
I don't see other instances of this syntax or a "gcs_name" variable in the file

gslib/commands/mb.py
```diff
@@ -147,8 +146,8 @@
                          for a discussion of this distinction. Default is US.
                          Locations are case insensitive.
 
-  -p proj_id             Specifies the project ID under which to create the
-                         bucket.
+  -p proj_id             Specifies the project ID or project number under which
```
NickGoog, line 149: "to create the bucket under"?

gslib/commands/ls.py
starsandskies: No we don't! Good catch on this - I had copy-pasted from the c.g.c do, where these variables are relevant. Fixed.

gslib/commands/mb.py
starsandskies: It's half a dozen of one, six of the other. Does the current wording read awkwardly for you?

NickGoog: Yeah, "under which" sounds vaguely pompous. I think the last time I heard it used was in, "After completing the assignment, please place it in the bin, under which you can find your homework."

NickGoog (approved):

gslib/commands/mb.py
starsandskies: Works for me - switched

PR #1156, opened 2020-12-01
Fix formatting for empty CORS JSON document

dannyqiu:
This ensures that a code block is generated in https://cloud.google.com/storage/docs/gsutil/commands/cors?hl=en


starsandskies:
This seems correct to me, but would like one of the gsutil experts to confirm.

NickGoog (approved):

PR #1153, opened 2020-11-20
Add GitHub Actions CI

NickGoog:

PR #1164, opened 2020-12-21
Ignore the .github directory for updates

nigelbabu:
Currently all auto-updates are broken due to this folder.


nigelbabu:
Error on Windows

```
C:\python_27\files\python.exe C:\gsutil\gsutil update -n                                                 
CommandException: A file (C:\gsutil\.github) that is not distributed with gsutil was                                            
found in the gsutil directory. The update command cannot run with user                                                          
data in the gsutil directory.
```

nigelbabu:
This has been broken since #1153

PR #1169, opened 2020-12-23
Make signurl use generation information.

rrauber:
Before this change, generations could be specified in signurl commands but were not used in the url generation logic.


gslib/cloud_api.py
```diff
@@ -120,15 +120,16 @@ def SetBucketIamPolicy(self, bucket_name, policy, provider=None):
...
+  def SignUrl(self, method, duration, path, generation, logger, region,
+              signed_headers, string_to_sign_debug):
     """Sign a url using service account's system managed private key.
 
     Args:
       method: The HTTP method to be used with the signed URL.
       duration: timedelta for which the constructed signed URL should be valid.
       path: String path to the bucket or object for signing, in the form
           'bucket' or 'bucket/object'.
+      generation: If not None, specifies a version of an object for signing.
```
NickGoog, line 132: Maybe this should be the last argument with a default of None?

gslib/tests/test_signurl.py
```diff
@@ -125,6 +125,16 @@ def testSignUrlOutputJSON(self):
     self.assertIn('%2Fus-central1%2F', stdout)
     self.assertIn('\tPUT\t', stdout)
 
+  def testSignUrlWithJSONKeyFileAndObjectGeneration(self):
+    """Tests signurl output of a sample object version with json keystore."""
```
NickGoog, line 129: "JSON"

gslib/utils/signurl_helper.py
```diff
@@ -84,6 +86,9 @@ def CreatePayload(client_id,
   if (billing_project is not None):
     signed_query_params['userProject'] = billing_project
 
+  if generation is not None:
+    signed_query_params['generation'] = generation
```
NickGoog, line 90: So this is the core logic for us, and the API does the rest?

gslib/tests/test_signurl.py
```diff
@@ -125,6 +125,16 @@ def testSignUrlOutputJSON(self):
...
+    bucket_uri = self.CreateBucket(versioning_enabled=True)
+    object_uri = self.CreateObject(bucket_uri=bucket_uri, contents=b'z')
+    cmd = ['signurl', self._GetJSONKsFile(), object_uri.version_specific_uri]
+    stdout = self.RunGsUtil(cmd, return_stdout=True)
+    self.assertIn('x-goog-credential=test%40developer.gserviceaccount.com',
```
NickGoog, line 134: Is there a constant for this somewhere?
I see other tests are typing out the string but that seems wrong

gslib/utils/signurl_helper.py
rrauber: Yep! That's exactly right.

gslib/tests/test_signurl.py
rrauber: There wasn't one -- it comes from a file somewhere. I added one though.

gslib/cloud_api.py
rrauber: I kind of like grouping it with path here, and giving it a default here doesn't change the code much since all the `generation=None` lines in the tests were in calls to a private helper function. I gave generation a default there, but left everything else as-is.

NickGoog (approved):

Issue #573, opened 2018-10-17
rsync: support --ignore-existing


hundeboll:
Image the the following workflow:

- A cloudbuild step downloads a bunch of (but not all) files from a public bucket using HTTP
- The same step creates som additional files and places them among the downloaded files
- The next cloudbuild step runs `gsutil rsync -r -d ./download-dir gs://bucket-name`

Since the files are downloaded using HTTP, the mtime doen't match, and so all files are re-uploaded to the bucket.

I know I could force checksum matching, but since the files have checksum digests included in their filenames, it would be enough to simply ignore existing files.

I have also seen suggestions to use `rsync cp` to avoid replacing existing files, but that doesn't support deleting destination files not present in the source tree.

The non-gsutil version of rsync has --ignore-existing option to support my request. Would it be possible to implement the same in `gsutil` ?

zardoz-bzar:
workaround for a single directory

I pulled my bucket, then touched a file called foo.

eval BUCKET="bzar-inf-1" grep -vf <( gsutil ls gs://$BUCKET/ | sed -e s,gs://$BUCKET/,,  ) <( ls -1 ) | xargs -l -i{} gsutil cp {} gs://$BUCKET/

Copying file://foo [Content-Type=application/octet-stream]...
/ [1 files][    0.0 B/    0.0 B]
Operation completed over 1 objects.

Just replace the bucket value, subdirectories? just a more fun problem to solve

catleeball:
Note to internal folks, I mirrored this FR to b/135626681

sergeylanzman:
added support https://github.com/GoogleCloudPlatform/gsutil/pull/1183

dilipped:
Thanks for the PR. Resolving this.

PR #1183, opened 2021-01-19
rsync add ignore-existing option

sergeylanzman:
Added `--ignore-existing ` as in non-gsutil version of rsync.

Some times need rsync with GCS folder with many huge, immutable files. On first time dest don't have mtime and gsutil starting calculate checksum....
With  immutable files don't need check mtime and checksum.

Use case for this cassandra backups.

https://github.com/GoogleCloudPlatform/gsutil/issues/573

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -414,6 +414,11 @@ _DETAILED_HELP_TEXT = ("""
                  ignored. Note that gsutil does not follow directory symlinks,
                  regardless of whether -e is specified.
 
+  -i             This forces rsync to skip any files which exist on the destination
+                 and have a modified time that is newer than the source file.
+                 (If an existing destination file has a modification time equal to
+                 the source file's, it will be updated if the sizes are different.)
+
   -j <ext,...>   Applies gzip transport encoding to any file upload whose
                  extension matches the -j extension list. This is useful when
                  uploading files with compressible content (such as .js, .css,
@@ -945,6 +950,7 @@ class _DiffIterator(object):
     self.base_dst_url = base_dst_url
     self.preserve_posix = command_obj.preserve_posix_attrs
     self.skip_old_files = command_obj.skip_old_files
+    self.ignore_existing = command_obj.ignore_existing
 
     self.logger.info('Building synchronization state...')
 
@@ -1141,6 +1147,8 @@ class _DiffIterator(object):
     use_hashes = (self.compute_file_checksums or
                   (StorageUrlFromString(src_url_str).IsCloudUrl() and
                    StorageUrlFromString(dst_url_str).IsCloudUrl()))
+    if self.ignore_existing:
+      return False, has_src_mtime, has_dst_mtime
     if (self.skip_old_files and has_src_mtime and has_dst_mtime and
         src_mtime < dst_mtime):
       return False, has_src_mtime, has_dst_mtime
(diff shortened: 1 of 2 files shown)


gslib/commands/rsync.py
```diff
@@ -448,6 +448,11 @@
                  what would be copied or deleted without actually doing any
                  copying/deleting.
 
+  -i             This forces rsync to skip any files which exist on the destination
```
rrauber, line 451: I think this section is alphabetized. Could you move this between -e and -j?

rrauber (changes requested):
Thanks for the PR! The code itself looks good, but could you add a test for the new flag? It can be similar to (but probably a lot simpler than) this one: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/tests/test_rsync.py#L2822

Basically just set up a file and a corresponding object with different contents, run `gsutil rsync -i`, then use `gsutil cat` to check if the object has its contents and not the file's contents.

gslib/commands/rsync.py
sergeylanzman: Done

sergeylanzman:
@rrauber All done

rrauber (approved):

gslib/commands/rsync.py
```diff
@@ -414,6 +414,11 @@
                  ignored. Note that gsutil does not follow directory symlinks,
                  regardless of whether -e is specified.
 
+  -i             This forces rsync to skip any files which exist on the destination
+                 and have a modified time that is newer than the source file.
+                 (If an existing destination file has a modification time equal to
+                 the source file's, it will be updated if the sizes are different.)
+
```
anantakrishna, line 421: As far as we can see in the implementation, modification time is not taken into consideration at all. But the description here states that rsync skips files only if the destination file “have a modified time that is newer than the source file,” which is misleading then.

If the author or any other project maintainer can confirm my concern, let us fix this.

P.S.
I've come here looking for an answer to the following question: [What is the difference between -i and -u options in gsutil rsync command?](https://serverfault.com/questions/1118444/what-is-the-difference-between-i-and-u-options-in-gsutil-rsync-command). And I've found out that we have this PR and earlier #583, which introduces the `-u` option, which indeed makes rsync skip the files only “if the destination mtime is newer.”

NickGoog: At first glance, looks like you're right--there's no `>` or `<` symbols for the times in this PR.

However, mtime-related values are still returned by the PR code, and the parent function they're returned to is complex. We're redesigning `rsync` for the new `gcloud storage` CLI soon, and we'll probably wait until the accompanying in-depth study of gsutil `rsync` to confirm the behavior.

NickGoog: Oh, and big thank you for catching this!

anantakrishna: > We're redesigning `rsync` for the new `gcloud storage` CLI soon

Do you mean that `rsync` will be **added** to the `gcloud storage` CLI? Or what kind of redesign do you mean? Currently I don't see `rsync` in the list of available collands: https://cloud.google.com/sdk/gcloud/reference/storage.

NickGoog: Yes, we're still designing `rsync` for `gcloud storage`. It's taking a while because we're prototyping other algorithms to see if we can get performance improvements.

PR #1198, opened 2021-02-12
Update "surprising naming" section with additional suggestion

MikeJeffrey:
Adding a new bullet to the "surprising subdirectory naming" workarounds list, to recommend using a trailing slash to indicate to gsutil that the target is a directory.

### DIFF
--- a/gslib/addlhelp/subdirs.py
+++ b/gslib/addlhelp/subdirs.py
@@ -121,13 +121,21 @@ _DETAILED_HELP_TEXT = ("""
   the second attempt will encounter an already existing source subdirectory
   and result in the above-described naming problem.
 
-  There are a couple of ways to avoid this problem:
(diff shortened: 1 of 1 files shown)


NickGoog (approved):

starsandskies:
Did you have a chance to test the behavior Mike? Looking closely at b/112340922 (the user feedback that motivates this update), I think they're saying the output we document on line 108 of cp.py doesn't match the actual behavior.

MikeJeffrey:
My interpretation was that it works as expected when the trailing slash is included. We mention the need for a trailing slash in the 1st ordered list item; then we go on to explain some weirdness that occurs if the trailing slash is *not* included.

I have to admit that my brain gets tied in knots trying to follow the examples in cp.py :)

PR #1199, opened 2021-02-12
Update wildcards.py

MikeJeffrey:
Calling out that some shells require quotes around wildcarded expressions.

### DIFF
--- a/gslib/addlhelp/wildcards.py
+++ b/gslib/addlhelp/wildcards.py
@@ -30,6 +30,10 @@ _DETAILED_HELP_TEXT = ("""
   will copy all objects that start with gs://bucket/data/abc followed by any
   number of characters within that subdirectory.
 
+  Note: Some shells require that wildcarded expressions be surrounded with
+  single quotes (on Linux) or double quotes (on Windows). For example:
+  
+    gsutil cp "gs://bucket/data/abc*" .
 
 <B>DIRECTORY BY DIRECTORY VS RECURSIVE WILDCARDS</B>
   The "*" wildcard only matches up to the end of a path within


starsandskies:
Looks good to me, but adding a gsutil SWE who'll be better able to review and approve!

starsandskies:
Did you have a chance to test the behavior Mike? Looking closely at b/112340922 (the user feedback that motivates this update), I think they're saying the output we document on line 108 of cp.py doesn't match the actual behavior.

rrauber:
There's a similar note in there already that's kind of buried: https://github.com/GoogleCloudPlatform/gsutil/blob/9460b6d43e0321c8d243fc108a43a6f359f6b1f5/gslib/addlhelp/wildcards.py#L57

Do you all think we should consolidate this information a bit more? Is the line linked above still needed?

MikeJeffrey:
> There's a similar note in there already that's kind of buried:
> 
> https://github.com/GoogleCloudPlatform/gsutil/blob/9460b6d43e0321c8d243fc108a43a6f359f6b1f5/gslib/addlhelp/wildcards.py#L57
> 
> Do you all think we should consolidate this information a bit more? Is the line linked above still needed?

You're right, it could be consolidated. My vote would be to move the note to the new location, closer to the top, since it applies to everything that comes after it; but I'm certainly open to discussion! What do you think?

starsandskies:
Since this has been quiet for awhile, I'll weigh in :P

I also vote for moving/removing the existing note and merging it with the newly proposed note. Mike, I think you can make the change. If this PR has been idle for too long and you've moved on / lost context on it, just let me know and I can pick it up!

PR #1214, opened 2021-03-18
Removed kms check when determining if cp should do a parallel upload.

nolanleastin:
gsutil will now be able to use compose to parallel upload large objects.


rrauber (changes requested):
The change looks great, but could you add a test covering a composite upload with KMS? It would look a lot like this test [here](https://github.com/nolanleastin/gsutil/blob/remove-kms-compose-restrictions/gslib/tests/test_cp.py#L4440), but you'd need to tune some config parameters as done [here](https://github.com/nolanleastin/gsutil/blob/remove-kms-compose-restrictions/gslib/tests/test_cp.py#L2984) to ensure a composite upload occurs.

rrauber (approved):

PR #1225, opened 2021-04-13
Doc clarifications

NickGoog:
Addresses a few bugs we've received.

### DIFF
--- a/gslib/commands/mb.py
+++ b/gslib/commands/mb.py
@@ -125,7 +125,7 @@ _DETAILED_HELP_TEXT = ("""
   retention policy.
 
 <B>UNIFORM BUCKET-LEVEL ACCESS</B>
-  You can specify turn on uniform bucket-level access for a bucket
+  You can enable or disable uniform bucket-level access for a bucket
   with the -b option.
 
   Examples:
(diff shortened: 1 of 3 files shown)


gslib/commands/mb.py
```diff
@@ -125,7 +125,7 @@
   retention policy.
 
 <B>UNIFORM BUCKET-LEVEL ACCESS</B>
-  You can specify one of the available settings for a bucket
+  You can specify turn on uniform bucket-level access for a bucket
```
starsandskies, line 128: Suggestion: "You can enable or disable uniform bucket-level access for a bucket with the -b option"

gslib/commands/rb.py
```diff
@@ -51,6 +51,8 @@
   -f          Continues silently (without printing error messages) despite
               errors when removing buckets. If some buckets couldn't be removed,
               gsutil's exit status will be non-zero even if this flag is set.
+              If no buckets could be removed, the command will raise a
```
starsandskies, line 54: "will raise" --> "raises"

starsandskies (approved):
Two nits, but overall LGTM!

gslib/commands/mb.py
NickGoog: Done

gslib/commands/rb.py
NickGoog: Done

PR #1240, opened 2021-04-16
Fix resumable upload tests

rrauber:
Some tests weren't working as intended because progress callbacks are called less frequently than they assumed. See the conditions [here](https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/progress_callback.py#L128).

### DIFF
--- a/gslib/tests/util.py
+++ b/gslib/tests/util.py
@@ -667,11 +667,10 @@ class HaltingCopyCallbackHandler(object):
   def call(self, total_bytes_transferred, total_size):
     """Forcibly exits if the transfer has passed the halting point.
 
-    Note that this function is only called when the linked conditions are
-    met: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/progress_callback.py#L128.
-
-    This means that self._halt_at_byte is only precise if it's divisible by
-    progress_callback._START_BYTES_PER_CALLBACK.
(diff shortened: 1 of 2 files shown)


gslib/tests/util.py
```diff
@@ -665,7 +665,14 @@ def __init__(self, is_upload, halt_at_byte):
 
   # pylint: disable=invalid-name
   def call(self, total_bytes_transferred, total_size):
-    """Forcibly exits if the transfer has passed the halting point."""
+    """Forcibly exits if the transfer has passed the halting point.
+
+    Note that this function is only called when the linked conditions are
+    met: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/progress_callback.py#L128.
```
NickGoog, line 671: Seems weird to HTTP link something in the same codebase. Maybe just give the file path?
I wouldn't provide the line number b/c it might change.

gslib/tests/util.py
```diff
@@ -665,7 +665,14 @@ def __init__(self, is_upload, halt_at_byte):
...
+
+    This means that self._halt_at_byte is only precise if it's divisible by
```
NickGoog, line 673: I would strike "This means that" from this sentence.

NickGoog (approved):

PR #1243, opened 2021-04-16
Enforce custom endpoints for complex transfers (JSON).

NickGoog:
gsutil doesn't have unit tests for the `GcsJsonApi` upload and download methods, and it looks like we mostly rely on `cp` integration tests for coverage.

I'm planning a design doc for E2E tests, so we should eventually get test coverage for all this code.

### DIFF
--- a/gslib/gcs_json_api.py
+++ b/gslib/gcs_json_api.py
@@ -40,6 +40,7 @@ from apitools.base.py.util import CalculateWaitForRetry
 from boto import config
 import httplib2
 import oauth2client
+from six.moves import urllib
 
 from gslib import context_config
 from gslib.cloud_api import AccessDeniedException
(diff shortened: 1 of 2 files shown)


gslib/gcs_json_api.py
```diff
@@ -1162,14 +1194,20 @@ def GetObjectMedia(self,
...
       # TODO: Add query parameter support to apitools downloads so there is
       # a well-defined way to express query parameters. Currently, we assume
       # the URL ends in ?alt=media, and this will break if that changes.
       if self.trace_token:
-        serialization_dict = json.loads(six.ensure_str(serialization_data))
         serialization_dict['url'] += '&trace=token%%3A%s' % self.trace_token
-        serialization_data = json.dumps(serialization_dict)
+
+      # Maintain custom endpoints that might have been overrided.
+      serialization_dict['url'] = self.api_client.FinalizeTransferUrl(
```
dilipped, line 1207: Do you need to call this here? Won't this be called by Apitools anyway given that you have patched the method already on line 338?

gslib/gcs_json_api.py
```diff
@@ -174,6 +175,34 @@
...
+      For example, this would be the "host.com:80" in "http://host.com:80/path."
+
+    Returns:
+      Patched FinalizeTransferUrl function.
+  """
+
+  def FinalizeTransferUrl(url):
+    modified_url = original_function(url)
+    parsed_url = urllib.parse.urlsplit(modified_url)
+    url_components = list(parsed_url)
```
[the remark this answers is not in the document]
NickGoog: Since I have to do an "arbitrary index assignment" on the next line, I want to be as clear as possible before that.
I don't like arbitrary index assignments like "url_components[1] = ..." because "1" seems like a magic number to a reader.
Normally, the solution is to deconstruct, but there are 5+ items in `url_components`, and we only need to change one of them, so it's clunky to split it up like, `http, netloc, path, auth, params, ... = list(parsed_url)`.

I didn't combine lines because then I could use descriptive variables that makes it easier for the reader to follow what's going on. I hope that compensates a bit for the arbitrary index assignment if someone has to debug this later.

gslib/gcs_json_api.py
NickGoog: Part of the cause of this bug is that Apitools doesn't call `FinalizeTransferUrl` on URLs received through `FromData`.

dilipped: So in that case, should there be a fix in Apitools?

NickGoog: It might be an intentional Apitools decision, and it could break other things, so I'd say no

NickGoog:
Replacing with: https://github.com/GoogleCloudPlatform/gsutil/pull/1250

PR #1247, opened 2021-04-19
XML API PSC Fix

NickGoog:
Overwrite server-returned URL with custom endpoint for complex downloads (XML API).

gsutil doesn't have unit tests for the `boto_translation.py`, and it looks like we mostly rely on `cp` integration tests to cover `GetObjectMedia`.

I'm planning a design doc for E2E tests, so we should eventually get test coverage for all this code.

### DIFF
--- a/gslib/boto_translation.py
+++ b/gslib/boto_translation.py
@@ -40,6 +40,7 @@ from xml.sax import _exceptions as SaxExceptions
 import six
 from six.moves import http_client
 import boto
+from boto import config
 from boto import handler
 from boto.gs.cors import Cors
 from boto.gs.lifecycle import LifecycleConfig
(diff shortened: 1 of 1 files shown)


gslib/boto_translation.py
```diff
@@ -39,7 +39,9 @@
 
 import six
 from six.moves import http_client
+from six.moves import urllib
```
dilipped, line 42: Not used?

dilipped (dismissed):

gslib/boto_translation.py
```diff
@@ -546,6 +561,7 @@ def GetObjectMedia(self,
     if serialization_data:
       serialization_dict = json.loads(serialization_data)
       key = pickle.loads(binascii.a2b_base64(serialization_dict['url']))
+      _AddCustomEndpointToKey(key)
```
dilipped, line 564: Will this run for S3 interaction as well? IN that case, will we end up adding the gs_host for s3?

gslib/boto_translation.py
NickGoog: Oops

gslib/boto_translation.py
NickGoog: Yes, it looks like `BotoTranslation` is used for both S3 and GCS XML.
Added a provider check!

dilipped (approved):

PR #1251, opened 2021-04-23
Fix daisy chain for windows

dilipped:
The Daisy chain copy for GCS to S3 is broken on windows because of different read and write size which happens as Windows uses 64KB TRANSFER_BUFFER_SIZE as of https://github.com/GoogleCloudPlatform/gsutil/commit/ffa1eb7d4d9fe8dd877452f28f55fdf51c2a6c12 and the daisy chain logic expects same read and write size. However, the http.client module will read in 8KiB chunks. 

This PR modifies the read logic so that a read size < TRANSFER_BUFFER_SIZE can be handled.

For reference: b/185961088

### DIFF
--- a/gslib/daisy_chain_wrapper.py
+++ b/gslib/daisy_chain_wrapper.py
@@ -269,7 +269,7 @@ class DaisyChainWrapper(object):
       # return 'foo', and in the subsequent calls
       # we will return 'barba' and 'z'. This is done with the assumption
       # that the length of data that gets written to self.buffer is always
-      # a multiple of "amt", which means the chances of such cases are pretty
+      # a multiple of "amt", which means the chance of such cases are pretty
       # rare. This assumption helps us keep this logic simple and avoid
       # string concatenations and splits ('foo' + 'ba' in the above case).
       if not valid_data:
--- a/gslib/daisy_chain_wrapper.py
+++ b/gslib/daisy_chain_wrapper.py
@@ -161,6 +161,7 @@ class DaisyChainWrapper(object):
     self.StartDownloadThread(progress_callback=self.progress_callback)
     if not self.download_started.wait(60):
       raise Exception('Could not start download thread after 60 seconds.')
+    self._unused_data_from_previous_read = b''
 
   def StartDownloadThread(self, start_byte=0, progress_callback=None):  # pylint: disable=invalid-name
     """Starts the download thread for the source object (from start_byte)."""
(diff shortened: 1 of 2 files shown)


gslib/tests/test_daisy_chain_wrapper.py
```diff
@@ -151,6 +151,38 @@ def testDownloadMultiChunk(self):
       with open(self.test_data_file, 'rb') as download_stream:
         self.assertEqual(upload_stream.read(), download_stream.read())
 
+  def testDownloadWithDifferentChunkSize(self):
+    """Tests multiple calls to GetObjectMedia."""
+    upload_file = self.CreateTempFile()
+    write_values = []
+    with open(self.test_data_file, 'rb') as stream:
+      # Use an arbitrary size for writing data to the buffer.
```
NickGoog, line 159: An arbitrary size "greater than TRANSFER_BUFFER_SIZE"?

gslib/daisy_chain_wrapper.py
```diff
@@ -239,34 +240,53 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
+              raise Exception('Download thread died suddenly.')
+        # Buffer was empty, yield thread priority so the download thread can fill.
+        time.sleep(0)
     with self.lock:
-      # TODO: Need to handle the caller requesting less than a
-      # transfer_buffer_size worth of data.
-      data = self.buffer.popleft()
+      # Note that we are not handling the rare case where len(valid_data) < amt
+      # and there are elements in the buffer.
+      # For e.g if valid_data = 'foo' and amt = 5, and we have 'barbaz' in
```
NickGoog, line 267: e.g. = "For example", so writing "For e.g." = "For for example"
https://www.aje.com/arc/editing-tip-using-eg-and-ie/#:~:text=e.g.%20is%20the%20abbreviation%20for,implies%20that%20there%20are%20other

gslib/daisy_chain_wrapper.py
```diff
@@ -239,34 +240,53 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
+      # self.buffer, instead of returning 'fooba' in the current call, we
+      # return 'foo' and in the subsequent calls we will return 'barba' and 'z'
```
NickGoog, line 269: Comma after `'foo'` and period after `'z'`

gslib/daisy_chain_wrapper.py
```diff
@@ -239,34 +240,53 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
+          else:
+            if not self.download_thread.is_alive():
+              raise Exception('Download thread died suddenly.')
+        # Buffer was empty, yield thread priority so the download thread can fill.
+        time.sleep(0)
     with self.lock:
-      # TODO: Need to handle the caller requesting less than a
-      # transfer_buffer_size worth of data.
-      data = self.buffer.popleft()
+      # Note that we are not handling the rare case where len(valid_data) < amt
```
NickGoog, line 265: Comma after `amt`
https://owl.purdue.edu/owl/general_writing/punctuation/independent_and_dependent_clauses/index.html

gslib/daisy_chain_wrapper.py
```diff
@@ -239,34 +240,53 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
+      # and there are elements in the buffer.
+      # For e.g if valid_data = 'foo' and amt = 5, and we have 'barbaz' in
+      # self.buffer, instead of returning 'fooba' in the current call, we
+      # return 'foo' and in the subsequent calls we will return 'barba' and 'z'
+      # This is done with the assumption in mind that the length of data
+      # that gets written to self.buffer is always in the multiple of "amt",
```
NickGoog, line 271: "in the multiple of" -> "a multiple of" ?

gslib/tests/test_daisy_chain_wrapper.py
```diff
@@ -151,6 +151,38 @@ def testDownloadMultiChunk(self):
...
+      # For reading from the buffer WriteFromWrapperToFile will
+      # use TRANSFER_BUFFER_SIZE.
+      buffer_write_size = TRANSFER_BUFFER_SIZE * 2 + 10
+      while True:
+        # Write data with size
```
NickGoog, line 164: Period

gslib/tests/test_daisy_chain_wrapper.py
```diff
@@ -151,6 +151,38 @@ def testDownloadMultiChunk(self):
...
+        write_values.append(data)
+    mock_api = self.MockDownloadCloudApi(write_values)
+    daisy_chain_wrapper = DaisyChainWrapper(
+        self._dummy_url,
+        self.test_data_file_len,
+        mock_api,
+        download_chunk_size=TRANSFER_BUFFER_SIZE)
+    self._WriteFromWrapperToFile(daisy_chain_wrapper, upload_file)
+    num_expected_calls = self.test_data_file_len // TRANSFER_BUFFER_SIZE
+    if self.test_data_file_len % TRANSFER_BUFFER_SIZE:
```
NickGoog, line 177: Can you accomplish this in one line with:
`num_expected_calls = math.ceil(self.test_data_file_len / TRANSFER_BUFFER_SIZE)`

gslib/daisy_chain_wrapper.py
```diff
@@ -239,34 +240,53 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
+      # which means the chances of such cases are pretty rare.
+      # This assumption helps us keep this logic simple and avoid
+      # string concatenations and splits ('foo' + 'ba' in the above case).
```
NickGoog, line 274: If I understand this correctly, in the "rare case," the only negative is an extra read call?

gslib/tests/test_daisy_chain_wrapper.py
dilipped: Done

gslib/daisy_chain_wrapper.py
dilipped: Done

gslib/daisy_chain_wrapper.py
dilipped: Done

gslib/daisy_chain_wrapper.py
dilipped: Done

gslib/tests/test_daisy_chain_wrapper.py
dilipped: Done

gslib/tests/test_daisy_chain_wrapper.py
dilipped: Good idea

gslib/daisy_chain_wrapper.py
dilipped: It might result in multiple extra calls. e.g say you are writing to the buffer in chunks of 64KiB, total data is 640KiB that means you have 10 such chunks in the buffer. Now assume that your read size is 10KiB, in this case, for the current solution, you will make 7 reads calls per chunk. 6 reads calls returning 10KiB each and one read which will return the remaining 4 KiB for that chunk. So overall you will have 7 x 10 = 70 calls being made.

If we do not do this, and instead concatenate that 4KiB data with the next chunk, we will make 64 read calls in total.

But I said that this is a rare case because the read size is hard coded in python's http.client to be 8KiB and our transfer buffer size is of 64KiB, so unless someone do some changes to this hard coded value, we should be good.

gslib/daisy_chain_wrapper.py
dilipped: Done

gslib/daisy_chain_wrapper.py
```diff
@@ -262,15 +262,15 @@ def read(self, amt=None):  # pylint: disable=invalid-name
...
       # self.buffer, instead of returning 'fooba' in the current call, we
-      # return 'foo' and in the subsequent calls we will return 'barba' and 'z'
-      # This is done with the assumption in mind that the length of data
-      # that gets written to self.buffer is always in the multiple of "amt",
-      # which means the chances of such cases are pretty rare.
-      # This assumption helps us keep this logic simple and avoid
+      # return 'foo', and in the subsequent calls
+      # we will return 'barba' and 'z'. This is done with the assumption
+      # that the length of data that gets written to self.buffer is always
+      # a multiple of "amt", which means the chances of such cases are pretty
```
NickGoog, line 272: "chance" not "chances"

NickGoog (approved):

gslib/daisy_chain_wrapper.py
dilipped: Done

PR #1263, opened 2021-05-07
Specify the purpose of -p option for -D example

matarrese:
The examples for -D option is confusing, as -p option is also specified. 

Some users end up thinking that these two options should be used together. This is not always the case (e.g. for bucket with uniform bucket-level access),  and an error will be returned in case OWNER permission is missing as OWNER permission is required for preserving ACLs. 
This cl provides additional clarifications to avoid further support tickets.


matarrese:
@starsandskies let me know if this is OK now.

starsandskies:
Looks both simpler and more relevant to me :)

Adding a gsutil engineer to the reviewer list for their thumbs up

NickGoog (approved):

PR #1271, opened 2021-05-13
Fix gsutil help for -o

### LINKED ISSUE #1270: gsutil flags help for `-o` is misleading
The usage form shows:
https://github.com/GoogleCloudPlatform/gsutil/blob/7ac2a5b4bd3a287aaa07d858c0fc86972c087f3e/gslib/commands/help.py#L77

The help for, e.g. `gsutil cp` shows something else:
https://github.com/GoogleCloudPlatform/gsutil/blob/7ac2a5b4bd3a287aaa07d858c0fc86972c087f3e/gslib/commands/cp.py#L528-L529

It seems like the proper usage would be:

`[-o section:flag=value]`

For reference, here's the implementation:
https://github.com/GoogleCloudPlatform/gsutil/blob/7ac2a5b4bd3a287aaa07d858c0fc86972c087f3e/gslib/__main__.py#L347-L356

jsoref:
`-o`/`--option` takes a single mandatory parameter of a fixed form and this should be clear in the help
`-o` flags can be included more than once (as with `-h`).

Fixes #1270

### DIFF
--- a/gslib/commands/help.py
+++ b/gslib/commands/help.py
@@ -74,7 +74,8 @@ _DETAILED_HELP_TEXT = ("""
 
 top_level_usage_string = (
     'Usage: gsutil [-D] [-DD] [-h header]... [-i service_account] '
-    '[-m] [-o section:flag=value]... [-q] [-u user_project] [command [opts...] args...]')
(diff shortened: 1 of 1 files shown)


dilipped (commented):
Thanks for the PR! Just a formatting fix is needed.

Issue #800, opened 2019-06-07
gsutil mv skips the trailing slashes but delete the file


AnkitSwarnkar:
I was trying to move one file from a gs location to another gs location on executing gsutil mv commands However I missed "gs://" in the destination. 
It attempted the file move but the attempt was skipped. However, once skipped it deleted the original file.
Copying gs://<bucket>/2019-05-19_2019-05-25/data_20190519-20190525.csv.gz...
Skipping attempt to download to filename ending with slash (target/sub/dir). This
typically happens when using gsutil to download from a subdirectory
created by the Cloud Console (https://cloud.google.com/console)
Removing  gs://<bucket>/2019-05-19_2019-05-25/data_20190519-20190525.csv.gz...

Ideally, remove should not happen if the previous action got skipped

lopezvit:
After all this time, this issue haven't been solved? It just happened to me and totally lost the original file.

YannisPap:
Happened to us also. In our case, it was a mistyped `gs:/bucketname` instead of `gs://bucketname`

PR #1273, opened 2021-05-14
Raise error for invalid destination.

dilipped:
Fixes https://github.com/GoogleCloudPlatform/gsutil/issues/800

Raising an error now if destination ends with a delimiter.

### DIFF
--- a/gslib/utils/copy_helper.py
+++ b/gslib/utils/copy_helper.py
@@ -2969,7 +2969,7 @@ def _DownloadObjectToFile(src_url,
             'typically happens when using gsutil to download from a subdirectory '
             'created by the Cloud Console (https://cloud.google.com/console)' %
             dst_url.object_name)))
-    # The warning above is needed because before this error might get ignored
+    # The warning above is needed because errors might get ignored
     # for parallel processing.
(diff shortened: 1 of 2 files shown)


gslib/utils/copy_helper.py
```diff
@@ -2968,7 +2969,7 @@ def _DownloadObjectToFile(src_url,
             'typically happens when using gsutil to download from a subdirectory '
             'created by the Cloud Console (https://cloud.google.com/console)' %
             dst_url.object_name)))
-    return (0, 0, dst_url, '')
+    raise InvalidUrlError('Invalid destination path: %s' % dst_url.object_name)
```
NickGoog, line 2972: What's the point of the warning if we're going to raise an error?
Maybe the warning text should be merged into the error?

dilipped: Error may get ignored if you are using parallel processing or if you are using the continue-on-error flag.

NickGoog: Ahh, maybe add a code comment about that

dilipped: Done.

NickGoog (approved):

PR #1280, opened 2021-05-24
Fix pyenv issue for macOS

dilipped:
See b/187701234#comment12 for context.

### DIFF
--- a/test/ci/kokoro/run_integ_tests.sh
+++ b/test/ci/kokoro/run_integ_tests.sh
@@ -60,6 +60,7 @@ function install_pyenv {
     export PYENV_ROOT="$HOME/.pyenv"
     export PATH="$PYENV_ROOT/bin:$PATH"
     eval "$(pyenv init -)"
+    pyenv init
   fi
   pyenv update
   # To address pyenv issue: See b/187701234#comment12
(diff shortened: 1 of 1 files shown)


test/ci/kokoro/run_integ_tests.sh
```diff
@@ -54,18 +54,16 @@ function latest_python_release {
...
-    if [[ $KOKORO_JOB_NAME =~ "macos" ]]; then
-      brew update
-      brew install pyenv
-      eval "$(pyenv init -)"
-    fi
+    # MacOS VM does not have pyenv installed by default.
+    git clone https://github.com/pyenv/pyenv.git ~/.pyenv
+    export PYENV_ROOT="$HOME/.pyenv"
+    export PATH="$PYENV_ROOT/bin:$PATH"
+    eval "$(pyenv init --path)"
```
NickGoog, line 61: Why not just run `pyenv init --path`?

NickGoog (approved):

NickGoog:
Make sure this isn't your invocation: http://sponge2/50a86890-70c3-4502-9dd0-ba6184299c14

test/ci/kokoro/run_integ_tests.sh
dilipped: The way that command works is that it will output a bunch of other commands, and we want to execute those.

dilipped:
> Make sure this isn't your invocation: http://sponge2/50a86890-70c3-4502-9dd0-ba6184299c14

That one is from continous build, so it is not using this code.

PR #1288, opened 2021-06-02
Fix flaky test for rm using preconditions

dilipped:
Rm tests are occasionally failing because the backend gives two different messages.

```
gsutil -h x-goog-if-generation-match:12345 rm gs://gsutil_test_bucket-4/t
est.txt
Removing gs://gsutil_test_bucket-4/test.txt...
PreconditionException: 412 At least one of the pre-conditions you specified did not hold.
```
and
```
$ gsutil -h x-goog-if-generation-match:12345 rm gs://gsutil_test_bucket-4/test.txt
Removing gs://gsutil_test_bucket-4/test.txt...
PreconditionException: 412 Precondition Failed
```

Making the test less strict so that it can work for both the messages.

### DIFF
--- a/gslib/tests/test_rm.py
+++ b/gslib/tests/test_rm.py
@@ -611,8 +611,7 @@ class TestRm(testcase.GsUtilIntegrationTestCase):
          suri(object_uri)],
         return_stderr=True,
         expected_status=1)
-    self.assertRegex(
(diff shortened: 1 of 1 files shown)


dilipped:
Merging this without review as this is blocking the release. I can revert it post merging if there are any objections.

PR #1302, opened 2021-06-30
Onboard mTLS support with AIP-4114 for gsutil

shinfan:
- The certificate discovery has the following logic:
   - Use the user specified command in `.boto` file if available
   - Otherwise, automatically locate the SecureConnectHelper Metadata JSON

Notes: This PR is backwards-compatible

For more information about the mTLS spec please read: [AIP-4114](https://google.aip.dev/auth/4114)

### DIFF
--- a/gslib/context_config.py
+++ b/gslib/context_config.py
@@ -20,8 +20,10 @@ from __future__ import division
 from __future__ import unicode_literals
 
 import atexit
+import json
 import os
 import subprocess
+import six
 
 from boto import config
 
@@ -30,15 +32,16 @@ import gslib
 # Maintain a single context configuration.
 _singleton_config = None
 
+# Metadata JSON that stores information about the default certificate provider
+_CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
+_CERT_PROVIDER_COMMAND = "cert_provider_command"
+_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION = "--with_passphrase"
 
 class CertProvisionError(Exception):
   """Represents errors when provisioning a client certificate."""
-  pass
-
 
 class ContextConfigSingletonAlreadyExistsError(Exception):
   """Error for when create_context_config is called multiple times."""
-  pass
 
 
 def _IsPemSectionMarker(line):
@@ -107,6 +110,58 @@ def _SplitPemIntoSections(contents, logger):
   return result
 
 
+def _check_path():
+  """Checks for context aware metadata. If it exists, returns the absolute path;
+  otherwise returns None.
+  Returns:
+      str: absolute path if exists and None otherwise.
+  """
+  metadata_path = os.path.expanduser(_CONTEXT_AWARE_METADATA_PATH)
+  if not os.path.exists(metadata_path):
+    _LOGGER.debug("%s is not found, skip client SSL authentication.", metadata_path)
+    return None
+  return metadata_path
+
+
+def _read_metadata_file(metadata_path):
+  """Loads context aware metadata from the given path.
+  Returns:
+      Dict[str, str]: The metadata.
+  Raises:
+      google.auth.CertProvisionError: If failed to parse metadata as JSON.
+  """
+  try:
+    with open(metadata_path) as f:
+      metadata = json.load(f)
+  except ValueError as caught_exc:
+    new_exc = CertProvisionError(caught_exc)
+    six.raise_from(new_exc, caught_exc)
+
+  return metadata
+
+
+def _default_command():
+  """Loads default cert provider command
+  Returns:
+      str: The default command
+  Raises:
+      google.auth.CertProvisionError: If command cannot be found
+  """
+  metadata_path = _check_path()
+  if metadata_path:
+    metadata_json = _read_metadata_file(metadata_path)
+
+    if _CERT_PROVIDER_COMMAND not in metadata_json:
+      raise CertProvisionError("Client certificate provider is not found")
+
+    command = metadata_json[_CERT_PROVIDER_COMMAND]
+    if (_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION not in command):
+      command.append(_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION)
+    return command
+
+  raise CertProvisionError("Client certificate provider is not found")
+
+
 class _ContextConfig(object):
   """Represents the configurations associated with context aware access.
 
@@ -124,37 +179,30 @@ class _ContextConfig(object):
     self.use_client_certificate = config.getbool('Credentials',
                                                  'use_client_certificate')
     self.client_cert_path = None
-    self.client_cert_password = None
-
     if not self.use_client_certificate:
       # Don't spend time generating values gsutil won't need.
       return
 
     # Generates certificate and deletes it afterwards.
-    atexit.register(self._UnprovisionClientCert)
-    command_string = config.get('Credentials', 'cert_provider_command', None)
-    if not command_string:
-      raise CertProvisionError('No cert provider detected.')
-
+    # atexit.register(self._UnprovisionClientCert)
     self.client_cert_path = os.path.join(gslib.GSUTIL_DIR, 'caa_cert.pem')
     try:
       # Certs provisioned using endpoint verification are stored as a
       # single file holding both the public certificate and the private key.
-      self._ProvisionClientCert(command_string, self.client_cert_path)
+      self._ProvisionClientCert(self.client_cert_path)
     except CertProvisionError as e:
       self.logger.error('Failed to provision client certificate: %s' % e)
 
-  def _ProvisionClientCert(self, command_string, cert_path):
+
+  def _ProvisionClientCert(self, cert_path):
     """Executes certificate provider to obtain client certificate and keys."""
-    # Monkey-patch command line args to get password-protected certificate.
-    # Adds password flag if it's not already there.
-    password_arg = ' --with_passphrase'
-    if ('--print_certificate' in command_string and
-        password_arg not in command_string):
-      command_string += password_arg
+    cert_command = config.get('Credentials', 'cert_provider_command', None)
+    if not cert_command:
+      # Load the default certificate provider if sit is not provided by user.
+      cert_command = _default_command()
 
     try:
-      command_process = subprocess.Popen(command_string.split(' '),
+      command_process = subprocess.Popen(cert_command,
                                          stdout=subprocess.PIPE,
                                          stderr=subprocess.PIPE)
       command_stdout, command_stderr = command_process.communicate()
@@ -168,12 +216,13 @@ class _ContextConfig(object):
       with open(cert_path, 'w+') as f:
         f.write(sections['CERTIFICATE'])
         f.write(sections['ENCRYPTED PRIVATE KEY'])
-      self.client_cert_password = sections['PASSPHRASE'].splitlines()[1]
+        self.client_cert_password = sections['PASSPHRASE'].splitlines()[1]
     except OSError as e:
       raise CertProvisionError(e)
     except KeyError as e:
       raise CertProvisionError(
           'Invalid output format from certificate provider, no %s' % e)
+
 
   def _UnprovisionClientCert(self):
     """Cleans up any files or resources provisioned during config init."""
--- a/gslib/context_config.py
+++ b/gslib/context_config.py
@@ -36,9 +36,11 @@ _CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
 _CERT_PROVIDER_COMMAND = "cert_provider_command"
 _CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION = "--with_passphrase"
 
+
 class CertProvisionError(Exception):
   """Represents errors when provisioning a client certificate."""
   pass
+
 
 class ContextConfigSingletonAlreadyExistsError(Exception):
   """Error for when create_context_config is called multiple times."""
@@ -191,7 +193,6 @@ class _ContextConfig(object):
     except CertProvisionError as e:
       self.logger.error('Failed to provision client certificate: %s' % e)
 
-
   def _ProvisionClientCert(self, cert_path):
     """Executes certificate provider to obtain client certificate and keys."""
     cert_command = config.get('Credentials', 'cert_provider_command', None)
@@ -220,7 +221,6 @@ class _ContextConfig(object):
     except KeyError as e:
       raise CertProvisionError(
           'Invalid output format from certificate provider, no %s' % e)
-
 
   def _UnprovisionClientCert(self):
     """Cleans up any files or resources provisioned during config init."""
--- a/gslib/context_config.py
+++ b/gslib/context_config.py
@@ -23,7 +23,6 @@ import atexit
 import json
 import os
 import subprocess
-import six
 
 from boto import config
 
@@ -39,9 +38,11 @@ _CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION = "--with_passphrase"
 
 class CertProvisionError(Exception):
   """Represents errors when provisioning a client certificate."""
+  pass
 
 class ContextConfigSingletonAlreadyExistsError(Exception):
   """Error for when create_context_config is called multiple times."""
+  pass
 
 
 def _IsPemSectionMarker(line):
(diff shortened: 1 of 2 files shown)


gslib/context_config.py
```diff
@@ -107,6 +110,56 @@ def _SplitPemIntoSections(contents, logger):
...
+      str: The default command
+  Raises:
+      google.auth.CertProvisionError: If command cannot be found
+  """
+  metadata_path = _check_path()
+  if metadata_path:
+    metadata_json = _read_metadata_file(metadata_path)
+
+    if _CERT_PROVIDER_COMMAND not in metadata_json:
+      raise CertProvisionError("Client certificate provider is not found")
```
NickGoog, line 153: Please add a period at the end of the sentence. Google Python style guide likes perfect grammar.

gslib/context_config.py
```diff
@@ -107,6 +110,56 @@ def _SplitPemIntoSections(contents, logger):
...
+
+    command = metadata_json[_CERT_PROVIDER_COMMAND]
+    if (_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION not in command):
+      command.append(_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION)
+    return command
+
+  raise CertProvisionError("Client certificate provider is not found")
```
NickGoog, line 160: Ditto

gslib/context_config.py
```diff
@@ -107,6 +110,58 @@ def _SplitPemIntoSections(contents, logger):
...
+      metadata = json.load(f)
+  except ValueError as caught_exc:
+    new_exc = CertProvisionError(caught_exc)
+    six.raise_from(new_exc, caught_exc)
+
+  return metadata
+
+
+def _default_command():
+  """Loads default cert provider command
```
NickGoog, line 142: Docstring needs periods at the end of this line, the "str:" returns line, and the "CertProvisionError" line.

gslib/context_config.py
```diff
@@ -30,6 +31,11 @@
 # Maintain a single context configuration.
 _singleton_config = None
 
+# Metadata JSON that stores information about the default certificate provider
```
NickGoog, line 34: Add a period at the end of the comment please

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
   return result
 
 
+def _check_path():
+  """Checks for context aware metadata. If it exists, returns the absolute path;
+  otherwise returns None.
+  Returns:
```
NickGoog, line 119: Hey, I shouldn't post internal links here, so can you search and check the internal Python style guide for docstring format?

It should be something like:

```
def _check_path():
  """Checks for content aware metadata.

  If content aware metadata exists, return its absolute path;
  otherwise, returns None.

  Returns:
    str: Absolute path if exists. Otherwise, None.
```

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
...
+  """
+  metadata_path = os.path.expanduser(_CONTEXT_AWARE_METADATA_PATH)
+  if not os.path.exists(metadata_path):
+    return None
+  return metadata_path
+
+
+def _read_metadata_file(metadata_path):
+  """Loads context aware metadata from the given path.
+  Returns:
```
NickGoog, line 130: Needs line breaks above "Returns" and "Raises"

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
...
+    with open(metadata_path) as f:
+      metadata = json.load(f)
+  except ValueError as e:
+    raise CertProvisionError(e)
+  return metadata
+
+
+def _default_command():
+  """Loads default cert provider command.
+  Returns:
```
NickGoog, line 145: Same docstring formatting issue

gslib/context_config.py
```diff
@@ -124,37 +179,29 @@ def __init__(self, logger):
...
       # Don't spend time generating values gsutil won't need.
       return
 
     # Generates certificate and deletes it afterwards.
-    atexit.register(self._UnprovisionClientCert)
-    command_string = config.get('Credentials', 'cert_provider_command', None)
-    if not command_string:
-      raise CertProvisionError('No cert provider detected.')
-
+    # atexit.register(self._UnprovisionClientCert)
```
NickGoog, line 187: Does this need to be uncommented?

gslib/context_config.py
```diff
@@ -30,6 +31,11 @@
 # Maintain a single context configuration.
 _singleton_config = None
 
+# Metadata JSON that stores information about the default certificate provider
+_CONTEXT_AWARE_METADATA_PATH = "~/.secureConnect/context_aware_metadata.json"
```
NickGoog, line 35: Maybe add "DEFAULT" to this constant name?

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
...
+def _read_metadata_file(metadata_path):
+  """Loads context aware metadata from the given path.
+  Returns:
+      Dict[str, str]: The metadata.
+  Raises:
+      CertProvisionError: If failed to parse metadata as JSON.
+  """
+  try:
+    with open(metadata_path) as f:
+      metadata = json.load(f)
```
NickGoog, line 137: Why not `return json.load(f)`?

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
...
+
+def _default_command():
+  """Loads default cert provider command.
+  Returns:
+      str: The default command.
+  Raises:
+      CertProvisionError: If command cannot be found.
+  """
+  metadata_path = _check_path()
+  if metadata_path:
```
NickGoog, line 151: I would restructure this as:

```
if not metadata_path:
  raise

metadata_json = _read...
```

That way, you can unnest most of the code, and you can stop reading earlier if you know _check_path returned false. Speaking of which, I would recommend differentiating the two `CertProvisionError` messages a little.

gslib/context_config.py
```diff
@@ -107,6 +113,55 @@ def _SplitPemIntoSections(contents, logger):
...
+    metadata_json = _read_metadata_file(metadata_path)
+
+    if _CERT_PROVIDER_COMMAND not in metadata_json:
+      raise CertProvisionError("Client certificate provider is not found.")
+
+    command = metadata_json[_CERT_PROVIDER_COMMAND]
+    if (_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION not in command):
+      command.append(_CERT_PROVIDER_COMMAND_PASSPHRASE_OPTION)
```
NickGoog, line 159: Wouldn't this be appending to a string, based on http://go/aip/auth/4114 ?

You can only append to lists in Python. Strings use `+` for concatenation.

gslib/context_config.py
```diff
@@ -168,7 +215,7 @@ def _ProvisionClientCert(self, command_string, cert_path):
       with open(cert_path, 'w+') as f:
         f.write(sections['CERTIFICATE'])
         f.write(sections['ENCRYPTED PRIVATE KEY'])
-      self.client_cert_password = sections['PASSPHRASE'].splitlines()[1]
+        self.client_cert_password = sections['PASSPHRASE'].splitlines()[1]
```
NickGoog, line 218: Why indent this?

NickGoog:
This will probably require unit test changes in https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/tests/test_context_config.py

But we can hold off on implementing those until the main body of logic is finalized.

gslib/context_config.py
shinfan: The '--with-passphrase' in this case for security purposes.

https://github.com/googleapis/google-auth-library-python/blob/e6976879b392508c022610ab3ea2ea55c7089c63/google/auth/transport/_mtls_helper.py#L176

shinfan:
> This will require unit test changes in https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/tests/test_context_config.py
> 
> But we can hold off on implementing those until the main body of logic is finalized.

I have updated the tests as well.

gslib/context_config.py
```diff
@@ -107,6 +113,62 @@ def _SplitPemIntoSections(contents, logger):
...
+  Returns:
+      str: The default command.
+
+  Raises:
+      CertProvisionError: If command cannot be found.
+  """
+  metadata_path = _check_path()
+
+  if not metadata_path:
+    raise CertProvisionError("Client certificate provider is not found.")
```
NickGoog, line 159: I would write this as "Client certificate provider file not found."

Then, you can write the next `CertProvisionError` as "Client certificate provider  command not found."

gslib/context_config.py
NickGoog: Please explain this better.

First, I don't understand how you're using `append` between two strings. That should error:
```
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 'abc'.append('123')
AttributeError: 'str' object has no attribute 'append'
```

Second, based on this line of the auth/4114, it looks like `cert_provider_command` is any user inputted command:
```
"cert_provider_command": "[absolute_path_to_provider_command] --fetch_client_cert"
```

How do we know the command will support "--with-passphrase"? The proposal doesn't require it: http://go/aip/auth/4114

gslib/tests/test_context_config.py
```diff
@@ -31,6 +31,10 @@
 from gslib.tests.util import SetBotoConfigForTest
 from gslib.tests.util import unittest
 
+DEFAULT_METADATA = """
```
NickGoog, line 34: Should this be named something like `DEFAULT_CERT_PROVIDER_FILE_CONTENTS`?

gslib/tests/test_context_config.py
```diff
@@ -230,53 +258,6 @@ def testConvertsAndLogsProvisoningKeyError(self, mock_Popen):
           " Invalid output format from certificate provider, no " +
           unicode_escaped_error_string)
 
-  @mock.patch.object(subprocess, 'Popen')
```
NickGoog, line 233: Why were these three tests deleted?

gslib/tests/test_context_config.py
```diff
@@ -179,12 +183,36 @@ def testDoesNotProvisionIfUseClientCertificateNotTrue(self, mock_Popen):
...
+        context_config.create_context_config(self.mock_logger)
+
+        mock_open.assert_called_once_with(
+            '~/.secureConnect/context_aware_metadata.json')
+        mock_Popen.assert_called_once_with(os.path.realpath('some/helper'),
+                                           '--print_certificate',
+                                           '--with_passphrase')
+
+  @mock.patch.object(subprocess, 'Popen')
+  def testCustomCommand(self, mock_Popen):
```
NickGoog, line 203: The test naming format is "testSomethingDoesSomething".
For example, "testCustomCommandGeneratesCorrectCert".

NickGoog:
Please add tests that trigger the different `CertProvisionError`s and a test for the "space in path to command" issue you mentioned (e.g. "~/folder with spaces/cert_provider_command.sh").

gslib/tests/test_context_config.py
shinfan: The password flag is no longer related to the custom command scenario so these tests are no longer needed.

gslib/context_config.py
shinfan: 1. The field 'cert_provider_command' is a list of strings, so I think `json.load(..)` returns a list in this case.
2. The metadata file and its associated command is Google provided so this '--with-passphrase' assumption should be safe to make.

shinfan: I can make a change to the AIP to indicate that `--with-passphrase` should be supported by the command.

NickGoog: 1) Maybe the AIP should also be updated to show a list of strings for "cert_provider_command"? It looks like a single string right now under "Obtaining the Default Device Certificate". The flag "--fetch_client_cert" is also unexplained.

2) Gotcha.

gslib/tests/test_context_config.py
```diff
@@ -179,12 +191,67 @@ def testDoesNotProvisionIfUseClientCertificateNotTrue(self, mock_Popen):
...
+      with self.assertRaises(ValueError):
+        context_config.create_context_config(self.mock_logger)
+
+        mock_open.assert_called_once_with(
+            '~/.secureConnect/context_aware_metadata.json')
+        self.mock_logger.error.assert_called_once_with(
+            'Client certificate provider command not found.')
+
+  @mock.patch.object(subprocess, 'Popen')
+  def testCustomProviderExecutesCommand(self, mock_Popen):
```
NickGoog, line 242: "testExecutesCustomProviderCommandInBotoConfig"

gslib/tests/test_context_config.py
```diff
@@ -179,12 +191,67 @@ def testDoesNotProvisionIfUseClientCertificateNotTrue(self, mock_Popen):
     context_config.create_context_config(self.mock_logger)
     mock_Popen.assert_not_called()
 
-  def testRaisesErrorIfCertProviderCommandAbsent(self):
+  @mock.patch.object(subprocess, 'Popen')
+  @mock.patch(OPEN_TO_PATCH, new_callable=mock.mock_open)
+  def testDefaultProviderExecutesCommand(self, mock_Popen, mock_open):
```
NickGoog, line 196: "testExecutesDefaultProviderCommandFromFile"

This ordering of words seems more accurate. Can you adjust some of the below test names as well?

gslib/tests/test_context_config.py
```diff
@@ -179,12 +191,67 @@ def testDoesNotProvisionIfUseClientCertificateNotTrue(self, mock_Popen):
...
+
+        mock_open.assert_called_once_with(
+            '~/.secureConnect/context_aware_metadata.json')
+        mock_Popen.assert_called_once_with(os.path.realpath('some/cert helper'),
+                                           '--print_certificate',
+                                           '--with_passphrase')
+
+  @mock.patch.object(subprocess, 'Popen')
+  @mock.patch(OPEN_TO_PATCH, new_callable=mock.mock_open)
+  def testDefaultProviderNoCommandError(self, mock_Popen, mock_open):
```
NickGoog, line 228: "testDefaultProviderCommandRaisesNoCommandError"

gslib/tests/test_context_config.py
```diff
@@ -179,12 +191,67 @@ def testDoesNotProvisionIfUseClientCertificateNotTrue(self, mock_Popen):
...
+
+        mock_open.assert_called_once_with(
+            '~/.secureConnect/context_aware_metadata.json')
+        mock_Popen.assert_called_once_with(os.path.realpath('some/helper'),
+                                           '--print_certificate',
+                                           '--with_passphrase')
+
+  @mock.patch.object(subprocess, 'Popen')
+  @mock.patch(OPEN_TO_PATCH, new_callable=mock.mock_open)
+  def testDefaultProviderExecutesCommandWithSpace(self, mock_Popen, mock_open):
```
NickGoog, line 212: "testExecutesProviderCommandWithSpaceFromDefaultFile"

NickGoog:
Please write tests confirming the three `CertProvisionError`s are raised at the right time

thomasmaclean:
I just want to note here, that this will feature have to wait until the following week's (July 27th) Cloud SDK release, as there isn't enough time this week to do the final review and run through the release process before the Cloud SDK cutoff on Friday 9am EST.

thomasmaclean:
@shinfan Any updates? I'd like to include this in this coming week's release.

shinfan:
@thomasmaclean Thanks. PTAL.

shinfan:
@thomasmaclean I am not sure why the kokoro is failing. From the log at https://fusion2.corp.google.com/invocations/731f1f30-ba29-4f68-942d-e66eb6a13896/targets/cloud_storage_gsutil%2Flinux%2Fpy35_json/log

The error is ERROR 0719 14:36:16.872922 context_config.py] Failed to provision client certificate: [Errno 2] No such file or directory: '/bin/cat /tmpfs/src/keystore/74008_mtls_test_cert'

Is it the test cert file not uploaded correctly?

gslib/context_config.py
```diff
@@ -124,37 +186,29 @@ def __init__(self, logger):
...
-        password_arg not in command_string):
-      command_string += password_arg
+    cert_command = config.get('Credentials', 'cert_provider_command', None)
+    if not cert_command:
+      # Load the default certificate provider if sit is not provided by user.
+      cert_command = _default_command()
 
     try:
-      command_process = subprocess.Popen(command_string.split(' '),
+      command_process = subprocess.Popen(cert_command,
```
dilipped, line 211: The kokoro failures are happening because we are no longer splitting the command here. Any reason why this was removed?

dilipped: Oh, I see that you cert command can be a list if it is not defined in the boto config file as per your change. But if it is defined in the config file, it will be a list. 
So you will have t omake sure that cert_command is of the same type, either a string or a list.

after line 205, you can add the split call.


For kokoro test, we are passing a cat <filename> command here https://github.com/GoogleCloudPlatform/gsutil/blob/5317bc2f19c2c6be8e0a502f209f369c39eb9d2f/test/ci/kokoro/config_generator.sh#L49

dilipped (changes requested):

gslib/context_config.py
shinfan: You are right. I put the split(' ') back for the boto config command. Thanks.

dilipped (approved):

dilipped:
I only reviewed the specific change that was making kokoro to fail because of cert command string parsing. Will defer to @NickGoog for final approval as he had left a bunch of comments here.

NickGoog:
Thanks @dilipped,

Waiting on three unit tests to confirm `CertProvisionError` is raised in the three correct places.
Also, didn't we decide to remove `split()`, so that commands with a space in the path would work? (e.g. `/pa th/command.sh`)

thomasmaclean:
@shinfan There was a fairly straightforward merge conflict preventing me from merging in this change which I manually patched, and now there are test failures coming back from that file. I attempted to put the file back to the re-merged state, but you may just want to `git reset 4f33b68fb1037edf9e18183c012805c3ada15a2f --hard` to get back to the pre-merged state and try again yourself. Sorry for any confusion here!

NickGoog (approved):

PR #1303, opened 2021-07-01
Fix raising-bad-type yapf errors.

rrauber:
We were getting a raising-bad-type error in the modified file because we were raising the output of a function with no return value: _TranslateExceptionAndRaise. Technically there's nothing wrong with this since the statement  `raise _TranslateExceptionAndRaise(...)` would never be fully evaluated since _TranslateExceptionAndRaise raises an error, but removing the extraneous `raise`s is cleaner, so I did that instead of disabling the check.


NickGoog:
There was a Kokoro error that wasn't just a timeout, so rerunning

NickGoog (commented):
Great, but got hit by HMAC limit and some timeouts.
Lots of Kokoro failures, so did not check every one after seeing this trend.

PR #1305, opened 2021-07-21
Make NotParallelizable decorator work for classes.

NickGoog:
Make NotParallelizable decorator work for classes and fix uncovered bugs.

### DIFF
--- a/gslib/tests/test_context_config.py
+++ b/gslib/tests/test_context_config.py
@@ -162,7 +162,14 @@ class TestContextConfig(testcase.GsUtilUnitTestCase):
 
   def setUp(self):
     super(TestContextConfig, self).setUp()
+    self._old_context_config = context_config._singleton_config
+    context_config._singleton_config = None
+
     self.mock_logger = mock.Mock()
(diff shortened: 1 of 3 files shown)


gslib/tests/test_context_config.py
```diff
@@ -164,6 +164,10 @@ def setUp(self):
     super(TestContextConfig, self).setUp()
     self.mock_logger = mock.Mock()
 
+  def tearDown(self):
+    super(TestContextConfig, self).tearDown()
+    context_config._singleton_config = None
```
dilipped, line 169: If in future we add a test that relies on the context config getting set, then this might create an issue because it will set it to None. To be future proof, maybe we should store the previous value so that this acts like a mock patch.

in setUp:
self._old_context_config = context_config._singleton_config
context_config._singleton_config = None

In tearDown:
context_config._singleton_config = self._old_context_config

NickGoog: Done

dilipped (approved):

PR #1312, opened 2021-08-19
Make certain utils generic for future projects

NickGoog:
Make temporary file naming utils generic.
Make binary running logic generic.
Fix an issue where unit tests fail if `cert_provider_command` is present in `.boto`.

### DIFF
--- a/gslib/tests/test_temporary_file_util.py
+++ b/gslib/tests/test_temporary_file_util.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


gslib/tests/test_temporary_file_util.py
```diff
@@ -0,0 +1,38 @@
...
+
+from gslib import storage_url
+from gslib.tests import testcase
+from gslib.utils import temporary_file_util
+
+
+class TestTemporaryUtil(testcase.GsUtilUnitTestCase):
+  """Test temporary file utils."""
+
+  def test_gets_temporary_file_name(self):
```
rrauber, line 30: I think gsutil tends to use lower camel case for test names. I don't think it's 100% consistent but may be good to change this to match the other test file.

gslib/utils/execution_util.py
```diff
@@ -0,0 +1,51 @@
...
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import subprocess
+
+
+def execute_external_command(command_and_flags):
```
rrauber, line 25: I think this should be upper camel case

rrauber (changes requested):

gslib/tests/test_temporary_file_util.py
NickGoog: Yes, great catch.

gslib/utils/execution_util.py
NickGoog: Good catch

rrauber (approved):

PR #1314, opened 2021-08-23
Fix AttributeError when running without stdin

wmark:
With a closed stdin, that is without any means to accept input from the user,
the session cannot be interactive. sys.stdin is NoneType in that case, and
NoneType knows no `isatty()`. The error will occur with any command, but
the fastest way to trigger it is:

    gsutil version <&-

### DIFF
--- a/gslib/utils/system_util.py
+++ b/gslib/utils/system_util.py
@@ -244,7 +244,8 @@ def IsRunningInCiEnvironment():
 
 def IsRunningInteractively():
   """Returns True if currently running interactively on a TTY."""
-  return sys.stdout.isatty() and sys.stderr.isatty() and sys.stdin.isatty()
+  return sys.stdout.isatty() and sys.stderr.isatty() and \
+         sys.stdin and sys.stdin.isatty()
 
 
 def MonkeyPatchHttp():


dilipped:
@wmark Thanks for the PR.

Can you please provide some instructions on how to reproduce the issue that you are trying to fix? Or point to an existing issue?

wmark:
Of course, succinct:

```bash
gsutil version <&-
```

The script I’ve used this in consumes *stdin* itself (and avoids any user input), so you find this instance of the above constellation (please note the order of command and *filedesc* does not matter using *bash*):

```bash
<&- gsutil \
  -o GSUtil:parallel_composite_upload_threshold=100M \
    "${set_headers[@]}" \
    cp -s "${storage_class}" -n \
      "${src}" \
      "${bucket}/${D%.*}/${dst}"
```

Expected: `gsutil` executes the command.

Got:
```console
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gsutil", line 21, in <module>
    gsutil.RunMain()
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gsutil.py", line 121, in RunMain
    import gslib.__main__
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gslib/__main__.py", line 79, in <module>
    gslib.USER_AGENT = GetUserAgent(args, metrics.MetricsCollector.IsDisabled())
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gslib/utils/user_agent_helper.py", line 38, in GetUserAgent
    user_agent += ' interactive/%s' % system_util.IsRunningInteractively()
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gslib/utils/system_util.py", line 247, in IsRunningInteractively
    sys.stdin.isatty()
AttributeError: 'NoneType' object has no attribute 'isatty'

```

dilipped:
Thanks for the example. Is there a reason to use the closed stream? Can you use other workaround?

I am a bit hesitant about merging this. The function is meant to check `isatty()` on all the input and output streams. If we return False because `stdin` was closed, it becomes a bit misleading. I think it is correct to expect an error for a closed stream.

dilipped:
Closing this PR for the reason mentioned above. Feel free to re-open or comment if there are any issues.

PR #1316, opened 2021-08-25
STET Util

NickGoog:
STET util for running binary.

Earlier commits include STET .boto config setting and temporary file type.
"[GSUtil]" isn't the best fit for the STET config setting, but everything else is less related, and it's probably not worth making a new section in the file. Briefly chatted with @thomasmaclean  about this.

### DIFF
--- a/gslib/commands/config.py
+++ b/gslib/commands/config.py
@@ -528,6 +528,14 @@ content_language = en
 # If this option is not supplied, those tests will be skipped.
 #test_notification_url = https://yourdomain.url/notification-endpoint
 
+# Used in conjunction with --stet flag on cp command for end-to-end encryption.
+# STET binary path. If not specified, gsutil checks PATH for "stet".
+#stet_binary_path = <Path to binary "/usr/local/bin/stet">
+
+# STET config path. If not specified, gsutil checks the default config location:
+# "~/.config/stet.yaml"
+#stet_config_path = <Path to config file "~/.config/my_config.yaml">
+
 """ % {
     'hash_fast_else_fail': CHECK_HASH_IF_FAST_ELSE_FAIL,
     'hash_fast_else_skip': CHECK_HASH_IF_FAST_ELSE_SKIP,
--- a/gslib/commands/config.py
+++ b/gslib/commands/config.py
@@ -45,6 +45,7 @@ from gslib.exception import CommandException
 from gslib.metrics import CheckAndMaybePromptForAnalyticsEnabling
 from gslib.sig_handling import RegisterSignalHandler
 from gslib.utils import constants
+from gslib.utils import stet_util
 from gslib.utils import system_util
 from gslib.utils.hashing_helper import CHECK_HASH_ALWAYS
 from gslib.utils.hashing_helper import CHECK_HASH_IF_FAST_ELSE_FAIL
@@ -533,7 +534,7 @@ content_language = en
 #stet_binary_path = <Path to binary "/usr/local/bin/stet">
 
 # STET config path. If not specified, gsutil checks the default config location:
-# "~/.config/stet.yaml"
+# "{default_stet_config_path}"
 #stet_config_path = <Path to config file "~/.config/my_config.yaml">
 
 """ % {
@@ -559,6 +560,7 @@ content_language = en
     'max_upload_compression_buffer_size':
         (DEFAULT_MAX_UPLOAD_COMPRESSION_BUFFER_SIZE),
     'gzip_compression_level': DEFAULT_GZIP_COMPRESSION_LEVEL,
+    'default_stet_config_path': stet_util.DEFAULT_STET_CONFIG_PATH,
 }
 
 CONFIG_OAUTH2_CONFIG_CONTENT = """
--- a/gslib/tests/test_stet_util.py
+++ b/gslib/tests/test_stet_util.py
@@ -1,0 +1,199 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for stet_util.py."""
+
+from __future__ import absolute_import
+from __future__ import print_function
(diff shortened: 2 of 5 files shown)


gslib/commands/config.py
```diff
@@ -528,6 +528,13 @@
 # If this option is not supplied, those tests will be skipped.
 #test_notification_url = https://yourdomain.url/notification-endpoint
 
+# Used in conjunction with --stet flag on cp command.
+# Override STET binary path for end-to-end encryption.
+#stet_binary_path = <Path to binary "/usr/local/bin/stet">
+
+# Override STET config path for end-to-end encryption.
+#stet_config_path = <Path to binary "~/.config/stet.yaml">
```
dilipped, line 536: Not a binary

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+DEFAULT_STET_CONFIG_PATH = '~/.config/stet.yaml'
+
+
+class StetSubcommandName(object):
+  """Enum class for available STET subcommands."""
+  ENCRYPT = 'encrypt'
+  DECRYPT = 'decrypt'
+
+
+def _get_stet_binary_from_path():
```
dilipped, line 40: I think this is not needed. Subprocess.Popen should be able to find the correct binary. Just doing
```
subprocess.Popen(['stet', ....]) should be enough.
```

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+import os
+import shutil
+
+from gslib import storage_url
+from gslib.utils import execution_util
+from gslib.utils import temporary_file_util
+
+from boto import config
+
+DEFAULT_STET_CONFIG_PATH = '~/.config/stet.yaml'
```
dilipped, line 31: will the user know about creating the path at this location? Is this a common place for stet yaml? If not then maybe we should just avoid the default and let the user explicitly mention this path in the boto config? I think usually hidden files are written out by some tool. If the stet tool or gsutil is not doing this for the user, we should not expect user to write out a hidden file.

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+  Raises:
+    OSError: STET binary was not present in gsutil's boto config file.
+  """
+  binary_path = config.get('GSUtil', 'stet_binary_path', None)
+  if not binary_path:
+    binary_path = _get_stet_binary_from_path()
+  if not binary_path:
+    raise KeyError('Could not find STET binary in boto config or PATH.')
+
+  config_path = config.get('GSUtil', 'stet_config_path', None)
```
dilipped, line 73: See my command about not using a default and having the user always specify this explicitly. 

However, if you still want to use the default, I think you can just replace the None here with the default path and then you can just do the existence check on `config_path`.

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+  flags for the STET binary.
+
+  Args:
+    subcommand (StetSubcommandName): Subcommand to call on STET binary.
+    blob_id (str): Cloud URL that binary uses for validation.
+    in_file_path (str): File to be transformed source.
+    out_file_path (str): Where to write result of transform.
+    logger (logging.Logger): For logging STET binary output.
+
+  Returns:
```
dilipped, line 61: No longer valid

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+
+def decrypt_download(source_url, destination_url, logger):
+  """STET-decrypts downloaded file.
+
+  Args:
+    source_url (StorageUrl): Copy source.
+    destination_url (StorageUrl): Copy destination.
+    logger (logging.Logger): For logging STET binary output.
+
+  Returns:
```
dilipped, line 118: Does not return anything

gslib/utils/stet_util.py
```diff
@@ -0,0 +1,128 @@
...
+    stet_temporary_file_url (StorageUrl): Path to STET-encrypted file.
+  """
+  in_file = source_url.object_name
+  out_file = temporary_file_util.GetStetTempFileName(source_url)
+  blob_id = destination_url.url_string
+
+  _stet_transform(StetSubcommandName.ENCRYPT, blob_id, in_file, out_file,
+                  logger)
+  return storage_url.StorageUrlFromString(out_file)
```
dilipped, line 107: Do we need to wrap this in storage_url? My guess is that the caller will be only interested in the out_file path?

gslib/utils/stet_util.py
NickGoog: Yes, but because we catch the missing binary, we can give a clearer error.
Otherwise, you get something similar to a `cp` upload source file missing.
```
>>> import subprocess
>>> subprocess.Popen('missing_stet')
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1823, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'missing_stet'
```

gslib/utils/stet_util.py
NickGoog: This is the default path the STET team gave me. Maybe their binary or the install instructions create it.

gslib/utils/stet_util.py
NickGoog: It's more consistent with `copy_helper` to work with StorageUrl.
I'd rather do the conversion here than in `copy_helper`, which is super cluttered.
Also, "StorageUrl in, StorageUrl out" makes some intuitive sense.

gslib/utils/stet_util.py
NickGoog: Using the default if the user doesn't specify is what the STET team and I agreed on Friday. Also held a poll with our team.

dilipped: I don't see a mention of default config on the poll. But if the STET team believes that their binary will create this file, then it should be fine.

NickGoog: Oh my bad

gslib/tests/test_stet_util.py
```diff
@@ -0,0 +1,210 @@
...
+    source_url = storage_url.StorageUrlFromString('in')
+    destination_url = storage_url.StorageUrlFromString('gs://bucket/obj')
+    with util.SetBotoConfigForTest([
+        ('GSUtil', 'stet_binary_path', None),
+        ('GSUtil', 'stet_config_path', fake_config_path),
+    ]):
+      with self.assertRaises(KeyError):
+        stet_util.encrypt_upload(source_url, destination_url, None)
+
+  def test_stet_util_errors_if_no_binary(self):
```
dilipped, line 171: test name repeated

dilipped:
Looks good overall. Please ensure that the XML API failures are because of the known XML API issues.

PR #1318, opened 2021-08-31
STET flag and cp integration

NickGoog:
Integration of STET binary with gsutil's cp command.

### DIFF
--- a/gslib/commands/config.py
+++ b/gslib/commands/config.py
@@ -45,7 +45,6 @@ from gslib.exception import CommandException
 from gslib.metrics import CheckAndMaybePromptForAnalyticsEnabling
 from gslib.sig_handling import RegisterSignalHandler
 from gslib.utils import constants
-from gslib.utils import stet_util
 from gslib.utils import system_util
 from gslib.utils.hashing_helper import CHECK_HASH_ALWAYS
 from gslib.utils.hashing_helper import CHECK_HASH_IF_FAST_ELSE_FAIL
@@ -533,9 +532,9 @@ content_language = en
 # STET binary path. If not specified, gsutil checks PATH for "stet".
 #stet_binary_path = <Path to binary "/usr/local/bin/stet">
 
-# STET config path. If not specified, gsutil checks the default config location:
-# "{default_stet_config_path}"
-#stet_config_path = <Path to config file "~/.config/my_config.yaml">
+# STET config path. If not specified, the STET binary will run with its default
+# settings.
+#stet_config_path = ~/.config/my_config.yaml
 
 """ % {
     'hash_fast_else_fail': CHECK_HASH_IF_FAST_ELSE_FAIL,
@@ -560,7 +559,6 @@ content_language = en
     'max_upload_compression_buffer_size':
         (DEFAULT_MAX_UPLOAD_COMPRESSION_BUFFER_SIZE),
     'gzip_compression_level': DEFAULT_GZIP_COMPRESSION_LEVEL,
-    'default_stet_config_path': stet_util.DEFAULT_STET_CONFIG_PATH,
 }
 
 CONFIG_OAUTH2_CONFIG_CONTENT = """
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -680,9 +680,8 @@ _OPTIONS_TEXT = """
                  as binary data, using this option may result in files taking up
                  more space in the cloud than they would if left uncompressed.
 
-  --stet         If the STET binary can be found in boto or PATH, and STET
-                 config file can be found in boto or default location, cp will
-                 use STET for end-to-end encryption.
+  --stet         If the STET binary can be found in boto or PATH, cp will
+                 use the split-trust encryption tool for end-to-end encryption.
 """
 
 _DETAILED_HELP_TEXT = '\n\n'.join([
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -316,7 +316,7 @@ _RETRY_HANDLING_TEXT = """
   the ``cp`` command skips that object and moves on. If any failures were not
   successfully retried by the end of the copy run, the ``cp`` command reports the
   number of failures, and exits with a non-zero status.
-  
+
   For details about gsutil's overall retry handling, see `Retry strategy
   <https://cloud.google.com/storage/docs/retry-strategy#tools>`_.
 """
@@ -679,6 +679,10 @@ _OPTIONS_TEXT = """
                  CAUTION: If some of the source files don't compress well, such
                  as binary data, using this option may result in files taking up
                  more space in the cloud than they would if left uncompressed.
+
+  --stet         If the STET binary can be found in boto or PATH, and STET
+                 config file can be found in boto or default location, cp will
+                 use the split-trust encryption tool for end-to-end encryption.
 """
 
 _DETAILED_HELP_TEXT = '\n\n'.join([
@@ -754,7 +758,9 @@ class CpCommand(Command):
       urls_start_arg=0,
       gs_api_support=[ApiSelector.XML, ApiSelector.JSON],
       gs_default_api=ApiSelector.JSON,
-      supported_private_args=['testcallbackfile='],
+      # Unfortunately, "private" args are the only way to support non-single
+      # character flags.
+      supported_private_args=['stet', 'testcallbackfile='],
       argparse_arguments=[
           CommandArgument.MakeZeroOrMoreCloudOrFileURLsArgument(),
       ],
@@ -919,7 +925,8 @@ class CpCommand(Command):
           manifest=self.manifest,
           gzip_encoded=self.gzip_encoded,
           gzip_exts=self.gzip_exts,
-          preserve_posix=preserve_posix)
+          preserve_posix=preserve_posix,
+          use_stet=self.use_stet)
       if copy_helper_opts.use_manifest:
         if md5:
           self.manifest.Set(exp_src_url.url_string, 'md5', md5)
@@ -1192,6 +1199,7 @@ class CpCommand(Command):
 
     test_callback_file = None
     dest_storage_class = None
+    self.use_stet = False
 
     # self.recursion_requested initialized in command.py (so can be checked
     # in parent class for all commands).
@@ -1251,6 +1259,8 @@ class CpCommand(Command):
         elif o == '-Z':
           gzip_local = True
           gzip_arg_all = GZIP_ALL_FILES
+        elif o == '--stet':
+          self.use_stet = True
 
     if preserve_acl and canned_acl:
       raise CommandException(
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -682,7 +682,7 @@ _OPTIONS_TEXT = """
 
   --stet         If the STET binary can be found in boto or PATH, and STET
                  config file can be found in boto or default location, cp will
-                 use the split-trust encryption tool for end-to-end encryption.
+                 use STET for end-to-end encryption.
 """
 
 _DETAILED_HELP_TEXT = '\n\n'.join([
@@ -1037,7 +1037,8 @@ class CpCommand(Command):
                   copy_helper_opts.skip_unsupported_objects,
                   copy_helper_opts.preserve_acl,
                   preserve_posix=self.preserve_posix_attrs,
-                  delete_source=copy_helper_opts.perform_mv)),
+                  delete_source=copy_helper_opts.perform_mv,
+                  file_size_will_change=self.use_stet)),
           DestinationInfo(exp_dst_url, have_existing_dst_container))
 
       self.has_file_dst = self.has_file_dst or exp_dst_url.IsFileUrl()
@@ -1081,7 +1082,6 @@ class CpCommand(Command):
     self.has_file_dst = False
     self.has_cloud_dst = False
     self.provider_types = set()
-
     # Because cp may have multiple source URLs and multiple destinations, we
     # wrap the name expansion iterator in order to collect analytics.
     name_expansion_iterator = CopyObjectsIterator(
@@ -1102,7 +1102,8 @@ class CpCommand(Command):
           self.recursion_requested or copy_helper_opts.perform_mv,
           all_versions=self.all_versions,
           project_id=self.project_id,
-          ignore_symlinks=self.exclude_symlinks)
+          ignore_symlinks=self.exclude_symlinks,
+          file_size_will_change=self.use_stet)
 
     # Use a lock to ensure accurate statistics in the face of
     # multi-threading/multi-processing.
--- a/gslib/tests/test_stet_cp.py
+++ b/gslib/tests/test_stet_cp.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Integration tests for cp STET binary integration.
+
+Could go with cp tests but that file is bulky."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os
+import stat
+
+from gslib import storage_url
+from gslib.tests import testcase
+from gslib.tests import util
+from gslib.tests.util import ObjectToURI as suri
+from gslib.tests.util import unittest
+from gslib.utils import system_util
+from gslib.utils import temporary_file_util
+
+FAKE_STET_BINARY = """\
+#!/bin/sh
+if [ "$#" -ne 5 ]; then
+  echo "Expected 5 arguments. Got $#."
+  exit 1
+fi
+
+echo "subcommand: $1
+config file: $2
+blob id: $3
+in file: $4
+out file: $5" > $5
+"""
+
+
+@unittest.skipIf(not system_util.IS_LINUX, 'STET binary supports only Linux.')
+class TestStetCp(testcase.GsUtilIntegrationTestCase):
+  """Integration tests for cp integration with STET binary."""
+
+  def setUp(self):
+    super(TestStetCp, self).setUp()
+    # Substitute for actual binary.
+    self.stet_binary_path = self.CreateTempFile(contents=FAKE_STET_BINARY)
+    # Make shell script executable in addition to current permissions.
+    current_stat = os.stat(self.stet_binary_path)
+    os.chmod(self.stet_binary_path, current_stat.st_mode | stat.S_IEXEC)
+    # Fake config.
+    self.stet_config_path = self.CreateTempFile()
+
+  def test_encrypts_upload_if_stet_is_enabled(self):
+    object_uri = self.CreateObject()
+    test_file = self.CreateTempFile(contents='will be rewritten')
+
+    stderr = self.RunGsUtil([
+        '-o', 'GSUtil:stet_binary_path={}'.format(self.stet_binary_path), '-o',
+        'GSUtil:stet_config_path={}'.format(
+            self.stet_config_path), 'cp', '--stet', test_file,
+        suri(object_uri)
+    ],
+                            return_stderr=True)
+
+    # Progress indicator should show transformed file size (variable based on
+    # random string generation above). 4.0 B is the pre-transform size.
+    self.assertNotIn('/4.0 B]', stderr)
+
+    stdout = self.RunGsUtil(['cat', suri(object_uri)], return_stdout=True)
+    self.assertIn('subcommand: encrypt', stdout)
+    self.assertIn('config file: --config-file={}'.format(self.stet_config_path),
+                  stdout)
+    self.assertIn('blob id: --blob-id={}'.format(suri(object_uri)), stdout)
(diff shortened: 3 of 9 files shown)


gslib/commands/cp.py
```diff
@@ -679,6 +679,10 @@
                  CAUTION: If some of the source files don't compress well, such
                  as binary data, using this option may result in files taking up
                  more space in the cloud than they would if left uncompressed.
+
+  --stet         If STET binary can be found in boto or PATH, and STET config
```
sethvargo, line 683:
```suggestion
  --stet         If the STET binary can be found in boto or PATH, and STET config
```

gslib/tests/test_stet_cp.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc. All Rights Reserved.
```
sethvargo, line 2: Per go/copyright, this should be `Google LLC`.

gslib/utils/copy_helper.py
```diff
@@ -2196,6 +2199,9 @@ def CallResumableUpload():
                               provider=dst_url.scheme)
       raise
 
+  if using_temporary_file:
+    os.unlink(src_obj_filestream.name)
```
sethvargo, line 2203: It might be nice if we catch this error and check if the error is NOT_EXISTS before re-raising.

sethvargo (approved):

gslib/utils/copy_helper.py
NickGoog: To make sure we're on the same page, in this upload scenario:
```
cp x.txt gs://my-bucket
```
The temporary file at `src_obj_filestream.name` is `x.txt_.stet_tmp`.

Deleting the file isn't related to an error. Are you referring to the possibility of an error during the unlink?

In that case, I think the fact the user added the `--stet` flag combined with the built-in not found error for a STET-extension file is sufficient:
```
>>> os.unlink('x.txt_.stet_tmp')
FileNotFoundError: [Errno 2] No such file or directory: 'x.txt_.stet_tmp'
```

gslib/commands/cp.py
NickGoog: Done

gslib/tests/test_stet_cp.py
NickGoog: Done, but now every copyright header in the repo is going to make me uncomfortable

gslib/utils/copy_helper.py
sethvargo: I'm suggesting that, as an end user, I would have no idea what that error meant, nor would I know what I needed to do to fix it.

In the case where unlinking fails because the destination was already deleted, I think we should swallow that error and return success. In other situations, it would be helpful to add text like "you may need to delete this file manually".

NickGoog: I can think of three scenarios where the temporary file would already be deleted:
1) The STET binary deletes its output file, but does not raise an error. In this case, gsutil continues execution to the `os.unlink`.
2) The user manually deletes the temporary file during a small window of time while cp is running.
3) Same as (2), but an automated program deletes the file.

(1) seems unlikely, and, if (2) occurred, the user would probably realize what happened based on their recent action and the built-in error message. I imagine a system policy blocking the creation of files is more likely than (3), in which case the error would happen for uploads within the STET binary.

***

For the second idea, it might be beneficial to add "you may need to delete x_.stet_tmp manually" to any errors the STET binary raises. However, I'm leaning away from it b/c I think it's clear from the binary raising an error message and the file extension of the temp file (".stet_tmp") what happened. Rerunning the command will also take care of the temp file.

Leaving this open in case other reviewers want to comment.

sethvargo: There are many, many reasons why a temporary file would be deleted without user interaction, especially if the file is in a directory that the OS has flagged as tmp. These could include, but are not limited to, cron jobs, configuration management tools from corp IT, routine system cleanup operations, software like CCleaner/CleanMyMac, unplugging a USB drive...

I'm mostly just advocating for something more actionable to the end-user.

gslib/commands/cp.py
```diff
@@ -679,6 +679,10 @@
                  CAUTION: If some of the source files don't compress well, such
                  as binary data, using this option may result in files taking up
                  more space in the cloud than they would if left uncompressed.
+
+  --stet         If the STET binary can be found in boto or PATH, and STET
+                 config file can be found in boto or default location, cp will
```
KeithMoyer, line 684: The criteria listed here don't seem quite right. If the STET config isn't in the boto, gsutil shouldn't be trying to find any default location; it should call stet without passing any config file (letting stet deal with finding a default config file).

gslib/utils/copy_helper.py
NickGoog: In the latest design, gsutil adds the temporary files to the same directory as the original source or final destination files. Automated tools would have to distinguish between legitimate copy files and temp files within the same directory to cause this error. Gsutil uses other temporary files for resuming downloads, and users currently don't let their automated tools delete these.

Handling every scenario like a cron job or cleaner tool sweeping in to delete a temp file moments after its created would also make the code unwieldy. We would have to check for this in at least these places:
1) After the STET binary completes writing the file and before the upload begins.
2) After a download completes and before the STET binary begins decrypting it.
3) Before the STET binary tries to delete the original encrypted version of a file after decrypting it.

Referring to an earlier comment, in my opinion, the user gets enough hint about what happened from three facts:
1) The command was run with "--stet".
2) The error is related to a file ending in "stet_tmp" not being found.
3a) The user is aware the delete action they just took OR (3b)
3b) The user has knowledge about automatic deletions on their system.

Agreed that (3b) is the weakest point, but gsutil doesn't fall into the category of software that has to take an inventory of everything else happening on a user device and defend against unlikely scenarios.

NickGoog: FYI, I'm OnDuty and inherited a backlog of 20+ bugs, so I don't really have time to discuss this further. If the majority of reviewers want to catch `FileNotFoundError` on delete operations, I'll do it.

gslib/commands/cp.py
NickGoog: Fixed

gslib/commands/cp.py
```diff
@@ -679,6 +679,9 @@
                  CAUTION: If some of the source files don't compress well, such
                  as binary data, using this option may result in files taking up
                  more space in the cloud than they would if left uncompressed.
+
+  --stet         If the STET binary can be found in boto or PATH, cp will
+                 use the split-trust encryption tool for end-to-end encryption.
```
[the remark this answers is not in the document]
NickGoog: Done

gslib/commands/config.py
```diff
@@ -533,8 +532,8 @@
 # STET binary path. If not specified, gsutil checks PATH for "stet".
 #stet_binary_path = <Path to binary "/usr/local/bin/stet">
 
-# STET config path. If not specified, gsutil checks the default config location:
-# "{default_stet_config_path}"
+# STET config path. If not specified, the STET binary will run with its default
+# settings.
 #stet_config_path = <Path to config file "~/.config/my_config.yaml">
```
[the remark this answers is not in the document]
NickGoog: Done

gslib/utils/copy_helper.py
```diff
@@ -3711,11 +3736,12 @@ def PerformCopy(logger,
...
     if src_url.IsStream() or src_url.IsFifo():
       src_obj_size = None
-    elif src_obj_metadata and src_obj_metadata.size:
+    elif src_obj_metadata and src_obj_metadata.size and not use_stet:
       # Iterator retrieved the file's size, no need to stat it again.
+      # Unless STET changed the file size.
       src_obj_size = src_obj_metadata.size
     else:
-      src_obj_size = os.path.getsize(src_url.object_name)
+      src_obj_size = os.path.getsize(source_stream_url.object_name)
```
dilipped, line 3744: Is the behavior verified for rsync? given that rsync does size comparison.

Also, we are not updating the size for downloads, does that have implications with rsync?

gslib/utils/copy_helper.py
```diff
@@ -2196,6 +2199,9 @@ def CallResumableUpload():
                               provider=dst_url.scheme)
       raise
 
+  if using_temporary_file:
```
dilipped, line 2202: Can we move this outside of this method? I think this kinda break the contract as this method is just supposed to upload the file and not delete the source. Maybe the caller can decide to delete the source.

gslib/utils/copy_helper.py
```diff
@@ -3222,6 +3234,9 @@ def _ValidateAndCompleteDownload(logger,
   if file_name != final_file_name:
     # Data is still in a temporary file, so move it to a permanent location.
     os.rename(file_name, final_file_name)
+  if use_stet:
+    # Decrypt data using STET binary.
+    stet_util.decrypt_download(src_url, dst_url, logger)
```
dilipped, line 3239: Can we do this in the caller? I think this breaks the contract a bit given that this method is only meant to download an object given a source and destination. WE shouldn't try to do anything more than that here.

gslib/utils/copy_helper.py
```diff
@@ -3472,13 +3489,14 @@ def GetSourceFieldsNeededForCopy(dst_is_cloud,
...
         'metadata',
         'metageneration',
-        'size',
         'storageClass',
         'timeCreated',
     ])
     # We only need the ACL if we're going to preserve it.
     if preserve_acl:
       src_obj_fields_set.update(['acl'])
+    if not file_size_will_change:
```
dilipped, line 3498: Is there any harm if the size is added even for stet? I was wondering if we can avoid having to check for this case and keep it generic enough.

gslib/utils/copy_helper.py
dilipped: I feel we will have even bigger problems if a cronjob is deleting gsutil's tmp files. What if it deletes the tmp file while the upload is in progress? In fact, if such a cronjob exists, the probability of it deleting an in-progress file is much higher compared to the short window when os.unlink happens. So I feel the protection against os.unlink failing is not really needed here

I agree with @sethvargo  that the message to the user is not that great, but we make certain assumptions here that the user will not mess up with gsutil specific files to keep the code complexity low like Nick has mentioned. (We will have to add these checks at a lot of places if we really want to give the user a good message for situations like these.)

dilipped (commented):
Yet to review the test files. Wanted to send out some early feedback.

gslib/gcs_json_api.py
```diff
@@ -1456,6 +1456,7 @@ def _UploadObject(self,
     Returns:
       Uploaded object metadata.
     """
+    print(object_metadata)
```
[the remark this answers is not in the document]
NickGoog: Good catch

gslib/utils/copy_helper.py
NickGoog: The use_stet flag wouldn't be accessible with rsync, so how would the new code be triggered?

gslib/utils/copy_helper.py
NickGoog: Going to resolve to clean up the code review screen. Feel free to reopen.

gslib/utils/copy_helper.py
NickGoog: It seems more awkward to move the delete to `PerformCopy` to me:
Before:
```
if dst_url.IsCloudUrl():
      return _UploadFileToObject(src_url,
                                 src_obj_filestream,
                                 ...
                                 using_temporary_file=use_stet)
```
After:
```
if dst_url.IsCloudUrl():
      uploaded_metadata = _UploadFileToObject(src_url,
                                 src_obj_filestream,
                                 ...
                                 using_temporary_file=use_stet)
      if use_stet:
        # Delete temporary file.
        os.unlink(src_obj_filestream.name)
```
It breaks a different pattern in `PerformCopy` where we have:
```
if [identify copy type]:
  return [copy method]
```

gslib/utils/copy_helper.py
NickGoog: No, `ValidateAndCompleteDownload` is the best place to perform a transform. We already perform the "ungzip" transform and "temp file rename" transform here.

gslib/utils/copy_helper.py
NickGoog: The changing file size results in things like "Copied 1.2/1.0GiB" if we leave seek-ahead iterator as it is.

gslib/commands/cp.py
```diff
@@ -1086,7 +1092,9 @@ def RunCommand(self):
     seek_ahead_iterator = None
     # Cannot seek ahead with stdin args, since we can only iterate them
     # once without buffering in memory.
-    if not copy_helper_opts.read_args_from_stdin:
+    # Cannot run with STET b/c file sizes will change. Could still
+    # get file count, but it's not worth the overhead at this point.
```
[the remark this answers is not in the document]
NickGoog: I'll try this out, but `copy_helper_opts` isn't available in `name_expansion.py`, so I'll be using a parameter.

gslib/utils/copy_helper.py
dilipped: I think gzip does make sense since it is required for integrity checking.

In case of stet, I think we are already telling `_DownloadObjectToFile` to use a source and a destination and download the file to that destination. I feel whatever outside transformation we need to do, we should be doing that after the download is finished as the DownloadObjectToFile has done the task of successfully downloading the file. Also we need to pass the `use_stet` flag two levels down which can be avoided.

As mentioned in the upload section, I am not strongly against what you have, so I am fine keeping it if other reviewers agree on this.

gslib/utils/copy_helper.py
dilipped: Ack

gslib/utils/copy_helper.py
dilipped: Ah, I though we are supporting it for rsync as well. Is there an expectation that it should work with rsync in the future? maybe we should confirm that requirement from the stet team.

gslib/utils/copy_helper.py
dilipped: Hmm, I would prefer that if it means keeping the job of private methods limited to what they are meant to do. Plus you also avoid passing the`use_stet` flag.

Regarding the pattern in performCopy, I think breaking a pattern is better than breaking a contract.

Having said that, I am not strongly opposed to having it in UploadFileToObject so I am fine if other reviewers are okay with it.

gslib/utils/copy_helper.py
dilipped: BTW, what's the dst_url here? Is it the one with .gstmp extension? Are we renaming the tmp file to final_file_name already on line 3239?

gslib/utils/copy_helper.py
NickGoog: I don't think that makes sense fundamentally. To run rsync's file comparison algorithms, we would have to encrypt all the local files before comparing to the cloud. So we would have to operate on all the files before deciding if they should be copied.

gslib/utils/copy_helper.py
NickGoog: The decryption can't happen anywhere else b/c ungzip has to run first.

`final_file_name` is based on `dst_url`.

Yes, we are renaming the temporary file to the final file in the line `os.rename(file_name, final_file_name)`.

I would rather `file_name` be called `temporary_file_name` but that would require some extra refactors.

gslib/utils/copy_helper.py
NickGoog: Is it really a contract? `DownloadObjectToFile` calls `ValidateAndCompleteDownload`, which deletes temporary files

gslib/commands/cp.py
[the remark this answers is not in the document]
NickGoog: Done

gslib/utils/copy_helper.py
```diff
@@ -3205,29 +3216,34 @@ def _ValidateAndCompleteDownload(logger,
...
       else:
-        os.unlink(file_name)
+        os.unlink(unzipped_temporary_file_name)
       raise
 
-  if file_name != final_file_name:
+  if unzipped_temporary_file_name != final_file_name:
     # Data is still in a temporary file, so move it to a permanent location.
-    os.rename(file_name, final_file_name)
+    os.rename(unzipped_temporary_file_name, final_file_name)
```
dilipped, line 3243: Shouldn't the renaming happen after the stet transform?

NickGoog: Oh, yikes, I did all the setup, but forgot to make the actual change.

gslib/tests/test_stet_cp.py
```diff
@@ -0,0 +1,146 @@
...
+                  downloaded_text)
+    self.assertIn('in file: {}'.format(test_file), downloaded_text)
+    self.assertIn('out file: {}_.stet_tmp'.format(test_file), downloaded_text)
+
+    self.assertFalse(
+        os.path.exists(
+            temporary_file_util.GetStetTempFileName(
+                storage_url.StorageUrlFromString(test_file))))
+
+  def test_does_not_seek_ahead_if_stet_transform(self):
```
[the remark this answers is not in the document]
NickGoog: Yes, see the code comments toward the end of the test body

gslib/utils/copy_helper.py
```diff
@@ -3199,29 +3216,36 @@ def _ValidateAndCompleteDownload(logger,
...
-        os.unlink(file_name)
+        os.unlink(unzipped_temporary_file_name)
       raise
 
-  if file_name != final_file_name:
+  if use_stet:
+    # Decrypt data using STET binary.
+    stet_util.decrypt_download(src_url, dst_url, unzipped_temporary_file_name,
+                               logger)
+  if unzipped_temporary_file_name != final_file_name:
```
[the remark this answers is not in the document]
NickGoog: Good catch!

gslib/tests/test_stet_cp.py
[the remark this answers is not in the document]
dilipped: I don't think I understand. The test description says "Tests that seek-ahead iterator does not run if file size will change.", but aren't we still running seek ahead iterator?

gslib/utils/copy_helper.py
NickGoog: I meant `DownloadObjectToFile` b/c it's another example of an "execute copy" function `PerformCopy` calls.

But moved the delete to `PerformCopy` so that it handles temp file creation and deletion.

gslib/tests/test_stet_cp.py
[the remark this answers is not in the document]
NickGoog: That's old, oops. I'll specify bytes.

gslib/tests/test_stet_cp.py
```diff
@@ -0,0 +1,146 @@
...
+          '-m', '-o', 'GSUtil:stet_binary_path={}'.format(
+              self.stet_binary_path), '-o', 'GSUtil:stet_config_path={}'.format(
+                  self.stet_config_path), 'cp', '-r', '--stet', tmpdir,
+          suri(bucket_uri)
+      ],
+                              return_stderr=True)
+      # Pre-encryption total file size that seek-ahead would find.
+      self.assertNotIn('18.0 B]', stderr)
+      # Post-encryption total file size found if seek-ahead is disabled.
+      self.assertIn('2.0 KiB]', stderr)
```
dilipped, line 146: Sry, I still don't understand how this test works. This is showing the updated size, but how does this prove that the seek ahead did not calculate the size? I have a feeling that you will get the same output even if you explicitly set `self.self.count_data_bytes` to True?

Also, full text for the progress callback output might be helpful to read what exactly this represents. something like
```
[3 files][   32.0 B/   32.0 B]
```

gslib/tests/test_stet_cp.py
```diff
@@ -0,0 +1,146 @@
...
+      stderr = self.RunGsUtil([
+          '-m', '-o', 'GSUtil:stet_binary_path={}'.format(
+              self.stet_binary_path), '-o', 'GSUtil:stet_config_path={}'.format(
+                  self.stet_config_path), 'cp', '-r', '--stet', tmpdir,
+          suri(bucket_uri)
+      ],
+                              return_stderr=True)
+      # Pre-encryption total file size that seek-ahead would find.
+      self.assertNotIn('18.0 B]', stderr)
+      # Post-encryption total file size found if seek-ahead is disabled.
```
dilipped, line 145: I think this command needs to be updated? seek-ahead is not technically disabled. You are just disabling the bytes count.

NickGoog: Made it a little more clear, but I think it makes sense given the context

gslib/tests/test_stet_cp.py
NickGoog: Added more comments

dilipped: Talked to Nick offline and he clarified how this is intended to work. Looks good.

dilipped (approved):

dilipped:
CI seems to be failing

gslib/tests/test_stet_cp.py
```diff
@@ -0,0 +1,152 @@
...
+                  downloaded_text)
+    self.assertIn('in file: {}'.format(test_file), downloaded_text)
+    self.assertIn('out file: {}_.stet_tmp'.format(test_file), downloaded_text)
+
+    self.assertFalse(
+        os.path.exists(
+            temporary_file_util.GetStetTempFileName(
+                storage_url.StorageUrlFromString(test_file))))
+
+  def test_does_not_seek_ahead_for_bytes_if_stet_transform(self):
```
dilipped, line 124: kokoro seems to be failing for this

dilipped (changes requested):

gslib/tests/test_stet_cp.py
NickGoog: Pushed +- 0.1 KiB regex, will wait for result

NickGoog: Looks fine

dilipped (approved):

PR #1321, opened 2021-09-07
Improve content type inference for some common extensions.

rrauber:


gslib/tests/test_copy_helper_funcs.py
```diff
@@ -413,6 +413,21 @@ def testSetContentTypeFromFile(self):
       _SetContentTypeFromFile(src_url_stub, dst_obj_metadata_mock)
     self.assertEqual('text/plain', dst_obj_metadata_mock.contentType)
 
+  def testSetsContentTypesForCommonFileExtensionsCorrectly(self):
+    extension_rules = copy_helper.COMMON_EXTENSION_RULES.items()
+    for extension, expected_content_type in extension_rules:
+      dst_obj_metadata_mock = mock.MagicMock(contentType=None)
+      src_url_stub = mock.MagicMock(object_name='file' + extension,
+                                    **{
```
NickGoog, line 421: This is neat, but isn't this simpler to understand syntax?
```
src_url_stub.IsFileUrl.return_value = True
src_url_stub.IsStream...
```

gslib/utils/copy_helper.py
```diff
@@ -288,6 +288,11 @@
              if CheckMultiprocessingAvailableAndInit().is_available else None))
 suggested_sliced_transfers_lock = parallelism_framework_util.CreateLock()
 
+COMMON_EXTENSION_RULES = {
+    '.md': 'text/markdown',
```
NickGoog, line 292: Is there a bug related to markdown?

gslib/utils/copy_helper.py
```diff
@@ -1612,7 +1617,12 @@ def _SetContentTypeFromFile(src_url, dst_obj_metadata):
               'Encountered OSError running "file -b --mime %s"\n%s' %
               (real_file_path, e))
       else:
-        content_type = mimetypes.guess_type(real_file_path)[0]
+        for extension, extension_content_type in COMMON_EXTENSION_RULES.items():
```
NickGoog, line 1620: This loop feels like it should be a set or dict look-up.
Can you do something like:
```
split_file_path = real_file_path.split('.')
if len(split_file_path) > 1:
  extension = split_file_path[-1]
  if extension in COMMON_EXTENSION_RULES:
    ...
```

That might break if there are double-extensions like `.txt.tar`, but it should work with all the currently supported rules.

gslib/utils/copy_helper.py
```diff
@@ -1612,7 +1617,12 @@ def _SetContentTypeFromFile(src_url, dst_obj_metadata):
...
+          if real_file_path.endswith(extension):
+            content_type = extension_content_type
+            break
+        if not content_type:
+          content_type = mimetypes.guess_type(real_file_path)[0]
```
NickGoog, line 1625: The zero index is a magic number.
Maybe add a code comment giving insight into what `guess_type` returns or destructure to give insight into how many return values there are:
```
content_type, _, _ = mimetypes.guess_type(real_file_path)
```

gslib/tests/test_copy_helper_funcs.py
rrauber: Sure!

gslib/utils/copy_helper.py
rrauber: yeah, adding double-extension rules was exactly what I was worried about. Made the change though.

gslib/utils/copy_helper.py
rrauber: There is, I'll send it to you somewhere else.

gslib/utils/copy_helper.py
rrauber: This came directly from the gsutil implementation, but now's a good time to improve it for sure!

gslib/utils/copy_helper.py
```diff
@@ -1612,7 +1617,11 @@ def _SetContentTypeFromFile(src_url, dst_obj_metadata):
               'Encountered OSError running "file -b --mime %s"\n%s' %
               (real_file_path, e))
       else:
-        content_type = mimetypes.guess_type(real_file_path)[0]
+        _, _, extension = real_file_path.rpartition('.')
```
NickGoog, line 1620: Nice use of rpartition!

NickGoog (approved):

gslib/utils/copy_helper.py
dilipped: BTW, https://docs.python.org/3/library/os.path.html#os.path.splitext is commonly used to get the extension.

dilipped (approved):

PR #1322, opened 2021-09-07
Adds a message encouraging py3 upgrade

thomasmaclean:
This PR does the following:
- Alerts users running the `upgrade` or `version` command that they should install Py3 to stay up to date with the latest code. This way if you can't use the `upgrade` command, there will still be messaging.
- Python 2 gets a different upgrade tarball, leaving the door open for releasing bug fixes later on.

### DIFF
--- a/gslib/command_runner.py
+++ b/gslib/command_runner.py
@@ -341,7 +341,8 @@ class CommandRunner(object):
       if system_util.IsRunningInteractively() and collect_analytics:
         metrics.CheckAndMaybePromptForAnalyticsEnabling()
 
-    if not do_shutdown:
+    # This will skip the Python update prompt for tests.
+    if do_shutdown:
       self.MaybePromptForPythonUpdate(command_name)
 
     if not args:
--- a/gslib/command_runner.py
+++ b/gslib/command_runner.py
@@ -341,6 +341,8 @@ class CommandRunner(object):
       if system_util.IsRunningInteractively() and collect_analytics:
         metrics.CheckAndMaybePromptForAnalyticsEnabling()
 
+    self.MaybePromptForPythonUpdate(command_name)
+
     if not args:
       args = []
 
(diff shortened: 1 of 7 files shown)


gslib/command_runner.py
```diff
@@ -424,6 +428,58 @@ def RunNamedCommand(self,
...
+    """
+    logger = logging.getLogger()
+    if (self.SkipUpdateCheck(command_name) or
+        boto.config.getbool('GSUtil', 'skip_python_update_prompt', False) or
+        sys.version_info.major != 2):
+      return False
+
+    # Notify the user about Python 2 deprecation.
+    print_to_fd(
+        'Gsutil 5 will drop Python 2 support. Please install Python 3 to '
```
starsandskies, line 479: Recommend: "will drop" --> "drops"

gslib/command_runner.py
```diff
@@ -424,6 +428,58 @@ def RunNamedCommand(self,
...
+        'continue using the latest version of Gsutil. https://goo.gle/py3\n')
```
starsandskies, line 480: Lowercase "g"?

starsandskies, line 480: To double check: This \n will render as a new line in the actual message output?

starsandskies:
I made a few nitty comments, but LGTM overall

gslib/command_runner.py
thomasmaclean: Yes, it's to visually separate this message from the other command output.

gslib/command_runner.py
```diff
@@ -424,6 +426,48 @@ def RunNamedCommand(self,
           )))
     return return_code
 
+  def SkipUpdateCheck(self, command_name):
```
dilipped, line 429: command_name is not used.

gslib/command_runner.py
```diff
@@ -424,6 +426,48 @@ def RunNamedCommand(self,
...
+    if (self.SkipUpdateCheck(command_name) or
+        command_name not in ('update', 'ver', 'version') or
+        boto.config.getbool('GSUtil', 'skip_python_update_prompt', False) or
+        sys.version_info.major != 2):
+      return False
+
+    # Notify the user about Python 2 deprecation.
+    print_to_fd(
+        'Gsutil 5 drops Python 2 support. Please install Python 3 to update '
+        'to the latest version of gsutil. https://goo.gle/py3\n')
```
dilipped, line 468: Should we add a message here about how they can turn off this warning?

gslib/command_runner.py
```diff
@@ -424,6 +426,48 @@ def RunNamedCommand(self,
...
+  def MaybePromptForPythonUpdate(self, command_name):
+    """Alert the user that they should install Python 3.
+
+    Args:
+      command_name: The name of the command being run.
+
+    Returns:
+      True if a prompt was output.
+    """
+    logger = logging.getLogger()
```
dilipped, line 458: Not used.

gslib/command_runner.py
```diff
@@ -424,6 +426,48 @@ def RunNamedCommand(self,
...
+    if (self.SkipUpdateCheck(command_name) or
+        command_name not in ('update', 'ver', 'version') or
+        boto.config.getbool('GSUtil', 'skip_python_update_prompt', False) or
+        sys.version_info.major != 2):
```
dilipped, line 462: Nit, `sys.version_info.major != 2`  can be the first check in the sequence as we don't have to bother about checking the rest if it is python3.

dilipped (approved):

gslib/command_runner.py
thomasmaclean: Knowing this is only used on `version` and `update` now i don't think the config option is worth documenting, although i still think it's good to have.

dilipped: Ah right, I totally missed that this is only for version and update command. In that case, I agree, that it is probably not worth documenting it.

PR #1325, opened 2021-09-13
Disable STET feature temporarily

dilipped:
Disabling STET features temporarily. This PR should be reverted for gsutil 5.x

### DIFF
--- a/gslib/tests/test_stet_cp.py
+++ b/gslib/tests/test_stet_cp.py
@@ -48,8 +48,7 @@ out file: $5" > $5
 
 
 @unittest.skipIf(not system_util.IS_LINUX, 'STET binary supports only Linux.')
-# TODO: Unskip by removing the line below once stet is enabled.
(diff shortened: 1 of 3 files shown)


gslib/tests/test_stet_cp.py
```diff
@@ -48,6 +48,8 @@
 
 
 @unittest.skipIf(not system_util.IS_LINUX, 'STET binary supports only Linux.')
+# TODO: Unskip by removing the line below once stet is enabled.
+@unittest.SkipTest
```
NickGoog, line 52: Isn't the syntax something like `SkipTest("Don't support X")`

NickGoog (approved):

gslib/tests/test_stet_cp.py
dilipped: It is `skip('reason')`. Updated it.

NickGoog (approved):

PR #1327, opened 2021-09-14
Update cp.py

cbonnie:
Adding example on using the -n flag for no clobber.


starsandskies:
LGTM, and for context Bonnie is a TW for GCS, but want to make sure we get signoff from one of the gsutil engineers.

NickGoog (approved):

PR #1328, opened 2021-09-14
Add better-formatted error handling for external commands.

NickGoog:


gslib/utils/execution_util.py
```diff
@@ -46,6 +46,10 @@ def ExecuteExternalCommand(command_and_flags):
     command_stderr = command_stderr.decode()
 
   if command_process.returncode != 0:
-    raise OSError(command_stderr)
+    # Remove period and any trailing newlines or whitespace.
+    # This is because __main__.py formats OSErrors with 'OSError: %s.'
+    error_text = command_stderr.rstrip().rstrip('.')
+    # First argument is 'errno', which is unknown.
+    raise OSError(None, error_text)
```
dilipped, line 53: why not `OSError(error_text)` if we are not going to add any Errno?

NickGoog: That's similar to the original code. The second position arg actually sets the error's text. The first one is the errno.

dilipped: Can you add some context? I think I am not able to understand how adding None helps here. The error message seems better without None.

```
>>> raise OSError('some error')
OSError: some error
>>> raise OSError(None, 'some error')
OSError: [Errno None] some error
```

NickGoog: https://docs.python.org/3/library/exceptions.html#OSError

https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/__main__.py#L670

Note the "e.strerror"

dilipped: Got it. That's exactly the context I was looking for. It wasn't clear to me earlier that `main` is reading a specific field from the error object instead of just stringifying it. Thanks.

I think it might be useful to have a custom error raised instead which indicates something was wrong with the external binary. Something like `ExternalBinaryError`. That way user can have a clear distinction that the custom binary that **they** have setup is faulty rather than something wrong with the gsutil code.

NickGoog: Done

dilipped (approved):

dilipped:
Please wait for kokoro to finish before merging

PR #1333, opened 2021-09-17
Use custom endpoints for resumable uploads on the XML API.

rrauber:


gslib/commands/rsync.py
```diff
@@ -942,7 +942,7 @@ class _DiffIterator(object):
   """Iterator yielding sequence of RsyncDiffToApply objects."""
 
   def __init__(self, command_obj, base_src_url, base_dst_url):
-    global _tmp_files
+    global _tmp_files  # pylint: disable=global-variable-not-assigned
```
dilipped, line 945: I am removing these in my PR. github.com/GoogleCloudPlatform/gsutil/pull/1332

I think these are safe to remove.

NickGoog (approved):
Nice!

gslib/commands/rsync.py
rrauber: Done!

PR #1347, opened 2021-09-27
Backfill cl/399213131

starsandskies:
Clarify that project IDs or project numbers are accepted identifiers for a project

### DIFF
--- a/gslib/addlhelp/command_opts.py
+++ b/gslib/addlhelp/command_opts.py
@@ -149,7 +149,7 @@ _DETAILED_HELP_TEXT = ("""
               running gsutil from a cron job that logs its output to a file, for
               which the only information desired in the log is failures.
 
-  -u          Allows you to specify the ID or number of a user project to be\
+  -u          Allows you to specify the ID or number of a user project to be
               billed for the request. For example:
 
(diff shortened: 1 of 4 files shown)


gslib/addlhelp/command_opts.py
```diff
@@ -149,8 +149,8 @@
               running gsutil from a cron job that logs its output to a file, for
               which the only information desired in the log is failures.
 
-  -u          Allows you to specify a user project to be billed for the request.
-              For example:
+  -u          Allows you to specify the ID or number of a user project to be\
```
NickGoog, line 152: I think this is the only change line that says "user project" instead of just "project"

NickGoog (approved):

gslib/addlhelp/command_opts.py
starsandskies: My understanding is the project used with this flag is a little different than how projects are normally thought of: it's not the project in which the resource resides, but instead a project the requester ("user") is affiliated with. It's probably fine to ax "user", but I also see the use in being explicit about the nature of this particular project. I figured I'd punt and leave that particular text as is for the time being :P

PR #1350, opened 2021-09-28
Allow the removal of project convenience groups.

thomasmaclean:
Currently we don't allow assigning roles to these groups, but we should be able to remove roles from these groups. You can, currently, with `iam set` but not with `iam ch`.

### DIFF
--- a/gslib/utils/iam_helper.py
+++ b/gslib/utils/iam_helper.py
@@ -219,37 +219,47 @@ def BindingStringToTuple(is_grant, input_str):
 
   # Allows user specified PUBLIC_MEMBERS and TYPES to be case insensitive.
   tokens = input_str.split(":")
+
+  discouraged_types = {s.lower(): s for s in DISCOURAGED_TYPES}
   public_members = {s.lower(): s for s in PUBLIC_MEMBERS}
   types = {s.lower(): s for s in TYPES}
+
   possible_public_member_or_type = tokens[0].lower()
   possible_type = '%s:%s' % (tokens[0].lower(), tokens[1].lower())
+
   if possible_public_member_or_type in public_members:
     tokens[0] = public_members[possible_public_member_or_type]
   elif possible_public_member_or_type in types:
     tokens[0] = types[possible_public_member_or_type]
   elif possible_type in types:
     (tokens[0], tokens[1]) = types[possible_type].split(':')
+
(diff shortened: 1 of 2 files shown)


gslib/utils/iam_helper.py
```diff
@@ -219,37 +219,47 @@ def BindingStringToTuple(is_grant, input_str):
...
     else:
       raise CommandException('Incorrect public member type for binding %s' %
                              input_str)
   elif input_str.count(':') == 2:
-    tokens = input_str.split(':')
-    if '%s:%s' % (tokens[0], tokens[1]) in TYPES:
+    if (tokens[0] == 'deleted' and tokens[1] in DISCOURAGED_TYPES):
+      # case "deleted:projectOwner:admin"
+      member = 'deleted:%s' % tokens[1]
+      roles = tokens[2]
```
dilipped, line 259: Should this be DROP_ALL?

gslib/utils/iam_helper.py
```diff
@@ -219,37 +219,47 @@ def BindingStringToTuple(is_grant, input_str):
...
       roles = DROP_ALL
     else:
       raise CommandException('Incorrect public member type for binding %s' %
                              input_str)
   elif input_str.count(':') == 2:
-    tokens = input_str.split(':')
-    if '%s:%s' % (tokens[0], tokens[1]) in TYPES:
+    if (tokens[0] == 'deleted' and tokens[1] in DISCOURAGED_TYPES):
+      # case "deleted:projectOwner:admin"
+      member = 'deleted:%s' % tokens[1]
```
dilipped, line 258: Should you be adding tokens[2] here as well? Looks like the elif condition below assing the entire enput_str as the member.

gslib/utils/iam_helper.py
thomasmaclean: This would allow specifying a role to drop (the previous if block I added handles the DROP_ALL case).

gslib/utils/iam_helper.py
thomasmaclean: It's a bit confusing, but this is actually correct. Deleting a user uses 3 tokens (deleted:user:foo@bar.com) and deleting one of the convenience types only takes 2 (deleted:projectOwner).

dilipped (approved):

PR #1351, opened 2021-09-28
Enable rpo

dilipped:


gslib/commands/mb.py
```diff
@@ -145,7 +147,13 @@
                          "enforced", objects in this bucket cannot be made
                          publicly accessible. Default is "unspecified".
 
-""")
+  --rpo setting          Specifies the `replication setting <https://cloud.google.com/storage/docs/turbo-replication>`_.
+                         Valid values are {rpo_values}. If unspecified,
```
starsandskies, line 151: My understanding is it's a bit more complicated, which we should probably convey here.

Suggestion: Specifies the `replication setting <https://cloud.google.com/storage/docs/turbo-replication>`_.
This flag is not valid for single-region buckets, and multi-region buckets only accept a value of DEFAULT. Valid values for dual region buckets are {rpo_values}. If unspecified, DEFAULT is applied for dual-region and multi-region buckets.

gslib/commands/rpo.py
```diff
@@ -0,0 +1,213 @@
...
+
+VALID_RPO_VALUES = ('ASYNC_TURBO', 'DEFAULT')
+VALID_RPO_VALUES_STRING = '({})'.format('|'.join(VALID_RPO_VALUES))
+
+_SET_SYNOPSIS = """
+  gsutil rpo set {} gs://<bucket_name>...
+""".format(VALID_RPO_VALUES_STRING)
+
+_GET_SYNOPSIS = """
+  gsutil rpo get bucket_url...
```
starsandskies, line 40: This and the _SET_SYNOPSIS appear inconsistent with regards to the bucket name/URL?

gslib/commands/mb.py
dilipped: Done.

gslib/commands/rpo.py
dilipped: Done.

gslib/commands/mb.py
```diff
@@ -147,11 +147,12 @@
                          "enforced", objects in this bucket cannot be made
                          publicly accessible. Default is "unspecified".
 
-  --rpo setting          Specifies the `replication setting <https://cloud.google.com/storage/docs/turbo-replication>`_.
-                         Valid values are {rpo_values}. If unspecified,
-                         the DEFAULT setting is applied for dual-region and
-                         multi-region buckets and
-                         None for single-region buckets.
+  --rpo setting          Specifies the replication setting <https://cloud.google.com/storage/docs/turbo-replication>_.
```
starsandskies, line 150: I think you'll still need the ` tickmarks for the linking to work?

starsandskies:
LGTM

gslib/commands/mb.py
dilipped: Good catch!

thomasmaclean (approved):

PR #1358, opened 2021-10-04
mb command: raise error if json only flags are used for xml API 

dilipped:


gslib/commands/mb.py
```diff
@@ -238,14 +239,14 @@ def RunCommand(self):
             raise CommandException(
                 'Invalid value for --rpo. Must be one of: {},'
                 ' provided: {}'.format(VALID_RPO_VALUES_STRING, a))
+          json_only_flags.append(o)
```
NickGoog, line 242: Does it make more sense to define all the JSON-only flags in a constant at the top of the file?

dilipped: I had it that way originally. But I changed it to each flag because we run the risk of the constant going out of sync with the actual flag checking.

e.g if there is a typo in the constant itself.
```
JSON_ONLY = ['--roo']
```
and the loop is checking for `--rpo`.

dilipped: The other option is to define a constant for each flag as well.

e.g 
RPO_FLAG = '--rpo'
JSON_ONLY - [RPO_FLAG]
then in the for loop
if o == RPO_FLAG:
   ... do something

NickGoog: Maybe we can make it more clear by changing the variable name `json_only_flags`, which sounds like a constant.

Something like `json_only_flags_in_command`?

NickGoog (approved):

gslib/commands/mb.py
dilipped: Done.

PR #1362, opened 2021-10-05
Add custom dual region

dilipped:


gslib/commands/mb.py
```diff
@@ -39,8 +39,9 @@
 from gslib.utils.text_util import NormalizeStorageClass
 
 _SYNOPSIS = """
-  gsutil mb [-b (on|off)] [-c <class>] [-l <location>] [-p <project>]
+  gsutil mb [-b (on|off)] [-c <class>] [-l <location>] [-p <proj_id>]
```
NickGoog, line 42: Is this accidentally undoing an earlier docs change?

gslib/commands/mb.py
```diff
@@ -247,6 +255,13 @@ def RunCommand(self):
         elif o == '--pap':
           public_access_prevention = a
           json_only_flags_in_command.append(o)
+        elif o == '--placement':
+          placements = a.split(',')
+          if len(placements) != 2:
+            raise CommandException(
+                'Please specify two regions separated by comma.'
```
NickGoog, line 262: I wonder if we need to say there's no white space after the comma. Maybe say, "two regions in CSV format"

gslib/commands/mb.py
dilipped: Good catch!

gslib/commands/mb.py
dilipped: I added "without space" to the message. I was not sure about "CSV format" because I feel someone might confuse that with an actual csv file.

NickGoog (approved):

gslib/commands/mb.py
```diff
@@ -147,6 +148,12 @@
                          "enforced", objects in this bucket cannot be made
                          publicly accessible. Default is "unspecified".
```
rebecca-pete, line 150: custom dual-region

dilipped: Done.

PR #1365, opened 2021-10-06
Revert "Allow the removal of project convenience groups."

thomasmaclean:
Reverts GoogleCloudPlatform/gsutil#1350

### DIFF
--- a/gslib/tests/test_iam.py
+++ b/gslib/tests/test_iam.py
@@ -373,15 +373,6 @@ class TestIamHelpers(testcase.GsUtilUnitTestCase):
     self.assertIn(
         bvle(members=['deleted:user:foo@bar.com?query=param,uid=123?uid=456'],
              role='roles/storage.admin2'), bindings)
-    # Project convenience groups should be removeable.
-    (_, bindings) = bstt(False, 'Deleted:projectOwner')
-    self.assertEquals(len(bindings), 1)
-    self.assertIn(bvle(members=['deleted:projectOwner'], role=''), bindings)
-    (_, bindings) = bstt(False, 'DELETED:projectOwner:admin')
-    self.assertEquals(len(bindings), 1)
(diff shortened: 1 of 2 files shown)


gslib/tests/test_iam.py
```diff
@@ -373,22 +397,26 @@ def test_valid_deleted_member(self):
...
-        bvle(members=['deleted:projectOwner'], role='roles/storage.admin'),
-        bindings)
 
   def test_duplicate_roles(self):
     """Tests that duplicate roles are ignored."""
     (_, bindings) = bstt(True, 'allUsers:a,a')
     self.assertEquals(len(bindings), 1)
     self.assertIn(bvle(members=['allUsers'], role='roles/storage.a'), bindings)
 
+  def test_project_convenience_groups(self):
```
dilipped, line 407: Maybe we can split this test further. It was not immediately clear what exactly were are testing here.

On test can be for making sure convenience role can be added and another to check for delete.

gslib/tests/test_iam.py
```diff
@@ -373,22 +397,26 @@ def test_valid_deleted_member(self):
...
+    (_, bindings) = bstt(False, 'projectViewer:123424:admin')
+    self.assertEquals(len(bindings), 1)
+    self.assertIn(
+        bvle(members=['projectViewer:123424'], role='roles/storage.admin'),
+        bindings)
+    (_, bindings) = bstt(False, 'projectViewer:123424')
+    self.assertEquals(len(bindings), 1)
+    self.assertIn(bvle(members=['projectViewer:123424'], role=''), bindings)
+    with self.assertRaises(CommandException):
+      bstt(True, 'projectViewer:123424:admin')
```
dilipped, line 418: It was not immediately clear that True indicates that we are trying to add a value. maybe leave a comment here about what true means here.

PR #1389, opened 2021-10-27
Updated setup.py

anukaal:
added support for python 3.10


thomasmaclean:
We included the majority of this PR already, but we still have a couple more things we want to check off before we fully support 3.10.

PR #1402, opened 2021-11-17
Translate command and flags to gcloud storage equivalent

dilipped:
This PR takes care of translating the input command and the arguments to their gcloud storage equivalent command.

Part of the series of PRs for creating a shim to trigger gcloud storage commands underneath if users have opted-in.


gslib/utils/shim_util.py
```diff
@@ -0,0 +1,57 @@
...
+      gcloud_command (dict|str): The corresponding name of the command to be
+        called in gcloud. If this command supports sub-commands, then this 
+        field must be a dict of sub-command-name:GcloudStorageMap pairs.
+      flag_map (dict): A dict of str to GcloudStorageFlag. Mapping of gsutil
+        flags to their equivalent gcloud storage flag names.
+      supports_output_translation (bool): Indicates if the corresponding
+        gcloud storage command supports the printing gsutil formatted output.
+    """
+    self.gcloud_command = gcloud_command
+    self.flag_map = flag_map
```
NickGoog, line 56: Could this be a constant in this file?

gslib/command.py
```diff
@@ -725,27 +729,47 @@ def RaiseInvalidArgumentException(self):
...
+                   update_sub_opts_and_args=True):
     """Parses sub-opt args.
 
     Args:
       check_args: True to have CheckArguments() called after parsing.
-
-    Populates:
-      (self.sub_opts, self.args) from parsing.
+      args: List of args. If None, self.args will be used.
+      update_sub_opts_and_args: True if self.sub_opts and self.args should
```
NickGoog, line 741: I would rename to `should_update_sub_opts_and_args`

gslib/command.py
```diff
@@ -725,27 +729,47 @@ def RaiseInvalidArgumentException(self):
...
+        be updated with the values returned after parsing. Else return a 
+        tuple of sub_opts, args returned by getopt.getopt. This is done
+        to allow this method to be called from GetGcloudStorageArgs in which
+        case we do not want to update self.sub_opts and self.args.
 
     Raises: RaiseInvalidArgumentException if invalid args specified.
```
NickGoog, line 747: https://google.github.io/styleguide/pyguide.html#244-decision
```
Raises:
  RaiseInvalidArgumentException: Invalid args specified.
```

gslib/command.py
```diff
@@ -725,27 +729,47 @@ def RaiseInvalidArgumentException(self):
...
+    if args is None:
+      unparsed_args = self.args
+    else:
+      unparsed_args = args
     try:
-      self.sub_opts, self.args = getopt.getopt(
-          self.args, self.command_spec.supported_sub_args,
+      parsed_sub_opts, parsed_args = getopt.getopt(
+          unparsed_args, self.command_spec.supported_sub_args,
           self.command_spec.supported_private_args or [])
```
NickGoog, line 756: I am afraid of `or []` for the same reason you shouldn't do `def f(x=[])`.
I'm concerned it creates one object that will have old data if this section of code runs again.
Did you experiment to see if it would persist data?

gslib/command.py
```diff
@@ -725,27 +729,47 @@ def RaiseInvalidArgumentException(self):
...
+      self.sub_opts, self.args = parsed_sub_opts, parsed_args
+      if check_args:
+        self.CheckArguments()
+    else:
+      if check_args:
+        # This is just for sanity check. Only GetGcloudStorageArgs will
+        # call this method with update_sub_opts_and_args=False and it
+        # does not set set check_args to True.
+        raise TypeError(
+            'Cannot check arguments if sub_opts and args are not updated.')
```
NickGoog, line 769: Should this be "No reason to check arguments..." b/c it seems like you "can" technically do it?

gslib/command.py
```diff
@@ -725,27 +729,47 @@ def RaiseInvalidArgumentException(self):
...
-
-  def CheckArguments(self):
+    if update_sub_opts_and_args:
+      self.sub_opts, self.args = parsed_sub_opts, parsed_args
+      if check_args:
+        self.CheckArguments()
+    else:
+      if check_args:
+        # This is just for sanity check. Only GetGcloudStorageArgs will
+        # call this method with update_sub_opts_and_args=False and it
```
NickGoog, line 766: Comma after `False` b/c what follows the conjunction is an independent clause.

https://depts.washington.edu/engl/askbetty/sentence_structure.php
https://google.github.io/styleguide/pyguide.html#386-punctuation-spelling-and-grammar

gslib/command.py
```diff
@@ -1929,6 +1953,61 @@ def _ApplyThreads(self, thread_count, process_count, recursive_apply_level,
         # block the semaphore on a WorkerThread completion.
         worker_semaphore.release()
 
+  def _GetGcloudStorageArgs(self, sub_opts, gsutil_args, gcloud_storage_map):
```
NickGoog, line 1956: I'm thinking this should also be in a separate file, but I see it relies on instance variables. Maybe a subclass solution is possible or a util that gets `self` as an argument (except renamed)?

gslib/command.py
```diff
@@ -1929,6 +1953,61 @@ def _ApplyThreads(self, thread_count, process_count, recursive_apply_level,
...
+    if isinstance(gcloud_storage_map.gcloud_command, str):
+      args.append(gcloud_storage_map.gcloud_command)
+    elif isinstance(gcloud_storage_map.gcloud_command, dict):
+      # If a command has sub-commands, e.g pap set, pap get.
+      # All the flags mapping must be present in the subcommand's map
+      # because gsutil does not have command specific flags
+      # if sub-commands are present.
+      if gcloud_storage_map.flag_map:
+        raise ValueError(
+            'Flags mapping found at command lavel for the command: {}.'.format(
```
NickGoog, line 1971: `command level`

gslib/command.py
```diff
@@ -1929,6 +1953,61 @@ def _ApplyThreads(self, thread_count, process_count, recursive_apply_level,
...
+  def _GetGcloudStorageArgs(self, sub_opts, gsutil_args, gcloud_storage_map):
+    if gcloud_storage_map is None:
+      raise GcloudStorageTranslationError(
+          'Command "{}" cannot be translated to gcloud storage because the'
+          ' translation mapping is missing.'.format(self.command_name))
+    args = []
+    if isinstance(gcloud_storage_map.gcloud_command, str):
+      args.append(gcloud_storage_map.gcloud_command)
+    elif isinstance(gcloud_storage_map.gcloud_command, dict):
+      # If a command has sub-commands, e.g pap set, pap get.
```
NickGoog, line 1965: Should this be all-caps "PAP"?

gslib/command.py
```diff
@@ -1929,6 +1953,61 @@ def _ApplyThreads(self, thread_count, process_count, recursive_apply_level,
...
+              ' gcloud storage'.format(option))
+        args.append(gcloud_storage_map.flag_map[option].gcloud_flag)
+        if value != '':
+          args.append(value)
+    return args + gsutil_args
+
+  def GetGcloudStorageArgs(self):
+    """Translates the gsutil command flags to gcloud storage flags.
+
+    It uses the command_spec.gcloud_storage_map field which provides the
```
NickGoog, line 1997: In this type of sentence, "which" is for optional info and "that" is for mandatory info.

gslib/tests/test_command.py
```diff
@@ -0,0 +1,211 @@
...
+  """Implementation of a fake gsutil command."""
+  command_spec = command.Command.CreateCommandSpec(
+      'fake_with_sub',
+      min_args=1,
+      max_args=constants.NO_MAX,
+      supported_sub_args='ay:',
+      file_url_ok=True,
+      argparse_arguments={
+          'set': [
+              command_argument.CommandArgument.
```
NickGoog, line 68: I'm surprised this doesn't need to be wrapped in parenthesis:
```
(command_argument.CommandArgument.
 MakeZeroOrMoreCloudBucketURLsArgument())
```

gslib/command.py
```diff
@@ -496,6 +497,9 @@ class Command(HelpProvider):
   # Each subclass must override this with an instance of CommandSpec.
   command_spec = None
 
+  # Mapping for translating gsutil command to gcloud storage.
+  gcloud_storage_map = None
```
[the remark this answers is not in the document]
dilipped: I agree that this file has grown too big, but this is one of the things that has to belong to the class. The idea is that all the commands that inherit from this class will override this field. I think there are other potential candidates if we really want to refactor this class and reduce this size.

gslib/command.py
dilipped: I think you are talking about the default argument case `def f(x=[])`. This is not same as default argument because here we are deciding the argument while calling the method. So it should be fine and there is no risk.

gslib/command.py
dilipped: Subclass is not an option because this class is the parent of all the other commands. If we decide to subclass it, we will have to update the parent for all the commands, which seems unnecessary just to handle a new method.

I personally don't like passing in an entire object (self) to a util either. I feel this method belongs in this class. Like I said, there are other potential candidates if we want to refactor the code here, which I agree has grown a lot. But given the priority of the team, I don't think refactoring is going to happen anytime soon, so I would prefer leaving it like this.

gslib/utils/shim_util.py
dilipped: Flag map will change from command to command, so it can't be a constant.

gslib/command.py
dilipped: Done.

gslib/command.py
dilipped: Done.

gslib/command.py
dilipped: Reworded this.

gslib/command.py
dilipped: Done.

gslib/command.py
dilipped: Done.

gslib/command.py
dilipped: I think this should be lowercase as I am talking about the command itself. I have added gsutil in front of it to make it explicit that it is a command that I am talking about.

gslib/tests/test_command.py
dilipped: Yeah, yapf behaves strangely sometimes, and I don't always like it the way it formats things.

gslib/command.py
dilipped: Fixed. That's good to know. Thanks for explaining it.

gslib/command.py
dilipped: On a second thought, I think we can pass over the entire command instance from command runner. 
What I typically like to avoid  is having a circular reference (a method calls a util with it's own instance and then the util calls a method on that instance).  But looking at the code in CommandRunner, it creates the command instance and calls the RunCommand on it. We will be eventually calling something like `RunGcloudStorage` and for that, we can probably pass over the entire command instance to the util which also avoids the circular reference issue mentioned above and at the same time move this logic to the util. I will go ahead and make that change.

NickGoog: Or you could only pass the variables of from the CommandRunner intance you need

dilipped: I think passing `Command` instance is better because there a lot of things that we need from that instance. Plus we also call the `ParseSubOpts` method on the instance.

NickGoog (approved):

gslib/command.py
dilipped: I should've checked this before I moved the method. I revisited my design doc and realized why I had decided to keep GetGcloudStorageArgs under `Command` class. We want the ability to override this generic behavior if a command is too complicated to be translated using this generic method. So having this under `Command` class makes more sense to me. An example would be signurl where a service account key file is a positional argument in gsutil but it's going to be a required flag in gcloud storage. Reverted to keeping this method under `Command`.

NickGoog: What about moving the gcloud logic to a mixin class?

dilipped: Good idea. Mixin should work.

dilipped: Done.

NickGoog:
Should you move the tests for the mixin out to the util test file?

dilipped:
Oops, I think I missed pushing that commit to move the tests. Done.

PR #1399, opened 2021-11-11
Update perfdiag.py

rebecca-pete:
Can someone please provide links to more information for lines 241 (boto num_retries configuration parameter) and 258 (`netstat -s` command)? Thanks!

PR #1412, opened 2021-11-29
Revert "Update perfdiag.py"

rebecca-pete:
Reverts GoogleCloudPlatform/gsutil#1399


thomasmaclean:
What's the context for the reversion?

dilipped:
I think given that there is another PR #1413 that addresses some of the comments that were not addressed in #1399, I will close this one.

PR #1420, opened 2021-12-03
Update ls.py

rebecca-pete:
Hi @dilipped I am unable to verify lines 117-128. For example, when adding an object to a bucket that uses a flat namespace to mimic a subdirectory: "images/cat.png" Would you please take a look?  

Thanks!


dilipped (approved):

dilipped:
@rebecca-pete Regarding your question, if you have multiple updates in your bucket under sub directories,
e.g
```
gs://bucket/images/cat.png
gs://bucket/images/dog.png
```

You can do `gsutil ls gs://bucket/**` and you would be able to see all the buckets. That's what those lines mean.

PR #1421, opened 2021-12-06
Fix splitting of -m flag

dilipped:
b/208968970

### DIFF
--- a/gslib/commands/notification.py
+++ b/gslib/commands/notification.py
@@ -561,7 +561,7 @@ class NotificationCommand(Command):
             raise CommandException(
                 'Custom attributes specified with -m should be of the form '
                 'key:value')
-          key, value = a.split(':')
+          key, value = a.split(':', 1)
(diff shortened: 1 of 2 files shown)


gslib/commands/notification.py
```diff
@@ -561,7 +561,7 @@ def _Create(self):
             raise CommandException(
                 'Custom attributes specified with -m should be of the form '
                 'key:value')
-          key, value = a.split(':')
+          key, value = a.split(':', 1)
```
NickGoog, line 564: Would str.partition be better for a single split?

NickGoog (approved):

gslib/commands/notification.py
dilipped: I guess either should be fine? I will leave it as it is for now.

dilipped:
Kokoro tests are failing because of mTLS issues, which are fixed now.

PR #1424, opened 2021-12-08
Update ls.py

rebecca-pete:
Add sorting example to ``gsutil ls`` reference page.

### DIFF
--- a/gslib/commands/ls.py
+++ b/gslib/commands/ls.py
@@ -59,7 +59,6 @@ _DETAILED_HELP_TEXT = ("""
 Retrieves a list of providers, buckets, or objects matching the criteria,
 ordered in the list lexicographically by name.
 
-
 <B>LISTING PROVIDERS, BUCKETS, SUBDIRECTORIES, AND OBJECTS</B>
   If you run ``gsutil ls`` without URLs, it lists all of the Google Cloud Storage
   buckets under your default project ID (or all of the Cloud Storage buckets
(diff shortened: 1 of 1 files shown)


gslib/commands/ls.py
```diff
@@ -194,6 +193,21 @@
 
   See also "gsutil help acl" for getting a more readable version of the ACL.
 
+  <B>LIST ORDER</B>
+    List results are ordered lexicographically by name. While sorting
+    capabilities are not available for the ``gsutil ls`` command, Linux users
+    can use the Linux ``sort`` option and gsutil ``-l`` option to sort results
```
dilipped, line 199: I am a bit on the fence for this. I can see that it might be handy for users who are looking for sorting the output on linux on timestamp and can directly use the command below. But at the same time, I feel this probably does not belong here because of the following reasons.
1. It is Platform specific. Windows users will have to figure this out on their own anyway.
2. It is not unique to gsutil. The `sort` command in linux is used for sorting any text, so I am not sure if this belongs here. Basically I am wondering where we should be drawing the line. This like these are handy for specific use cases, but they also increase the documentation size and bury other important info.
3. This seems more like a recipe, which can probably belong to a separate recipe page. e.g there are recipe pages for python https://code.activestate.com/recipes/576611-counter-class/

rebecca-pete: Makes perfect sense. Closing the pull request.

PR #1428, opened 2021-12-16
Eliminate the multiprocessing warning

MOZGIII:
This is super annoying and should've been made opt-outable. Ethier way, not a major concern anymore, as the linked issue is closed long time ago.

### DIFF
--- a/gslib/command.py
+++ b/gslib/command.py
@@ -1349,15 +1349,6 @@ class Command(HelpProvider, GcloudStorageCommandMixin):
                'update your config file(s) (located at %s) and set '
                '"parallel_process_count = 1".') %
               (os_name, ', '.join(GetFriendlyConfigFilePaths())))))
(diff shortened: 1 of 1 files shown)


dilipped:
Thanks for the PR, but I think the warning is important as we have a lot of customers being confused about multiprocessing leading to deadlocks and other issues on Mac OS. The lined issue has been closed, but it has not been resolved, i.e. we still force python to use fork for multiprocessing which will create issues on Mac OS. 

I would recommend setting process_count to 1 to ensure that multiprocessing is disabled for MacOS which should also take care of not displaying the log message. Other ways to avoid the message would be to run the command in quiet mode.

Feel free to re-open or leave a comment if there are any other concerns.

MOZGIII:
Ok, how about we allow to silence it then?

The rationale is if I'm not seeing any issues with my local setup, why would I have to see this all the time?

Issue #579, opened 2018-11-14
rsync should support -z and -Z


greggdonovan:
We would like to use gzip Content-Encoding for all files that are uploaded via `gsutil rsync`, as we do with `gsutil cp`.

FlorianBurgevin:
+1

thiagopnts:
+1

thobrla:
Does -J help for your use case?  I think the main reason -z/-Z are not supported is that they are asymmetric, which causes challenges in reasoning for all future executions of rsync.

stanhu:
I looked into doing this. It seems the `cp` command sets the `gzip_local` flag: https://github.com/GoogleCloudPlatform/gsutil/blob/d53d1fc5c27da067c22e1dc61a27491b56e232ce/gslib/commands/cp.py#L1184

However, I can't see where this is actually used.

kosich:
@stanhu , afau, `gzip_local` flag is there only to understand if both it and `gzip_encoded` have been set to true to raise an error:

https://github.com/GoogleCloudPlatform/gsutil/blob/d53d1fc5c27da067c22e1dc61a27491b56e232ce/gslib/commands/cp.py#L1262-L1264

When using actual copying helper, both `cp` and `rsync` use only `gzip_encoded` flag and one of the `gzip_arg_exts`|`gzip_arg_all` params:

- setting gzip options in `cp` https://github.com/GoogleCloudPlatform/gsutil/blob/d53d1fc5c27da067c22e1dc61a27491b56e232ce/gslib/commands/cp.py#L1272-L1273
- setting gzip options in `rsync` 
https://github.com/GoogleCloudPlatform/gsutil/blob/d53d1fc5c27da067c22e1dc61a27491b56e232ce/gslib/commands/rsync.py#L1789-L1790

So, we could copy this behaviour in `rsync` and only set the `gzip_arg_exts` or `gzip_arg_all`  params, while keeping the `gzip_encoded` falsy:

https://github.com/GoogleCloudPlatform/gsutil/blob/d53d1fc5c27da067c22e1dc61a27491b56e232ce/gslib/commands/cp.py#L1244-L1249

——

NOTE: I'm guessing here, haven't debugged the code.

stanhu:
We might want to close this issue because of https://github.com/GoogleCloudPlatform/gsutil/pull/1430#issuecomment-1010234090.

dilipped:
[...]

PR #1430, opened 2021-12-21
Add -z and -Z options for gzip encoding to rsync command

stanhu:
This enables `rsync` to use the same options used in `cp` to compress
files before uploading.

Closes https://github.com/GoogleCloudPlatform/gsutil/issues/579

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -527,6 +527,58 @@ _DETAILED_HELP_TEXT = ("""
                  use ^ as an escape character instead of \\ and escape the |
                  character. When using Windows PowerShell, use ' instead of "
                  and surround the | character with ".
+
+  -z <ext,...>   Applies gzip content-encoding to any file upload whose
+                 extension matches the ``-z`` extension list. This is useful when
+                 uploading files with compressible content such as .js, .css,
+                 or .html files, because it reduces network bandwidth and storage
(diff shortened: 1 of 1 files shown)


stanhu:
@dilipped Would you mind having a look?

dilipped:
Thanks for the PR! I am a bit busy this week, so I will take a look at this next week.

dilipped:
The -z option compresses the file before uploading, so the size and hash of the file stored at the destination will not be the same as the file at the source, and hence future rsync commands will not be able to correctly identify the diff. For this reason, we should not be supporting the -z and -Z flag in rsync. Please consider using the -j or -J option or using the `cp` command which supports this flag. I am going to close this PR, but feel free to leave comments if you have any questions. Thanks!

stanhu:
@dilipped Thanks for that information. Ah, I see that these compressed files are always overwritten in an `rsync`. That's not such a bad problem for us; we'd rather have the files copied again if necessary as long as they are compressed. I suppose we could just switch to `cp`, but then large files that aren't compressed would then be copied unnecessarily. I wonder if there is a middle ground here?

dilipped:
Rsync is meant to sync the data between two folders, so having the files compressed at one end does not seem like the correct behavior for rsync. Like I said, `gsutil cp` is the right command to achieve something like this, and I am glad you have found the workaround :)

PR #1435, opened 2022-01-05
Fix shim test for gcloud

dilipped:
The release script for gcloud GCE tests were breaking because a test in shm_util relies on an env variable set by Cloud SDK.


NickGoog:
What is the cert being replaced with?

dilipped:
Oops, thanks for catching it. I think that got modified by mistake. Reverted it.

NickGoog (approved):

PR #1437, opened 2022-01-06
Remove shim util

dilipped:
The release test was still failing, so removing the left over shim_util stuff as well to unblock the release.


dilipped:
Merging it without review as this one is time sensitive. I am still going to wait for the review from Nick before I submit the release CL.

PR #1438, opened 2022-01-07
Bring back shim 

dilipped:
Reverting https://github.com/GoogleCloudPlatform/gsutil/pull/1437 and https://github.com/GoogleCloudPlatform/gsutil/pull/1436

The shim related changes were temporarily removed because the Gcloud GCE release tests were failing. I am rolling out a fix for the release infrastructure that takes care of this issue. It is now safe to re-introduce the shim-related changees.

### DIFF
--- a/gslib/command_runner.py
+++ b/gslib/command_runner.py
@@ -50,6 +50,7 @@ from gslib.gcs_json_api import GcsJsonApi
 from gslib.no_op_credentials import NoOpCredentials
 from gslib.tab_complete import MakeCompleter
 from gslib.utils import boto_util
+from gslib.utils import shim_util
 from gslib.utils import system_util
 from gslib.utils.constants import RELEASE_NOTES_URL
 from gslib.utils.constants import UTF8
@@ -410,7 +411,14 @@ class CommandRunner(object):
                                sub_opts=command_inst.sub_opts,
                                command_alias=command_name)
 
-    return_code = command_inst.RunCommand()
+    if command_inst.translate_to_gcloud_storage_if_requested():
+      # This does not mean that the gcloud storage command worked.
+      # It only means that we succesfully attempted running gcloud storage.
+      # The command itself might have failed.
+      return_code = command_inst.run_gcloud_storage()
+    else:
+      # Run gsutil.
+      return_code = command_inst.RunCommand()
 
     if CheckMultiprocessingAvailableAndInit().is_available and do_shutdown:
       ShutDownGsutil()
--- a/gslib/tests/test_shim_util.py
+++ b/gslib/tests/test_shim_util.py
@@ -1,0 +1,379 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for shim_util.py."""
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+import os
+import subprocess
+from unittest import mock
+
+from gslib import command
+from gslib import command_argument
+from gslib import exception
+from gslib.tests import testcase
+from gslib.utils import constants
+from gslib.utils import shim_util
+from gslib.tests import util
+
+
+class FakeCommandWithGcloudStorageMap(command.Command):
+  """Implementation of a fake gsutil command."""
+  command_spec = command.Command.CreateCommandSpec('fake_shim',
+                                                   min_args=1,
+                                                   max_args=constants.NO_MAX,
+                                                   supported_sub_args='irz:',
+                                                   file_url_ok=True)
+  gcloud_storage_map = shim_util.GcloudStorageMap(
+      gcloud_command='objects fake',
+      flag_map={
+          '-r': shim_util.GcloudStorageFlag(gcloud_flag='-x'),
+          '-z': shim_util.GcloudStorageFlag(gcloud_flag='--zip'),
+      })
+  help_spec = command.Command.HelpSpec(
+      help_name='fake_shim',
+      help_name_aliases=[],
+      help_type='command_help',
+      help_one_line_summary='Fake one line summary for the command.',
+      help_text='Help text for fake command.',
+      subcommand_help_text={},
+  )
+
+  def RunCommand(self):
+    print('FakeCommandWithGcloudStorageMap called')
+
+
+class FakeCommandWithSubCommandWithGcloudStorageMap(command.Command):
+  """Implementation of a fake gsutil command."""
+  command_spec = command.Command.CreateCommandSpec(
+      'fake_with_sub',
+      min_args=1,
+      max_args=constants.NO_MAX,
+      supported_sub_args='ay:',
+      file_url_ok=True,
+      argparse_arguments={
+          'set': [
+              command_argument.CommandArgument.
+              MakeZeroOrMoreCloudBucketURLsArgument()
+          ],
+          'get': [
+              command_argument.CommandArgument.MakeNCloudBucketURLsArgument(1)
+          ],
+      })
+  gcloud_storage_map = shim_util.GcloudStorageMap(gcloud_command={
+      'set':
+          shim_util.GcloudStorageMap(
+              gcloud_command='buckets update',
+              flag_map={
+                  '-a': shim_util.GcloudStorageFlag(gcloud_flag='-x'),
+                  '-y': shim_util.GcloudStorageFlag(gcloud_flag='--yyy'),
+              }),
+      'get':
+          shim_util.GcloudStorageMap(gcloud_command='buckets describe',
+                                     flag_map={})
+  },
+                                                  flag_map={})
+  help_spec = command.Command.HelpSpec(
+      help_name='fake_with_sub',
+      help_name_aliases=[],
+      help_type='command_help',
+      help_one_line_summary='Fake one line summary for the command.',
+      help_text='Help text for fake command with sub commands.',
+      subcommand_help_text={},
+  )
+
+
+class TestGetGCloudStorageArgs(testcase.GsUtilUnitTestCase):
+  """Test Command.get_gcloud_storage_args method."""
+
+  def setUp(self):
+    super().setUp()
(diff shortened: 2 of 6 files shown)


gslib/tests/test_shim_util.py
```diff
@@ -0,0 +1,379 @@
...
+          ['fake', 'gcloud', 'command'], {'fake_env_var': 'val'}, dry_run=True)
+      expected_calls = [
+          mock.call('Gcloud Storage Command: fake gcloud command'),
+          mock.call('Enviornment variables for Gcloud Storage:'),
+          mock.call('%s=%s', 'fake_env_var', 'val'),
+      ]
+      self.assertEqual(mock_logger.info.mock_calls, expected_calls)
+
+
+class TestRunGcloudStorage(testcase.GsUtilUnitTestCase):
```
NickGoog, line 346: Are you going to develop E2E tests at some point?
You could do what I did for STET, which is pretend a shell script is the gcloud binary, and have the shell script confirm it was called with the correct arguments.

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+      supports_output_translation (bool): Indicates if the corresponding
+        gcloud storage command supports the printing gsutil formatted output.
+    """
+    self.gcloud_command = gcloud_command
+    self.flag_map = flag_map
+    self.supports_output_translation = supports_output_translation
+
+
+def _get_gcloud_binary_path():
+  cloudsdk_root = os.environ.get('CLOUDSDK_ROOT_DIR')
```
NickGoog, line 75: Is this automatically added to env vars by a gcloud install?

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+
+  def _get_gcloud_storage_args(self, sub_opts, gsutil_args, gcloud_storage_map):
+    if gcloud_storage_map is None:
+      raise exception.GcloudStorageTranslationError(
+          'Command "{}" cannot be translated to gcloud storage because the'
+          ' translation mapping is missing.'.format(self.command_name))
+    args = []
+    if isinstance(gcloud_storage_map.gcloud_command, str):
+      args = gcloud_storage_map.gcloud_command.split()
+    elif isinstance(gcloud_storage_map.gcloud_command, dict):
```
NickGoog, line 103: Why do we have multiple types for `gcloud_command` (`str` and `dict`)?

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+      # If a command has sub-commands, e.g gsutil pap set, gsutil pap get.
+      # All the flags mapping must be present in the subcommand's map
+      # because gsutil does not have command specific flags
+      # if sub-commands are present.
+      if gcloud_storage_map.flag_map:
+        raise ValueError(
+            'Flags mapping found at command level for the command: {}.'.format(
```
NickGoog, line 110: This might sound too much like a success message

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+          self.command_name))
+
+    if sub_opts:
+      for option, value in sub_opts:
+        if option not in gcloud_storage_map.flag_map:
+          raise exception.GcloudStorageTranslationError(
+              'Command option "{}" cannot be translated to'
+              ' gcloud storage'.format(option))
+        args.append(gcloud_storage_map.flag_map[option].gcloud_flag)
+        if value != '':
```
NickGoog, line 129: Why not `if value`?

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+                                         self.gcloud_storage_map)
+
+  def _print_gcloud_storage_command_info(self,
+                                         gcloud_command,
+                                         env_variables,
+                                         dry_run=False):
+    logger_func = self.logger.info if dry_run else self.logger.debug
+    logger_func('Gcloud Storage Command: {}'.format(' '.join(gcloud_command)))
+    if env_variables:
+      logger_func('Enviornment variables for Gcloud Storage:')
```
NickGoog, line 157: "Environment" typo

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+                                  top_level_flags)
+        # TODO(b/206149936): Translate boto config to CLOUDSDK envs.
+        env_variables = {}
+        if use_gcloud_storage == 'dry_run':
+          self._print_gcloud_storage_command_info(gcloud_storage_command,
+                                                  env_variables,
+                                                  dry_run=True)
+        elif not os.environ.get('CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL'):
+          raise exception.GcloudStorageTranslationError(
+              'Gsutil is not using the same credentials as gcloud.'
```
NickGoog, line 195: Should we mention this is only an issue b/c the shim is active?

NickGoog (commented):
I didn't review the actual business logic b/c it was all approved earlier

gslib/command_runner.py
```diff
@@ -410,7 +411,14 @@ def RunNamedCommand(self,
                                sub_opts=command_inst.sub_opts,
                                command_alias=command_name)
 
-    return_code = command_inst.RunCommand()
+    if command_inst.translate_to_gcloud_storage_if_requested():
```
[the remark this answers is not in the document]
dilipped: The method is on `Command`, so you can read it as "translate the command to gcloud storage". Explaining what mechanism decides if the translation will happen in the function name will make it too long which is already pretty lengthy, and it is hard to capture all the mechanisms anyway. So I think the docstring is apt for that use case.

gslib/tests/test_shim_util.py
dilipped: Depends. I might add one test. But majority of this will be E2E tested for test parity anyway, which I guess we will have to run periodically eventually.

gslib/utils/shim_util.py
dilipped: yes.

gslib/utils/shim_util.py
dilipped: It can be a string if there is one-to-one mapping. But for cases which has sub-commands, we use a dict, because a sub-command will have its own mapping. See the test case test_get_gcloud_storage_args_parses_subcommands or the design doc for more info.

gslib/command_runner.py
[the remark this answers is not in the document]
NickGoog: It seems like the naming is reflecting an underlying code smell.
This structure might be better:
```
if command_inst.should_translate_to_gcloud_storage():
  command_inst.translate_to_gcloud_storage()
  ...
```

gslib/utils/shim_util.py
dilipped: The value is going to be empty string if it is not defined as parsed by the ParseSubOpts. So this check makes it more specific.

NickGoog: If you're checking for a specific type of falsey value, I would comment why

gslib/command_runner.py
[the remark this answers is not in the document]
dilipped: Yeah, I had thought of breaking it down that way, but I did not do it because the conditions on whether to call gsutil or gcloud are based on whether the translation was happened successfully as well, along with knowing whether we should perform the translation. 

Keeping it this way help us do the following in the command runner.
```
    if command_inst.translate_to_gcloud_storage_if_requested():
      # This does not mean that the gcloud storage command worked.
      # It only means that we succesfully attempted running gcloud storage.
      # The command itself might have failed.
      return_code = command_inst.run_gcloud_storage()
    else:
      # Run gsutil.
      return_code = command_inst.RunCommand()
```

I had used the structure that you've mentioned earlier and it led to too much handling in command_runner (even with using helper) and I wanted to keep the command_runner code as clean as possible. Ideally, I wanted it to do "if condition 'call gcloud storage' else 'call gsutil' "

This is just a cost of keeping the command_runner straight-forward, which I feel is not a bad trade-off

gslib/utils/shim_util.py
```diff
@@ -0,0 +1,219 @@
...
+    Returns:
+      True if the command was successfully translated, else False.
+    """
+    use_gcloud_storage = config.get('GSUtil', 'use_gcloud_storage', 'never')
+    if use_gcloud_storage not in VALID_USE_GCLOUD_STORAGE_VALUES:
+      raise exception.CommandException(
+          'Invalid option specified for'
+          ' GSUtil:use_gcloud_storage config setting. Should be one of: {}'.
+          format(' | '.join(VALID_USE_GCLOUD_STORAGE_VALUES)))
+    if use_gcloud_storage != 'never':
```
NickGoog, line 178: You can unindent everything below if you follow the "guard pattern."
```
if use_gcloud_storage == 'never'
  return
```

Also, ideally we use an enum instead of the raw string value "never"

gslib/utils/shim_util.py
dilipped: Good point. Done.

gslib/utils/shim_util.py
dilipped: Done

gslib/utils/shim_util.py
dilipped: Updated it.

gslib/command_runner.py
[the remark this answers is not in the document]
NickGoog: That's fine. I guess it avoids:
```
return_code = None
if is_configured_to_translate():
  did_translation = attempt_translation()
  if did_translation:
    return_code = run_translation()
if not return_code:
  return_code = run_original()
```

gslib/utils/shim_util.py
NickGoog: Looks the same

gslib/utils/shim_util.py
NickGoog: Looks the same

NickGoog: Nvm

gslib/utils/shim_util.py
dilipped: True, but used != to avoid repeating `return False` twice. 

Else we end up doing
```
if use_gcloud_storage == `never`:
  return False

# Perform bunch of other logic

return False.
```

To me, the way it is currently felt more intuitive. 

Regarding enum, I avoided it because it is python3 only, and for python2 we need to add another dependency which felt like overkill for a shim. I am instead using constants now which is slightly better than using raw stirngs directly.

NickGoog: Gsutil doesn't support Python 2 anymore

dilipped: yes, but we still support backporting things in maintenance mode, and an invalid import will break all the tests. It's best to keep this until we have totally deprecated python 2.

NickGoog: Aren't the backports done in a separate branch? https://github.com/GoogleCloudPlatform/gsutil/tree/py2

NickGoog (approved):
Discussed offline, Dilip is making an enum for "should_translate" setting values

gslib/utils/shim_util.py
dilipped: Done

PR #1440, opened 2022-01-07
Starsandskies backfills

starsandskies:
Backfill documentation changes that were revealed during the most recent generation of the c.g.c documentation

### DIFF
--- a/gslib/commands/pap.py
+++ b/gslib/commands/pap.py
@@ -30,7 +30,7 @@ from gslib.third_party.storage_apitools import storage_v1_messages as apitools_m
 from gslib.utils.constants import NO_MAX
 
 _SET_SYNOPSIS = """
-  gsutil pap set (enforced|unspecified) gs://<bucket_name>...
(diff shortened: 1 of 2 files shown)


gslib/commands/pap.py
```diff
@@ -30,7 +30,7 @@
 from gslib.utils.constants import NO_MAX
 
 _SET_SYNOPSIS = """
-  gsutil pap set (enforced|unspecified) gs://<bucket_name>...
+  gsutil pap set (enforced|inherited) gs://<bucket_name>...
```
dilipped, line 33: Oh, I have made similar changes in another PR, which does this in addition to removing unspecified from the code itself.  https://github.com/GoogleCloudPlatform/gsutil/pull/1445

dilipped (approved):

gslib/commands/pap.py
dilipped: This can be merged btw, my comment was more of an FYI.

PR #1443, opened 2022-01-11
Replace git with https in .gitmodules

dilipped:
Git Actions CI started breaking with https://github.com/GoogleCloudPlatform/gsutil/runs/4779653292?check_suite_focus=true

As per the link in the error message, this is happening because github is no longer supporting unauthenticated git protocol. This only affects git:// and not https://


NickGoog (approved):
Please confirm Kokoro passes

dilipped:
Kokoro passes.

PR #1445, opened 2022-01-12
PAP: Replace unspecified with inherited

dilipped:
Now that inherited is supported and public, we can replace unspecified with inherited.


starsandskies:
LGTM for documentation changes!

rrauber (approved):

PR #1450, opened 2022-01-18
Docs: Use case for deleting objects, not bucket

cbonnie:
Add a use case for how to delete all objects in a bucket without deleting the bucket itself.


starsandskies:
LGTM, but adding a gsutil SWE for final approval!

NickGoog (approved):
Let me know if you don't have permissions to hit the "squash and merge" button.

PR #1452, opened 2022-01-19
Allow external account based credentials

ScruffyProdigy:
In Gcloud, users can specify external account based credentials, and we 
should be able to use that here in gsutil.


NickGoog:
@ScruffyProdigy , adding someone more familiar with the project

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
...
+
+  @property
+  def token_expiry(self):
+    return self._base.expiry
+
+  @token_expiry.setter
+  def token_expiry(self, value):
+    self._base.expiry = value
+
+  def _to_json(self, strip, to_serialize=None):
```
dilipped, line 55: This seems to be very similar to the implementation present in Oauth2client https://github.com/googleapis/oauth2client/blob/50d20532a748f18e53f7d24ccbe6647132c979a9/oauth2client/client.py#L241

Any reason why we don't just call super().to_json() and then update the serialized data?

```
def to_json(self):
  serialized_data = super().to_json()
  deserialized_data = json.loads(serialized_data)
  deserialized_data['_base'] = ...
  #. add/replace any other additional data.
  return json.dumps(deserialized_data)
```

This will be much shorter and also  avoid having the need to override a private method.

dilipped (commented):
Yet to review the tests. Left some initial comments. BTW, did you try running any gsutil command with external account? Does it work as expected?

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
+import copy
+import datetime
+import json
+
+import google.auth.aws
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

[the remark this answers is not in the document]
ScruffyProdigy: Done.  Same with the other modules used from this library

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
+import copy
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/gcs_json_credentials.py
```diff
@@ -272,6 +275,20 @@ def _HasImpersonateServiceAccount():
   return _GetImpersonateServiceAccount() not in (None, '')
 
 
+def _HasExternalAccountCreds():
+  return config.has_option('Credentials', 'gs_external_account_file')
+
+
+def _GetExternalAccountCredentials():
+  if not _HasExternalAccountCreds():
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
...
+]
+
+
+class WrappedCredentials(oauth2client.client.OAuth2Credentials):
+
+  def __init__(self, base):
+    if not isinstance(base, google.auth.credentials.Credentials):
+      raise TypeError("Invalid Credentials")
+    self._base = base
+    super(WrappedCredentials, self).__init__(None, base._audience, None, None,
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
...
+                                                  scopes=DEFAULT_SCOPES)
+  except Exception as e:
+    try:
+      # Check if configuration corresponds to an Identity Pool credentials.
+      creds = google.auth.identity_pool.Credentials.from_file(
+          filename, scopes=DEFAULT_SCOPES)
+    except Exception as e:
+      # If the configuration is invalid or does not correspond to any
+      # supported external_account credentials, raise an error.
+      return None
```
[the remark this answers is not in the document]
ScruffyProdigy: Fixed

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,166 @@
...
+      creds = google.auth.identity_pool.Credentials.from_info(
+          info, scopes=DEFAULT_SCOPES)
+    except Exception as e:
+      # If the configuration is invalid or does not correspond to any
+      # supported external_account credentials, raise an error.
+      return None
+  return creds
+
+
+def _get_external_account_credentials_from_file(filename):
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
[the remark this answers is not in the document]
ScruffyProdigy: You know, I'm not sure.  This file was originally written by another team member who is currently on leave.  I'm setting all of the values to None for right now, and will try to go through manual testing again tomorrow morning to confirm that everything still works.  Leaving all of the arguments as keyword arguments to make things easier to edit

gslib/utils/wrapped_credentials.py
ScruffyProdigy: Done

gslib/gcs_json_credentials.py
```diff
@@ -228,18 +231,18 @@ def _CheckAndGetCredentials(logger):
     # If we didn't actually try to authenticate because there were multiple
     # types of configured credentials, don't emit this warning.
     if failed_cred_type:
-      if logger.isEnabledFor(logging.DEBUG):
+      if logger and logger.isEnabledFor(logging.DEBUG):
```
dilipped, line 234: Was this giving any error? Any reason for modifying it?

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,155 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Classes and functions to allow google.auth credentials to be used within oauth2client."""
+
+import copy
+import datetime
+import io
+import json
+
+from google.auth import aws, credentials, identity_pool
```
dilipped, line 22: One import per line, please
```
from google.auth import aws
from google.auth import credentials
...
```

gslib/utils/wrapped_credentials.py
dilipped: I think we should override the `to_json` method instead of `_to_json`.

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,155 @@
...
+        to_serialize: dict, (Optional) The properties for this object
+                      that will be serialized. This allows callers to
+                      modify before serializing.
+
+    Returns:
+        string, a JSON representation of this instance, suitable to pass to
+        from_json().
+    """
+
+    serialized_data = super()._to_json(strip=strip, to_serialize=to_serialize)
```
dilipped, line 90: You can call `to_json` instead.

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,155 @@
...
+    deserialized_data = json.loads(serialized_data)
+    deserialized_data['_base'] = copy.copy(self._base.info)
```
dilipped, line 92: Do you need to add access token and expiry information as well that you had in the previous code?

.gitmodules
```diff
@@ -55,3 +55,6 @@
 [submodule "gslib/vendored/oauth2client"]
 	path = gslib/vendored/oauth2client
 	url = https://github.com/gsutil-mirrors/oauth2client.git
+[submodule "third_party/google-auth-library-python"]
```
dilipped, line 58: You should also update https://github.com/GoogleCloudPlatform/gsutil/blob/6bfe364782ab3dc25c29606c421bc58e9ff7c6dc/gsutil.py#L71

dilipped (commented):
I think you shou

gslib/utils/wrapped_credentials.py
ScruffyProdigy: done

gslib/utils/wrapped_credentials.py
ScruffyProdigy: done

gslib/utils/wrapped_credentials.py
ScruffyProdigy: The parent function actually takes care of expiry information, but it looks liek access token should be handled manually

.gitmodules
ScruffyProdigy: Done

gslib/gcs_json_credentials.py
ScruffyProdigy: It was, but doesn't seem to be doing so anymore.  Removed

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+      self.assertIsInstance(creds, ServiceAccountCredentials)
+
+  def testOauth2UserCredential(self):
+    with SetBotoConfigForTest(self.botoCredentialsSet(user_account_creds="?")):
+      self.assertTrue(gcs_json_credentials._HasOauth2UserAccountCreds())
+      client = gcs_json_api.GcsJsonApi(None, None, None, None)
+      self.assertIsInstance(client.credentials,
+                            reauth_creds.Oauth2WithReauthCredentials)
+
+  @mock.patch(
```
dilipped, line 84: Prefer mock.patch.object

```
from apitools.base.py import credentials_lib
...
...

@mock.patch.object(credentials_lib, 'GceAssertionCredentials', autospec=True)
def ...
```

This way you don't have to worry about the module path which might differ based on where the module is getting imported.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+      "gslib.gcs_json_credentials.credentials_lib.GceAssertionCredentials",
+      autospec=True,
+  )
+  def testGCECredential(self, mock_credentials):
+
+    def set_store(store):
+      mock_credentials.return_value.store = store
+
+    mock_credentials.return_value.client_id = None
+    mock_credentials.return_value.refresh_token = "rEfrEshtOkEn"
```
dilipped, line 94: Ideally if you are setting a return_value on a mock, that value should be checked after the call has been performed. I don't see any checks for these return values.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+    mock_credentials.return_value.set_store = set_store
```
dilipped, line 95: Do you get any error if you don't set this? Given that the return value is a mock object, I feel it should still work without this. Also, note that you can set side_effects for a mock object as well. Let me know if you need any additional help with setting up this test.

If you care about checking that set_store is getting called, you can do
```
mock_credentials.return_value.set_store.assert_called_once_with(.....)
```

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
```
dilipped, line 1: Thanks for adding these tests!

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
+# -*- coding: utf-8 -*-
```
dilipped, line 1: Can you please add test coverage for this file?

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+
+    mock_credentials.return_value.client_id = None
+    mock_credentials.return_value.refresh_token = "rEfrEshtOkEn"
+    mock_credentials.return_value.set_store = set_store
+    with SetBotoConfigForTest(self.botoCredentialsSet(gce_creds="?")):
+      self.assertTrue(gcs_json_credentials._HasGceCreds())
+      client = gcs_json_api.GcsJsonApi(None, None, None, None)
+      self.assertIsInstance(client.credentials, GceAssertionCredentials)
+
+  def testExternalAccountCredential(self):
```
dilipped, line 101: Can you also add a test to check the failure state. i.e to ensure that `failed_cred_types` is getting used for external account as expected. I know that we do not have this coverage for other stuff, but going forward we want to ensure that we add 100% test coverage for any new code.

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+  NON_SERIALIZED_MEMBERS = frozenset(
+      list(oauth2client.client.OAuth2Credentials.NON_SERIALIZED_MEMBERS) +
+      ['_base'])
+
+  def __init__(self, base):
+    if not isinstance(base, credentials.Credentials):
+      raise TypeError("Invalid Credentials")
+    self._base = base
+    super(WrappedCredentials, self).__init__(access_token=None,
+                                             client_id=base._audience,
```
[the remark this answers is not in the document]
ScruffyProdigy: When I change the value to `None`, my manual tests start failing, so I'm going to leave this here

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+    return self._base.expiry
+
+  @token_expiry.setter
+  def token_expiry(self, value):
+    self._base.expiry = value
+
+  def to_json(self):
+    """Utility function that creates JSON repr. of a Credentials object.
+
+    Args:
```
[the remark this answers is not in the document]
ScruffyProdigy: Fixed

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+from google.auth import aws
+from google.auth import credentials
+from google.auth import identity_pool
+
+import oauth2client
+from google.auth.transport import requests
+from gslib.utils import constants
+
+# Expiry is stored in RFC3339 UTC format
+EXPIRY_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: You're welcome!

gslib/utils/wrapped_credentials.py
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: Done

dilipped:
@ScruffyProdigy  Looks like the CI tests are failing with the error `ModuleNotFoundError: No module named 'cachetools'`. It might be possible that the google auth library has another dependencies. If that's the case then you will have to ensure that those dependencies get adding to the third_party folder as well.

dilipped:
Also, I see that some of the review comments are yet to be addressed. Please leave a reply on all the review comments if you think you have addressed them. Thanks.

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: The test fails if I don't set each of these fields on the Mock.  I added in a test to check the refresh token, though

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: It's not actually super important that set_store gets called, but the test does indeed fail if I don't set this method, because `store` doesn't get set properly, and this is the method that sets it. If you have any suggestions for how to avoid this, let me know

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+]
+
+
+class WrappedCredentials(oauth2client.client.OAuth2Credentials):
+  NON_SERIALIZED_MEMBERS = frozenset(
+      list(oauth2client.client.OAuth2Credentials.NON_SERIALIZED_MEMBERS) +
+      ['_base'])
+
+  def __init__(self, base):
+    if not isinstance(base, credentials.Credentials):
```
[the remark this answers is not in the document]
ScruffyProdigy: `google.auth.external_account.Credentials` is the only class that we use here.  The code base only assumes methods defined in `google.auth.credentials.Credentials`, and any subclass of that will technically work here.  Currently, the unit test I use subclasses from `google.auth.credentials.Credentials`, and adds a couple of mocks as necessary

gslib/utils/wrapped_credentials.py
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+    constants.Scopes.CLOUD_PLATFORM,
+    constants.Scopes.CLOUD_PLATFORM_READ_ONLY,
+    constants.Scopes.FULL_CONTROL,
+    constants.Scopes.READ_ONLY,
+    constants.Scopes.READ_WRITE,
+]
+
+
+class WrappedCredentials(oauth2client.client.OAuth2Credentials):
+  NON_SERIALIZED_MEMBERS = frozenset(
```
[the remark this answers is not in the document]
ScruffyProdigy: Yes, because otherwise the default to_json will try to turn the credentials into a JSON object and fail

gslib/utils/wrapped_credentials.py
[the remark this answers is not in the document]
ScruffyProdigy: The only credentials that get used are `google.auth.external_account.Credentials`, but nothing in the method requires this.  The only methods that get assumed are from its superclass: `google.auth.credentials.Credentials`.  The unit test that I created subclasses from `google.auth.credentials.Credentials`

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+
+add_move(MovedModule("mock", "mock", "unittest.mock"))
+from six.moves import mock
+
+
+class TestGcsJsonCredentials(testcase.GsUtilUnitTestCase):
+  """Test logic for interacting with GCS JSON Credentials."""
+
+  @staticmethod
+  def botoCredentialsSet(
```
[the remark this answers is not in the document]
ScruffyProdigy: I moved it to a module function.  I think it does avoid a lot of repetition, because each test needs to make sure to include "None" for the other credential types to make sure to overwrite any existing credentials in the BOTO File

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,118 @@
...
+  def testOauth2ServiceAccountCredential(self):
+    contents = pkgutil.get_data("gslib", "tests/test_data/test.p12")
+    tmpfile = self.CreateTempFile(contents=contents)
+    with SetBotoConfigForTest(
+        self.botoCredentialsSet(service_account_creds={
+            "keyfile": tmpfile,
+            "client_id": "?"
+        })):
+      self.assertTrue(gcs_json_credentials._HasOauth2ServiceAccountCreds())
+      client = gcs_json_api.GcsJsonApi(None, None, None, None)
```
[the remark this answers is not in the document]
ScruffyProdigy: Fixed

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+        data['token_expiry'] = None
+    creds.token_expiry = data.get("token_expiry")
+    return creds
+
+
+def _get_external_account_credentials_from_info(info):
+  try:
+    # Check if configuration corresponds to an AWS credentials.
+    creds = aws.Credentials.from_info(info, scopes=DEFAULT_SCOPES)
+  except Exception as e:
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
```diff
@@ -0,0 +1,162 @@
...
+    try:
+      # Check if configuration corresponds to an Identity Pool credentials.
+      creds = identity_pool.Credentials.from_info(info, scopes=DEFAULT_SCOPES)
+    except Exception as e:
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_data/test_credentials.json
```diff
@@ -0,0 +1,9 @@
+{
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/utils/wrapped_credentials.py
ScruffyProdigy: Done

third_party/cachetools/.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,29 @@
+---
+name: Bug report
```
dilipped, line 2: Looks like the code has been directly added. You have to instead add the gitmodule. This is similar to how you have added the google auth library. See that in `third_party` directory, it should look like a reference `cachetools @ <commit>`

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,175 @@
...
+from apitools.base.py import GceAssertionCredentials
+from google_reauth import reauth_creds
+from gslib import gcs_json_api
+from gslib import gcs_json_credentials
+from gslib.cred_types import CredTypes
+from gslib.exception import CommandException
+from gslib.tests import testcase
+from gslib.tests.util import SetBotoConfigForTest
+from gslib.utils.wrapped_credentials import WrappedCredentials
+import logging
```
dilipped, line 31: std lib imports usually go on  the top. go/pystyle#imports-formatting

gslib/tests/test_wrapped_credentials.py
```diff
@@ -0,0 +1,78 @@
...
+  """Test logic for interacting with Wrapped Credentials."""
+
+  @mock.patch("oauth2client.transport.httplib2.Http", autospec=True)
+  def testWrappedCredentialUsage(self, http):
+    http.return_value.request.return_value = (RESPONSE, CONTENT)
+
+    creds = WrappedCredentials(MockCredentials(token=ACCESS_TOKEN))
+
+    http = oauth2client.transport.get_http_object()
+    creds.authorize(http)
```
dilipped, line 62: It's not really clear what we are trying to test here. The creds instance is not getting used as you might expect.
i.e I think even if we remove this line, this test will still pass. So we are not really testing any of the WrappedCredentials functionality.

It might be worth testing the specific methods in this class, because looks like that's where most of the logic is present. e.g to_json, from_json, for_external_account etc. Let me know if you need any help on this.

gsutil.py
```diff
@@ -84,6 +84,8 @@ def OutputAndExit(message):
     ('httplib2', submodule_pyvers),
     ('retry-decorator', ''),
     ('six', ''),  # Python 2 / 3 compatibility dependency
+    ('google-auth-library-python', ''),
+    ('cachetools', ''),
```
dilipped, line 88: I think this line should be before google-auth as it is the dependency for googl-auth. Please also leave a comment saying that this is for google-auth dependency.

gslib/tests/test_wrapped_credentials.py
ScruffyProdigy: I updated the test to make sure that requests being sent actually use the information we set in the mocked credentials.  I also added some tests for to_json and from_json.  Also, while testing, I noticed that WrappedCredentials doesn't *really* work with any type of credentials; at least not if we intend serialization to work, because from_json assumes that we are using either AWS or IdentityPool based credentials, so I've updated some of WrappedCredentials accordingly

gsutil.py
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: Done

gslib/tests/test_wrapped_credentials.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,175 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

third_party/cachetools/.github/ISSUE_TEMPLATE/bug_report.md
ScruffyProdigy: I might need some help with this

gslib/utils/wrapped_credentials.py
```diff
@@ -23,12 +23,12 @@
...
 from google.auth import identity_pool
 
-import oauth2client
 from google.auth.transport import requests
 from gslib.utils import constants
+import oauth2client
 
 # Expiry is stored in RFC3339 UTC format
-EXPIRY_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+EXPIRY_FORMAT = oauth2client.client.EXPIRY_FORMAT
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gsutil.py
ScruffyProdigy: Done

dilipped:
Looks like we will need to add all the dependencies for requests lib as well. Also, when I tried this locally, looks not all the dependencies for requests library work with python 3.5 if we are using the latest versions. So you might have to point to a commit that can work with python 3.5. We can sync up offline and try to setup the dependencies correctly.

.gitmodules
```diff
@@ -64,3 +64,24 @@
...
+[submodule "third_party/charset_normalizer"]
+	path = third_party/charset_normalizer
+	url = https://github.com/Ousret/charset_normalizer
+[submodule "third_party/chardet"]
+	path = third_party/chardet
+	url = https://github.com/chardet/chardet
+[submodule "third_party/certifi"]
+	path = third_party/certifi
+	url = https://github.com/certifi/python-certifi.git
+[submodule "third_party/pyopenssl"]
```
dilipped, line 85: This is not needed

You can run `git rm -f third_party/pyopenssl`

gsutil.py
```diff
@@ -85,7 +85,14 @@ def OutputAndExit(message):
...
     ('six', ''),  # Python 2 / 3 compatibility dependency
     ('cachetools', 'src'),  # google auth dependency
+    ('urllib3', 'src'),  # requests dependency
+    ('charset_normalizer', ''),  # requests dependency
+    ('chardet', ''),  # requests dependency
+    ('certifi', ''),  # requests dependency
+    ('idna', ''),  # requests dependency
     ('requests', ''),  # google auth dependency
+    ('aiohttp', ''),  # google auth dependency
+    ('pyopenssl', ''), # google auth dependency
```
dilipped, line 95: Please remove this.

.gitmodules
ScruffyProdigy: Done

gsutil.py
ScruffyProdigy: DOne

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,184 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 Google Inc. All Rights Reserved.
```
NickGoog, line 2: Please use "Google LLC"
https://opensource.google/documentation/reference/copyright

Repeated below

dilipped:
So looks like kokoro for windows is failing on Python 3.7 because Python 3.7 has google-auth installed (maybe because of the bundled python) and the module gets loaded from site-packages. So when `from google.auth import aws` is run, it is not actually calling the package from `third_party/`. This is happening because `google` module gets already loaded when python is invoked. I am not sure why this package gets loaded just on python invocation (even before gsutil.py is called). I will be looking into that further.

This does point to another issue. Even if the package was loaded from `site-packages`, it should still have worked. But looks like it is not able to find `aws`. On further investigation, it seems like `aws` module was recently added and perhaps was not released for pip.

gslib/tests/test_wrapped_credentials.py
```diff
@@ -0,0 +1,124 @@
...
+CONTENT = "content"
+RESPONSE = httplib2.Response({
+    "content-type": "text/plain",
+    "status": "200",
+    "content-length": len(CONTENT),
+})
+
+
+class MockCredentials(external_account.Credentials):
+  refresh = mock.Mock()
```
NickGoog, line 43: Why not also declare this in `__init__`?
```
self.refresh = mock.Mock(side_effect=side_effect)
```

gslib/tests/test_wrapped_credentials.py
```diff
@@ -0,0 +1,124 @@
...
+                        token_url="baz",
+                        credential_source="qux"))
+
+    http = oauth2client.transport.get_http_object()
+    creds.authorize(http)
+    response, content = http.request(uri="www.google.com")
+    self.assertEquals(content, CONTENT)
+    MockCredentials.refresh.assert_called_once_with(mock.ANY)
+
+    # Make sure the default request gets called with the correct token
```
NickGoog, line 91: Please add a period https://google.github.io/styleguide/pyguide.html#386-punctuation-spelling-and-grammar

NickGoog (commented):
Reviewed for readability. Feel free to focus on logic issues first.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,184 @@
...
+  def testOauth2ServiceAccountFailure(self, _):
+    contents = pkgutil.get_data("gslib", "tests/test_data/test.p12")
+    tmpfile = self.CreateTempFile(contents=contents)
+    with SetBotoConfigForTest(
+        botoCredentialsSet(service_account_creds={
+            "keyfile": tmpfile,
+            "client_id": "?"
+        })):
+      with self.assertLogs() as logger:
+        with self.assertRaises(Exception) as exc:
```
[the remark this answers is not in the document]
ScruffyProdigy: I think the problem here is in `gcs_json_credentials.py`, because it only raises Exceptions.  We could update this file to toss a more specific exception if you think one is applicable?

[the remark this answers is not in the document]
NickGoog: If the implementation code is raising too broad an exception, then this is fine. But if you see that only a certain set of errors are raised, then please be more specific in the test code.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,184 @@
...
+try:
+  from OpenSSL.crypto import load_pkcs12
+  HAS_OPENSSL = True
+except ImportError:
+  HAS_OPENSSL = False
+
+ERROR_MESSAGE = "This is the error message"
+
+
+def botoCredentialsSet(
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,184 @@
...
+):
+  config = []
+  if service_account_creds:
+    config.append(("Credentials", "gs_service_key_file",
+                   service_account_creds["keyfile"]))
+    config.append(("Credentials", "gs_service_client_id",
+                   service_account_creds["client_id"]))
+  else:
+    config.append(("Credentials", "gs_service_key_file", None))
+  config.append(("Credentials", "gs_oauth2_refresh_token", user_account_creds))
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -0,0 +1,184 @@
...
+  """Test logic for interacting with GCS JSON Credentials."""
+
+  @unittest.skipUnless(HAS_OPENSSL, 'signurl requires pyopenssl.')
+  def testOauth2ServiceAccountCredential(self):
+    contents = pkgutil.get_data("gslib", "tests/test_data/test.p12")
+    tmpfile = self.CreateTempFile(contents=contents)
+    with SetBotoConfigForTest(
+        botoCredentialsSet(service_account_creds={
+            "keyfile": tmpfile,
+            "client_id": "?"
```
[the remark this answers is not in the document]
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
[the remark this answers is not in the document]
ScruffyProdigy: The implementation code is raising too broad of an exception, and testing for Exception being raised here is the only way to catch it

gslib/tests/test_wrapped_credentials.py
ScruffyProdigy: Done

gslib/tests/test_wrapped_credentials.py
ScruffyProdigy: Done

gslib/tests/test_gcs_json_credentials.py
ScruffyProdigy: Done

dilipped:
Since `google-auth` has namespace package, the `google` namespace will get loaded when the interpreter starts which leads to this issue where the installed `google-auth` lib gets read instead of the one in `third_party`.  Thanks to another googler who explained this to me.

dilipped:
Here is the hacky solution: https://github.com/GoogleCloudPlatform/gsutil/pull/1469

 I don't think there is any other way to deal with it, so I guess we will just have to settle for this.

NickGoog:
@dilipped that reminds me of something I saw in gcloud. You might want to cite gcloud as the source.

dilipped:
@NickGoog Yes, this is based off of what gcloud does, but their code is not public, so can't link it directly. @ScruffyProdigy maybe mention that in the comment as well that gcloud does the same thing. Once caveat is that gcloud does not have the issue with python 3.5 probably because it does not use multiprocessing

gslib/tests/test_gsutil.py
```diff
@@ -58,13 +58,13 @@ class TestGsUtilUnit(testcase.GsUtilUnitTestCase):
   """Unit tests for top-level gsutil command."""
 
   @mock.patch.object(importlib, "reload")
-  def test_fix_google_module(self, importlib_reload):
+  def test_fix_google_module(self, mock_reload):
     with mock.patch.dict('sys.modules', {"google": "google"}):
       gsutil._fix_google_module()
-      importlib_reload.assert_called()
+      mock_reload.assert_called()
```
dilipped, line 64: `assert_called` will not work for 3.5. Please use assert_called_once_with instead.

gslib/tests/test_gsutil.py
```diff
@@ -58,13 +58,13 @@ class TestGsUtilUnit(testcase.GsUtilUnitTestCase):
   """Unit tests for top-level gsutil command."""
 
   @mock.patch.object(importlib, "reload")
```
dilipped, line 60: Please add autospec=True here.

e.g @mock.patch.object(importlib, 'reload', autospec=True)

gslib/tests/test_gsutil.py
```diff
@@ -58,13 +58,13 @@ class TestGsUtilUnit(testcase.GsUtilUnitTestCase):
...
-  def test_fix_google_module(self, importlib_reload):
+  def test_fix_google_module(self, mock_reload):
     with mock.patch.dict('sys.modules', {"google": "google"}):
       gsutil._fix_google_module()
-      importlib_reload.assert_called()
+      mock_reload.assert_called()
 
   @mock.patch.object(importlib, "reload")
-  def test_fix_google_module(self, importlib_reload):
+  def test_fix_google_module(self, mock_reload):
```
dilipped, line 67: The test name is same as above

Also, ideally the test name should explain what's being tested. So maybe `test_fix_google_module_does_not_reload_if_module_missing`.

gslib/tests/test_gsutil.py
ScruffyProdigy: Done

gslib/tests/test_gsutil.py
ScruffyProdigy: Done

gslib/tests/test_gsutil.py
ScruffyProdigy: Done

dilipped (commented):
Sorry, few more nit items that I missed catching in my previous review. Once these are addressed we should be good to merge it once the kokoro tests run successfully.

@NickGoog since you had left some comments, can you please LGTM this if everything looks good?

dilipped:
Looks like `testHasConfiguredCredentialsNoCreds` is failing.

gslib/tests/test_boto_util.py
```diff
@@ -52,3 +53,58 @@ def testSetsHostBaseToMtlsIfClientCertificateEnabled(self,
         domain='',
         password='password',
     )
+
+  def testHasConfiguredCredentialsNoCreds(self):
```
dilipped, line 57: So this one is probably failing on XML because it is probably reading the config from a parent config. You might have to do something like this

https://github.com/GoogleCloudPlatform/gsutil/blob/9b3767664199da0c71ac59f77fd4a9c0fbb43d78/gslib/tests/test_shim_util.py#L42

For now you can just copy over that function and use it. I can roll out a refactor later on to move this function to gslib.tests.util and use it in both these module

You will probably need to use it for all the other tests here as well.

ScruffyProdigy: I found another solution that directly targets the method being tested

dilipped (approved):

gsutil.py
```diff
@@ -42,6 +42,28 @@ def OutputAndExit(message):
   sys.exit(1)
 
 
+def _fix_google_module():
+  """Reloads the google module to prefer our third_party copy.
+
+  When python is not invoked with the -S option, it can preload google module
```
NickGoog, line 48: Capitalize "Python"

gsutil.py
```diff
@@ -42,6 +42,28 @@ def OutputAndExit(message):
...
+  via .pth file. After this happens, our third_party google
+  package may not in the __path__. After our third_party dependency directory is
+  put at the first place in the sys.path, google module should be reloaded,
+  so that our vendored copy can be preferred.
```
NickGoog, line 52: No comma before "so that"

gsutil.py
```diff
@@ -42,6 +42,28 @@ def OutputAndExit(message):
...
+  """
+  if 'google' not in sys.modules:
+    return
+  # This is a hacky solution as reloading a module is never recommended.
+  # But given that it is rare that users will run into this issue where
+  # they might have google-auth already installed in their python environment,
```
NickGoog, line 58: Why not just put all this information in the docstring instead of half-docstring and half-code comment?
Also, "Python" should be capitalized here.

gsutil.py
ScruffyProdigy: Done

gsutil.py
ScruffyProdigy: Done

gsutil.py
ScruffyProdigy: Done

ScruffyProdigy:
> @NickGoog Explaining .pth file might be an overkill here, but I think we can throw in a link where someone can read more about it.
> 
> I have modified the text below, if this looks, @ScruffyProdigy can use this instead of the current one.
> 
> ```
> When Python is not invoked with the -S option, it can preload google module
> via .pth file (https://docs.python.org/3/library/site.html). This means,
> instead of using the google package from our "third_party" directory, it will end up
> using the package from site_packages. To ensure
> that the google package from our "third_party" directory is preferred,
> we should insert the path at the start of sys.path and reload the google module.
> 
> This is a hacky solution as reloading a module is never recommended.
> But given that it is rare that users will run into this issue where
> they might have google-auth already installed in their python environment,
> and the fact that we are reloading it right at the beginning, it should be
> fine. Note that this reload might be an issue for Python 3.5.3 and lower
> because of the weakref issue which was fixed in Python 3.5.4:
> https://github.com/python/cpython/commit/9cd7e17640a49635d1c1f8c2989578a8fc2c1de6.
> ```

Okay, I used this

gsutil.py
```diff
@@ -46,10 +46,11 @@ def _fix_google_module():
...
   When Python is not invoked with the -S option, it can preload google module
-  via .pth file. After this happens, our third_party google
-  package may not in the __path__. After our third_party dependency directory is
-  put at the first place in the sys.path, google module should be reloaded
-  so that our vendored copy can be preferred.
+  via .pth file (https://docs.python.org/3/library/site.html). This means,
+  instead of using the google package from our "third_party" directory, it will end up
+  using the package from site_packages. To ensure
+  that the google package from our "third_party" directory is preferred,
+  we should insert the path at the start of sys.path and reload the google module.
```
NickGoog, line 53: Still a bit tough to understand in its current form.
I'm pretty sure this shorter version works.
@dilipped can confirm.
```
When Python is *not* invoked with the -S option, it may preload the google module via .pth file.
This adds the "site_packages" version to "__path__" instead of the gsutil "third_party" version.
To force the "third_party" version, reload the google module.

This is a hacky. Reloading is required for the rare case that users have
google-auth already installed in their Python environment.
Note that this reload may cause an issue for Python 3.5.3 and lower
because of the weakref issue, fixed in Python 3.5.4:
https://github.com/python/cpython/commit/9cd7e17640a49635d1c1f8c2989578a8fc2c1de6.
```

I noticed Dilip's second version got rid of the term "vendored." This is good if it's intentional b/c we have a separate "vendored" folder with a specific meaning.

dilipped: >When Python is *not* invoked with the -S option, it may preload the google module via .pth file.

Based of your previous question , I think adding the link here would be helpful. I guess other people might be unaware of `.pth` as well and they would know where to read more about it.

>This adds the "site_packages" version to "__path__" instead of the gsutil "third_party" version.

I intentionally removed `__path__` out because it was not clear which `__path__` we are talking about. So I would probably change this to just say `"site_packages" version is preferred over gsutil "third_party" version`.

>To force the "third_party" version, reload the google module.`

This seems partially correct. You need to have the path present in front of `sys.path` before you reload and I think that's an important detail to mention here.

NickGoog: Sounds good

ScruffyProdigy: So:

> When Python is *not* invoked with the -S option, it may preload the google module via .pth file.
> This "site_packages" version is preferred over gsutil "third_party" version.
> To force the "third_party" version, insert the path at the start of sys.path and reload the google module.
> 
> This is a hacky. Reloading is required for the rare case that users have
> google-auth already installed in their Python environment.
> Note that this reload may cause an issue for Python 3.5.3 and lower
> because of the weakref issue, fixed in Python 3.5.4:
> https://github.com/python/cpython/commit/9cd7e17640a49635d1c1f8c2989578a8fc2c1de6.

dilipped: >This "site_packages" version is preferred over gsutil "third_party" version.

Maybe `This lead to the "site_packages" version being preferred over gsutil "third_party" version.

dilipped:
Kokoro passes for everything!

dilipped:
Merged! @ScruffyProdigy Thank you very much for the PR and addressing all the review comments!

PR #1453, opened 2022-01-25
Shim for running gcloud storage: Add rm support

dilipped:


dilipped:
Kokoro is failing because of an issue with cp and ls tests. I have fixed those in #1455

gslib/commands/rm.py
```diff
@@ -221,6 +223,19 @@ class RmCommand(Command):
       subcommand_help_text={},
   )
 
+  # TODO(b/b/188092601) Add -f after continue-on-error is supported in
```
rrauber, line 226: b/ is repeated here

gslib/tests/test_rm.py
```diff
@@ -61,6 +67,22 @@ def _CleanRmUiOutputBeforeChecking(self, stderr):
...
+    """Remove irrelevant lines from the output lines."""
+    stderr_lines = stderr.splitlines()
+    valid_lines = []
+    strings_to_remove = set(['Removing objects:', 'Removing Buckets:', '  '])
+    for line in stderr_lines:
+      if line in strings_to_remove:
+        continue
+      # Gcloud storage logs a '.' for each resource in non-interactive mode.
+      # This might get added to the start of a valid line.
+      # e.g. .Removing gs://bucket/obj
```
rrauber, line 80: Is the period before "Removing" intentional?

rrauber (approved):

gslib/tests/test_rm.py
dilipped: Yes, it is intentional. I added quotes around the string to make it obvious.

gslib/commands/rm.py
dilipped: Actually, this entire TODO should be removed because it was already taken care of.

PR #1461, opened 2022-02-02
Fixes an issue with setting user when preserve POSIX is enabled.

thomasmaclean:
Note that this does not have any tests associated with it given the fact this file is uncovered and currently lacks a mechanism for mocking user IDs.


dilipped:
Can you please add some info on why this change is needed? Any issue number?  Also, regarding the test, maybe we can add a unit test in cp for this one. Maybe we can mock `os.geteuid()`

dilipped (approved):

PR #1465, opened 2022-02-11
Fix copying duplicate file name conflicts.

NickGoog:
Applies to mv logic as well. Basically treats destination as an existing
directory if multiple top-level sources are detected.

### DIFF
--- a/gslib/command.py
+++ b/gslib/command.py
@@ -737,7 +737,7 @@ class Command(HelpProvider, GcloudStorageCommandMixin):
       check_args: True to have CheckArguments() called after parsing.
       args: List of args. If None, self.args will be used.
       should_update_sub_opts_and_args: True if self.sub_opts and self.args
-        should be updated with the values returned after parsing. Else return a 
+        should be updated with the values returned after parsing. Else return a
         tuple of sub_opts, args returned by getopt.getopt. This is done
         to allow this method to be called from get_gcloud_storage_args in which
         case we do not want to update self.sub_opts and self.args.
@@ -896,7 +896,12 @@ class Command(HelpProvider, GcloudStorageCommandMixin):
           for blr in self.WildcardIterator(
               url.url_string).IterBuckets(bucket_fields=['id']):
             name_expansion_for_url = NameExpansionResult(
-                url, False, False, blr.storage_url, None)
+                source_storage_url=url,
+                is_multi_source_request=False,
+                is_multi_top_level_source_request=False,
+                names_container=False,
+                expanded_storage_url=blr.storage_url,
+                expanded_result=None)
             acl_func(self, name_expansion_for_url)
       else:
         multi_threaded_url_args.append(url_str)
--- a/gslib/tests/test_cp.py
+++ b/gslib/tests/test_cp.py
@@ -1703,6 +1703,56 @@ class TestCp(testcase.GsUtilIntegrationTestCase):
 
     _CopyAndCheck()
 
+  def test_copy_duplicate_nested_object_names_to_new_cloud_dir(self):
+    """Tests copying from bucket to same bucket preserves file structure."""
+    bucket_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+
+    # Use @Retry as hedge against bucket listing eventual consistency.
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _CopyAndCheck():
+      self.RunGsUtil(
+          ['cp', '-R',
+           suri(bucket_uri) + '/*',
+           suri(bucket_uri) + '/dst'])
+      stdout = self.RunGsUtil(['ls', '-R', bucket_uri.uri], return_stdout=True)
+      self.assertIn(suri(bucket_uri) + '/dst/dir1/file.txt', stdout)
+      self.assertIn(suri(bucket_uri) + '/dst/dir2/file.txt', stdout)
+
+    _CopyAndCheck()
+
+  def test_copy_duplicate_nested_object_names_to_existing_cloud_dir(self):
+    """Tests copying from bucket to same bucket preserves file structure."""
+    bucket_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dst/existing_file.txt',
+                      contents=b'data')
+
+    # Use @Retry as hedge against bucket listing eventual consistency.
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _CopyAndCheck():
+      self.RunGsUtil(
+          ['cp', '-R',
+           suri(bucket_uri) + '/*',
+           suri(bucket_uri) + '/dst'])
+      stdout = self.RunGsUtil(['ls', '-R', bucket_uri.uri], return_stdout=True)
+      self.assertIn(suri(bucket_uri) + '/dst/dir1/file.txt', stdout)
+      self.assertIn(suri(bucket_uri) + '/dst/dir2/file.txt', stdout)
+      self.assertIn(suri(bucket_uri) + '/dst/existing_file.txt', stdout)
+
+    _CopyAndCheck()
+
   @SkipForGS('Only s3 V4 signatures error on location mismatches.')
   def test_copy_bucket_to_bucket_with_location_redirect(self):
     # cp uses a sender function that raises an exception on location mismatches,
--- a/gslib/tests/test_mv.py
+++ b/gslib/tests/test_mv.py
@@ -21,7 +21,6 @@ from __future__ import unicode_literals
 
 import os
 
-import crcmod
 from gslib.cs_api_map import ApiSelector
 from gslib.tests.test_cp import TestCpMvPOSIXBucketToLocalErrors
 from gslib.tests.test_cp import TestCpMvPOSIXBucketToLocalNoErrors
@@ -36,7 +35,77 @@ from gslib.utils.retry_util import Retry
 from gslib.utils.system_util import IS_WINDOWS
 
 
-class TestMv(testcase.GsUtilIntegrationTestCase):
+class TestMvUnitTests(testcase.GsUtilUnitTestCase):
+  """Unit tests for mv command."""
+
+  def test_move_bucket_objects_with_duplicate_names_inter_bucket(self):
+    """Tests moving multiple top-level items between buckets."""
+    bucket1_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket1_uri,
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket1_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    bucket2_uri = self.CreateBucket()
+
+    self.RunCommand('mv', [suri(bucket1_uri, '*'), suri(bucket2_uri)])
+
+    actual = set(
+        str(u)
+        for u in self._test_wildcard_iterator(suri(bucket2_uri, '**')).IterAll(
+            expand_top_level_buckets=True))
+    expected = set([
+        suri(bucket2_uri, 'dir1', 'file.txt'),
+        suri(bucket2_uri, 'dir2', 'file.txt'),
+    ])
+    self.assertEqual(actual, expected)
+
+  def test_move_bucket_objects_with_duplicate_names_to_bucket_subdir(self):
+    """Tests moving multiple top-level items between buckets."""
+    bucket1_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket1_uri,
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket1_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    bucket2_uri = self.CreateBucket()
+
+    self.RunCommand('mv', [suri(bucket1_uri, '*'), suri(bucket2_uri, 'dir')])
+
+    actual = set(
+        str(u)
+        for u in self._test_wildcard_iterator(suri(bucket2_uri, '**')).IterAll(
+            expand_top_level_buckets=True))
+    expected = set([
+        suri(bucket2_uri, 'dir', 'dir1', 'file.txt'),
+        suri(bucket2_uri, 'dir', 'dir2', 'file.txt'),
+    ])
+    self.assertEqual(actual, expected)
+
+  def test_move_dirs_with_duplicate_file_names_to_bucket(self):
+    """Tests moving multiple top-level items to a bucket."""
+    bucket_uri = self.CreateBucket()
+    dir_path = self.CreateTempDir(test_files=[
+        ('dir1', 'file.txt'),
+        ('dir2', 'file.txt'),
+    ])
+
+    self.RunCommand('mv', [dir_path + '/*', suri(bucket_uri)])
+
+    actual = set(
+        str(u)
+        for u in self._test_wildcard_iterator(suri(bucket_uri, '**')).IterAll(
+            expand_top_level_buckets=True))
+    expected = set([
+        suri(bucket_uri, 'dir1', 'file.txt'),
+        suri(bucket_uri, 'dir2', 'file.txt'),
+    ])
+    self.assertEqual(actual, expected)
+
+
+class TestMvE2ETests(testcase.GsUtilIntegrationTestCase):
   """Integration tests for mv command."""
 
   def test_moving(self):
@@ -201,3 +270,28 @@ class TestMv(testcase.GsUtilIntegrationTestCase):
         'Warning: moving nearline object %s may incur an early deletion '
         'charge, because the original object is less than 30 days old '
         'according to the local system time.' % suri(object_uri), stderr)
+
+  def test_move_bucket_objects_with_duplicate_names_to_dir(self):
+    """Tests moving multiple top-level items to a bucket."""
+    bucket_uri = self.CreateBucket()
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    self.AssertNObjectsInBucket(bucket_uri, 2)
+
+    tmpdir = self.CreateTempDir()
+    self.RunGsUtil(['mv', suri(bucket_uri, '*'), tmpdir])
+
+    file_list = []
+    for dirname, _, filenames in os.walk(tmpdir):
+      for filename in filenames:
+        file_list.append(os.path.join(dirname, filename))
+    self.assertEqual(len(file_list), 2)
+    self.assertIn('{}{}dir1{}file.txt'.format(tmpdir, os.sep, os.sep),
+                  file_list)
+    self.assertIn('{}{}dir2{}file.txt'.format(tmpdir, os.sep, os.sep),
+                  file_list)
+    self.AssertNObjectsInBucket(bucket_uri, 0)
--- a/gslib/tests/test_naming.py
+++ b/gslib/tests/test_naming.py
@@ -975,15 +975,75 @@ class GsutilNamingTests(testcase.GsUtilUnitTestCase):
       ])
       self.assertEqual(expected, actual)
 
-  # @SequentialAndParallelTransfer
+  def testRecursiveCopyFileToExistingBucketSubDirInvalidSourceParent(self):
+    """Tests recursive copy of invalid path file to existing bucket subdir."""
+    src_dir = self.CreateTempDir(test_files=[('nested', 'f0')])
+    dst_bucket_uri = self.CreateBucket(test_objects=['dst_subdir/existing_obj'])
+
+    for relative_path_string in ['.', '.' + os.sep]:
+      self.RunCommand('cp', [
+          '-R',
+          src_dir + os.sep + relative_path_string,
+          suri(dst_bucket_uri, 'dst_subdir'),
+      ])
+      actual = set(
+          str(u) for u in self._test_wildcard_iterator(
+              suri(dst_bucket_uri, 'dst_subdir', '**')).IterAll(
+                  expand_top_level_buckets=True))
+      expected = set([
+          suri(dst_bucket_uri, 'dst_subdir', 'nested', 'f0'),
+          suri(dst_bucket_uri, 'dst_subdir', 'existing_obj'),
+      ])
+      self.assertEqual(expected, actual)
+
+  def testRecursiveCopyFilesToExistingBucketSubDirInvalidSourceParent(self):
+    """Tests recursive copy of invalid paths files to existing bucket subdir."""
+    src_dir = self.CreateTempDir(test_files=['f0'])
+    dst_bucket_uri = self.CreateBucket(test_objects=['dst_subdir/existing_obj'])
+
+    for relative_path_string in ['.', '.' + os.sep, '..', '..' + os.sep]:
+      with self.assertRaises(InvalidUrlError):
+        self.RunCommand('cp', [
+            '-R',
+            src_dir,
+            relative_path_string,
+            suri(dst_bucket_uri, 'dst_subdir'),
+        ])
+
   def testRecursiveCopyObjsAndFilesToNonExistentBucketSubDir(self):
     """Tests recursive copy of objs + files to non-existent bucket subdir."""
     src_bucket_uri = self.CreateBucket(test_objects=['f0', 'nested/f1'])
-    src_dir = self.CreateTempDir(test_files=['f2', ('nested', 'f3')])
+    src_dir = self.CreateTempDir(test_files=[
+        ('parent', 'f2'),
+        ('parent', 'nested', 'f3'),
+    ])
     dst_bucket_uri = self.CreateBucket()
     self.RunCommand('cp', [
-        '-R', src_dir,
+        '-R',
+        os.path.join(src_dir, 'parent'),
         suri(src_bucket_uri),
+        suri(dst_bucket_uri, 'dst_subdir')
+    ])
+    actual = set(
+        str(u) for u in self._test_wildcard_iterator(suri(
+            dst_bucket_uri, '**')).IterAll(expand_top_level_buckets=True))
+    expected = set([
+        suri(dst_bucket_uri, 'dst_subdir', src_bucket_uri.bucket_name, 'f0'),
+        suri(dst_bucket_uri, 'dst_subdir', src_bucket_uri.bucket_name, 'nested',
+             'f1'),
+        suri(dst_bucket_uri, 'dst_subdir', 'parent', 'f2'),
+        suri(dst_bucket_uri, 'dst_subdir', 'parent', 'nested', 'f3')
+    ])
+    self.assertEqual(expected, actual)
+
+  def testRecursiveCopyNestedObjsToNonExistentBucketSubDir(self):
+    """Tests recursive copy of objs + files to non-existent bucket subdir."""
+    src_bucket_uri = self.CreateBucket(
+        test_objects=['parent/f0', 'parent/nested/f1'])
+    dst_bucket_uri = self.CreateBucket()
+    self.RunCommand('cp', [
+        '-R',
+        suri(src_bucket_uri, 'parent'),
         suri(dst_bucket_uri, 'dst_subdir')
     ])
     actual = set(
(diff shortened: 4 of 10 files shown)


gslib/utils/copy_helper.py
```diff
@@ -670,59 +673,67 @@ def ConstructDstUrl(src_url,
...
   if (global_copy_helper_opts.perform_mv and recursion_requested and
-      src_url_names_container and not have_existing_dest_subdir):
+      src_url_names_container and not preserve_src_top_level_dirs):
     # Case 1. Handle naming rules for bucket subdir mv. Here we want to
     # line up the src_url against its expansion, to find the base to build
     # the new name. For example, running the command:
     #   gsutil mv gs://bucket/abcd gs://bucket/xyz
     # when processing exp_src_url=gs://bucket/abcd/123
     # exp_src_url_tail should become /123
     # Note: mv.py code disallows wildcard specification of source URL.
```
dilipped, line 693: This is no longer valid?

gslib/utils/copy_helper.py
```diff
@@ -670,59 +673,67 @@ def ConstructDstUrl(src_url,
...
+  # If there are multiple top-level source items, copy full file hierarchies.
+  # This avoids conflicts such as "dir1/f.txt" and "dir2/f.txt" both getting
+  # copied to "gs://bucket/f.txt". Linux normally errors on these conflicts,
+  # but we cannot do that because we need to give users the ability to create
+  # dirs as they copy to the cloud.
+  # Ignore this rule if using double wildcard ** because that treats all files
+  # as top-level, in which case the user doesn't want to preserve directories.
+  preserve_src_top_level_dirs = '**' not in src_url.versionless_url_string and (
+      has_multiple_top_level_srcs or have_existing_dest_subdir)
   if (global_copy_helper_opts.perform_mv and recursion_requested and
```
dilipped, line 685: Is this still specific to mv? I thought we were planning to change this behavior for cp as well?

gslib/utils/copy_helper.py
```diff
@@ -670,59 +673,67 @@ def ConstructDstUrl(src_url,
...
+      # where src_url ends. For example, for src_url=gs://bucket/ and
+      # exp_src_url=gs://bucket/src_subdir/obj, dst_key_name should be
+      # src_subdir/obj.
+      src_url_path_sans_final_dir = GetPathBeforeFinalDir(src_url, exp_src_url)
+      dst_key_name = exp_src_url.versionless_url_string[
+          len(src_url_path_sans_final_dir):].lstrip(src_url.delim)
+
+      # Handle case where dst_url is a non-existent subdir unless
+      # multiple top-level sources, in which case preserve source hierarchy.
+      if not preserve_src_top_level_dirs:
```
dilipped, line 713: If we remove the mv condition from the above if condition, cand we drop this condition in that case. From the way I understand it now, the behavior for mv and cp is going to be the same. In that case, I think if we drop the perform_mv condition, do we still need this check?

gslib/utils/copy_helper.py
```diff
@@ -670,59 +673,67 @@ def ConstructDstUrl(src_url,
...
-
-  if (not recursive_move_to_new_subdir and
-      (exp_dst_url.IsFileUrl() or _ShouldTreatDstUrlAsBucketSubDir(
-          have_multiple_srcs, exp_dst_url, have_existing_dest_subdir,
-          src_url_names_container, recursion_requested))):
-    if exp_dst_url.object_name and exp_dst_url.object_name.endswith(
-        exp_dst_url.delim):
-      dst_key_name = '%s%s%s' % (exp_dst_url.object_name.rstrip(
-          exp_dst_url.delim), exp_dst_url.delim, dst_key_name)
+    if preserve_src_top_level_dirs or (src_url_names_container and
```
dilipped, line 699: I wonder if we need the second condition after the OR here. preserve_src_top_level_dirs kinda guarantees that the destination is an existing dir correct? And I think we don't care if the source is a container as long as we have to preserve the top level dirs?

dilipped (commented):
Yet to review the tests.

gslib/command.py
```diff
@@ -896,7 +896,12 @@ def ApplyAclFunc(self,
           for blr in self.WildcardIterator(
               url.url_string).IterBuckets(bucket_fields=['id']):
             name_expansion_for_url = NameExpansionResult(
-                url, False, False, blr.storage_url, None)
+                source_storage_url=url,
```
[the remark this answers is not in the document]
NickGoog: kwargs are kool

gslib/utils/copy_helper.py
NickGoog: The `mv` code in the "Case 1" block is:
```
if (global_copy_helper_opts.perform_mv and recursion_requested
    and src_url_names_container and not preserve_src_top_level_dirs):
  exp_src_url_tail = (exp_src_url.url_string[len(src_url.url_string):])
  dst_key_name = '%s/%s' % (exp_dst_url.object_name.rstrip('/'),
                              exp_src_url_tail.strip('/'))
```
To get to this block you're commenting on, you have to meet this condition (combining "Case 2" and `not preserve_src_top_level_dirs`):
```
if src_url_names_container and (exp_dst_url.IsCloudUrl() or exp_dst_url.IsDirectory():
  src_url_path_sans_final_dir = GetPathBeforeFinalDir(src_url, exp_src_url)
  dst_key_name = exp_src_url.versionless_url_string[
          len(src_url_path_sans_final_dir):].lstrip(src_url.delim)
  dst_key_name = dst_key_name.partition(src_url.delim)[-1]
```

It looks like there isn't an overlap of logic we can condense.

gslib/utils/copy_helper.py
dilipped: Right, but what I am saying is that wouldn't the behavior for mv and cp end up being the same with the new naming rules? do we have to treat mv and cp differently?

NickGoog: I don't think the behavior is the same. This block is more like "rename" behavior.
More details in responses below.

dilipped: Discussed offline. Waiting for Nick to provide command examples/scenarios where mv and cp will have different behavior.

NickGoog: Removing Case I looks like a success

gslib/utils/copy_helper.py
```diff
@@ -663,41 +654,36 @@ def ConstructDstUrl(src_url,
...
+  # If there are multiple top-level source items, copy full file hierarchies.
+  # This avoids conflicts such as "dir1/f.txt" and "dir2/f.txt" both getting
+  # copied to "gs://bucket/f.txt". Linux normally errors on these conflicts,
+  # but we cannot do that because we need to give users the ability to create
+  # dirs as they copy to the cloud.
+  # Ignore this rule if using double wildcard ** because that treats all files
+  # as top-level, in which case the user doesn't want to preserve directories.
+  preserve_src_top_level_dirs = '**' not in src_url.versionless_url_string and (
+      has_multiple_top_level_srcs or have_existing_dest_subdir)
+  if preserve_src_top_level_dirs or (src_url_names_container and
```
dilipped, line 673: Do we need the case (src_url_names_container and (exp_dst_url.IsCloudUrl() or exp_dst_url.IsDirectory())) 

I am a bit confused here. Based on the description about for case 1, seems like we are handing only the preserve_src_top_level_dirs for case 1. This means a rename like operation `gsutil mv dir gs://non-existing_dir` will be handled by the else part correct? So do we need to have anything other than preserve_src_top_level_dirs here?

gslib/utils/copy_helper.py
```diff
@@ -663,41 +654,36 @@ def ConstructDstUrl(src_url,
...
-    # exp_src_url_tail should become /123
-    # Note: mv.py code disallows wildcard specification of source URL.
-    recursive_move_to_new_subdir = True
-    exp_src_url_tail = (exp_src_url.url_string[len(src_url.url_string):])
-    dst_key_name = '%s/%s' % (exp_dst_url.object_name.rstrip('/'),
-                              exp_src_url_tail.strip('/'))
-
-  elif src_url_names_container and (exp_dst_url.IsCloudUrl() or
-                                    exp_dst_url.IsDirectory()):
+  # If there are multiple top-level source items, copy full file hierarchies.
```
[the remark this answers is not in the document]
NickGoog: Done

gslib/utils/copy_helper.py
```diff
@@ -663,41 +654,36 @@ def ConstructDstUrl(src_url,
...
+  # but we cannot do that because we need to give users the ability to create
+  # dirs as they copy to the cloud.
+  # Ignore this rule if using double wildcard ** because that treats all files
+  # as top-level, in which case the user doesn't want to preserve directories.
+  preserve_src_top_level_dirs = '**' not in src_url.versionless_url_string and (
+      has_multiple_top_level_srcs or have_existing_dest_subdir)
+  if preserve_src_top_level_dirs or (src_url_names_container and
+                                     (exp_dst_url.IsCloudUrl() or
+                                      exp_dst_url.IsDirectory())):
     # Case 2.  Container copy to a destination other than a file.
```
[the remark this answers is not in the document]
NickGoog: Oops, outdated code comments

gslib/utils/copy_helper.py
```diff
@@ -706,15 +692,13 @@ def ConstructDstUrl(src_url,
     # gs://dest/./abc.
     if dst_key_name.startswith('.%s' % os.sep):
       dst_key_name = dst_key_name[2:]
-
   else:
     # Case 3.
```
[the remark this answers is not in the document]
NickGoog: Ditto

gslib/utils/copy_helper.py
```diff
@@ -663,41 +654,36 @@ def ConstructDstUrl(src_url,
   #    should create objects named like gs://bucket/subdir/dir2/a/b/c. In
   #    contrast, if gs://bucket/subdir does not exist, this same command
   #    should create objects named like gs://bucket/subdir/a/b/c.
-  # 3. Copying individual files or objects to dirs, buckets or bucket subdirs
+  # 2. Copying individual files or objects to dirs, buckets or bucket subdirs
```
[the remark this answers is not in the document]
NickGoog: Restructured the entire block

gslib/utils/copy_helper.py
NickGoog: I responded to the [earlier thread](https://github.com/GoogleCloudPlatform/gsutil/pull/1465#discussion_r813417927) about this line by explaining the boolean logic differences. Also note there is a `if not preserve_src_top_level_dirs` block for handling the case `(src_url_names_container and (exp_dst_url.IsCloudUrl() or exp_dst_url.IsDirectory()))`.

And tests fail without the case. Here's one example:
```
FAIL: testCopyingSubdirRecursiveToNonexistentSubdir (gslib.tests.test_naming.GsutilNamingTests)
Tests copying a directory with a single file recursively to a bucket.
  File "/usr/local/google/home/hartunian/Desktop/ide/gsutil/gslib/tests/test_naming.py", line 634, in testCopyingSubdirRecursiveToNonexistentSubdir
    self.assertEqual(expected, actual)
AssertionError: Items in the first set but not the second:
'gs://gsutil-test-testcopyingsubdirrecursivetononexisten-5c92d77f/dir3/dir2/foo'
Items in the second set but not the first:
'gs://gsutil-test-testcopyingsubdirrecursivetononexisten-5c92d77f/dir3/foo'
```

gslib/utils/copy_helper.py
```diff
@@ -663,58 +655,58 @@ def ConstructDstUrl(src_url,
...
-  # 3. Copying individual files or objects to dirs, buckets or bucket subdirs
+  #
+  #    If there are multiple top-level source items, preserve source parent
+  #    dirs. This is similar to when the destination dir already exists and
+  #    avoids conflicts such as "dir1/f.txt" and "dir2/f.txt" both getting
+  #    copied to "gs://bucket/f.txt". Linux normally errors on these conflicts,
+  #    but we cannot do that because we need to give users the ability to create
+  #    dirs as they copy to the cloud.
+  #
+  #    Note: "mv" is similar to running "cp -r" followed by source deletion.
```
dilipped, line 666: I like the new comment!

gslib/utils/copy_helper.py
dilipped: Got it. Thanks for checking

gslib/utils/copy_helper.py
```diff
@@ -663,58 +655,58 @@ def ConstructDstUrl(src_url,
...
   else:
-    # Case 3.
+    # Case 2.
     dst_key_name = exp_src_url.object_name.rpartition(src_url.delim)[-1]
 
-  if (not recursive_move_to_new_subdir and
-      (exp_dst_url.IsFileUrl() or _ShouldTreatDstUrlAsBucketSubDir(
-          have_multiple_srcs, exp_dst_url, have_existing_dest_subdir,
-          src_url_names_container, recursion_requested))):
+  if (exp_dst_url.IsFileUrl() or _ShouldTreatDstUrlAsBucketSubDir(
```
dilipped, line 707: Don't we have to handle special cases for `.` and `..` like we do in the `gcloud storage` code? Or is that handled already?

gslib/tests/test_cp.py
```diff
@@ -1700,6 +1700,31 @@ def _CopyAndCheck():
...
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+
+    # Use @Retry as hedge against bucket listing eventual consistency.
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _CopyAndCheck():
+      self.RunGsUtil(
+          ['cp', '-R',
+           suri(src_bucket_uri) + '/*',
+           suri(dst_bucket_uri)])
```
dilipped, line 1720: I think a similar test should be created for non-existing destination dir as well, because that's the key change that this PR is introducing.

gslib/tests/test_mv.py
```diff
@@ -119,6 +119,61 @@ def test_move_dir_to_bucket(self):
...
+                      object_name='dir1/file.txt',
+                      contents=b'data')
+    self.CreateObject(bucket_uri=bucket1_uri,
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    self.AssertNObjectsInBucket(bucket1_uri, 2)
+    bucket2_uri = self.CreateBucket()
+    self.AssertNObjectsInBucket(bucket2_uri, 0)
+
+    self.RunGsUtil(['mv', suri(bucket1_uri, '*'), suri(bucket2_uri)])
```
dilipped, line 135: If I understand correctly, this would work fine even without the change made for multiple top level sources given that the destination exists. I think we should check for a non-existing destination dir.

gslib/tests/test_naming.py
```diff
@@ -979,19 +979,56 @@ def testRecursiveCopyObjsAndFilesToExistingBucketSubDir(self):
   def testRecursiveCopyObjsAndFilesToNonExistentBucketSubDir(self):
     """Tests recursive copy of objs + files to non-existent bucket subdir."""
     src_bucket_uri = self.CreateBucket(test_objects=['f0', 'nested/f1'])
-    src_dir = self.CreateTempDir(test_files=['f2', ('nested', 'f3')])
+    src_dir = self.CreateTempDir(test_files=[('parent',
+                                              'f2'), ('parent', 'nested',
+                                                      'f3')])
```
dilipped, line 984: If you change this to `'f3'),])` (adding a comma after the tuple), I think yapf will format this in a slightly better way.

NickGoog:
Removing the two commit structure to make merging in the requested changes simpler

gslib/utils/copy_helper.py
NickGoog: Thanks

gslib/utils/copy_helper.py
NickGoog: Thanks

gslib/tests/test_mv.py
NickGoog: I'm not sure it's the case that this would work before the PR.
This also copies to a non-existing destination dir (it copies to bucket).
But will specifically add a test copying to a non-existing bucket subdir.

gslib/utils/copy_helper.py
NickGoog: Don't want to merge this PR with fixes for the `..` issue @rrauber is handling.

gslib/tests/test_cp.py
dilipped: bucket will get treated as an existing dir.

gslib/utils/copy_helper.py
dilipped: That will lead to differences in the behavior for `..` upload case between gsutil and gcloud. Also, I think Ross is going to simply skip downloading any file that has `..` in the object name, which seems to be different than the upload case anyway.

gslib/tests/test_mv.py
dilipped: Bucket gets treated as an existing dir, since a bucket always exists if it is created.

gslib/utils/copy_helper.py
dilipped: This is not addressed?

gslib/tests/test_mv.py
```diff
@@ -119,6 +119,80 @@ def test_move_dir_to_bucket(self):
...
+  def test_move_dirs_with_duplicate_file_names_to_bucket(self):
+    """Tests moving multiple top-level items to a bucket."""
+    bucket_uri = self.CreateBucket()
+    dir_path = self.CreateTempDir(test_files=[
+        ('dir1', 'file.txt'),
+        ('dir2', 'file.txt'),
+    ])
+
+    self.RunGsUtil(['mv', dir_path + '/*', suri(bucket_uri)])
+    # Expect two "gs://bucket/dir1/file.txt" and "gs://bucket/dir2/file.txt"
```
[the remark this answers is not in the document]
NickGoog: Adds a lot of overhead and checking the expected number of result items has been sufficient for the other tests in this file.

gslib/tests/test_mv.py
```diff
@@ -119,6 +119,80 @@ def test_move_dir_to_bucket(self):
...
+                      object_name='dir2/file.txt',
+                      contents=b'data')
+    self.AssertNObjectsInBucket(bucket1_uri, 2)
+    bucket2_uri = self.CreateBucket()
+    self.AssertNObjectsInBucket(bucket2_uri, 0)
+
+    self.RunGsUtil(['mv', suri(bucket1_uri, '*'), suri(bucket2_uri, 'dir')])
+
+    self.AssertNObjectsInBucket(bucket1_uri, 0)
+    self.AssertNObjectsInBucket(bucket2_uri, 2)
```
dilipped, line 156: For this test, I think it's important to check the destination objects path. e.g how can we tell if the destination object is gs://bucket/dir/dir1/file.txt or gs://bucket/dir1/file1.txt

gslib/tests/test_mv.py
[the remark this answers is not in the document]
dilipped: Are you talking about the overhead in terms of code? In that case, you can have a helper method. But anyway, I have left a comment for test_move_bucket_objects_with_duplicate_names_to_bucket_subdir where such kind of checking is required. So if you create such a helper, you might be able to reuse it here.

gslib/tests/test_mv.py
NickGoog: This is the focus of `test_naming.py`, specifically `testRecursiveCopyObjsAndFilesToNonExistentBucketSubDir`

dilipped: Agreed, but there is nothing in this test module that tells us that naming gets called. Plus, it's a less toll on whoever is not familiar with the behavior. Tests are great for documenting the behaviors and having the expected output checked will make it easier to understand how the mv behavior is meant to work for complicated naming behavior like these. Also, we check for the output for the cp tests anyway, why not do the same here?

NickGoog: >> but there is nothing in this test module that tells us that naming gets called

Not sure what you mean here. E2E tests are normally higher level than making sure underlying functions get called.

>> having the expected output checked will make it easier to understand how the mv behavior is meant to work

Sure, but I'd rather describe what's happening with a code comment than adding `ls` API calls and breaking the pattern of the rest of the file.

>> we check for the output for the cp tests anyway, why not do the same here

The `mv` tests are largely duplicates of functionality  tested in `test_cp.py`. The original authors probably meant for this to be lighter weight because (1) E2E tests are costly to run and (2) `test_cp.py` already does more detailed output analysis. 

I technically didn't need to add anything to `test_mv.py` because we already get  coverage from `test_cp.py`, but I wanted to be thorough. It's pushing it too far to break the lightweight testing pattern of the rest of the file, and I'd rather delete all the tests I added here.

gslib/tests/test_mv.py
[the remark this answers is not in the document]
NickGoog: ACK

gslib/tests/test_mv.py
dilipped: > Not sure what you mean here. E2E tests are normally higher level than making sure underlying functions get called.

You mentioned that test_naming is taking care of testing the behavior. I am saying that `test_naming` is testing the naming behavior, but how do we ensure that mv is actually using those naming behavior. We do not have any unittests that check if naming module is getting called.

> Sure, but I'd rather describe what's happening with a code comment than adding ls API calls and breaking the pattern of the rest of the file.

Regarding the API call, `self.AssertNObjectsInBucket(bucket2_uri, 2)` itself makes the `ls` call, so I don't think you are increasing the number of API calls. Also, I don't think we are breaking any pattern, instead we would be enhancing the quality of the test. The key thing that this test is looking for is the behavior of trying to move multiple directories with same file underneath. For this test, I think the naming of the destination file is important and hence it should be tested. That's the whole point of this PR, to avoid naming conflict. In the current state, all we are testing is that the file is getting moved over. There is nothing here that tells me that the naming was handled correctly for mv command specifically.

> The mv tests are largely duplicates of functionality tested in test_cp.py. The original authors probably meant for this to be lighter weight because (1) E2E tests are costly to run and 

Like I said above, checking for count involves an ls call anyway. So I don't think it was light weight in the first place.

> (2) test_cp.py already does more detailed output analysis.

Right. And if we want to rely on that behavior, we should really ensure that mv calls the cp behavior i.e have some unittests that check if copy_helper etc gets called. Basically if A depends on B, and we have tested B, we can be sure that A is tested by ensuring that A is calling B correctly. But in this case, we do not test if mv is calling cp. So there are two ways to test this, either via E2E test or via unittest test and ensuring that correct calls were made to cp. We have chosen to test this via E2E test, and if we are doing that, we should ensure that the end result is correct and not rely on inner knowledge that mv relies on cp.

> I technically didn't need to add anything to test_mv.py because we already get coverage from test_cp.py, but I wanted to be thorough. It's pushing it too far to break the lightweight testing pattern of the rest of the file, and I'd rather delete all the tests I added here.

I appreciate you writing these tests, but like I said, there is no lightweight testing pattern that we are breaking. The number of API calls remain the same with what I am asking here. 

Regarding deleting all these tests, we need tests for mv because that's what led to having this PR in the first place. We need to ensure the behavior of mv in such cases. Remember that mv had a separate if condition originally which we removed. But, like I said, you need not have e2e tests if you can ensure using unittests that mv will call the cp behavior. That is what would actually make the tests lightweight.

All this is my opinion of course, and feel free to bring in a second reviewer. I am happy to approve this if others feel what I am proposing is not really needed.

gslib/utils/copy_helper.py
NickGoog: Done

gslib/tests/test_mv.py
NickGoog: Lightweight as in amount of code and smaller scope. Sorry, I meant to remove the part about using extra resources b/c extra API calls are not being made.

But I think it's definitely breaking a pattern if we start adding `ls` output analysis in a file where all similar tests do not have it (with one exception `test_stdin_args`). I'm not a fan of rewriting redundant code to what I wrote for `test_cp.py`, especially if it breaks a pattern.

NickGoog: > I think we should replace this test with a unit test that simply checks that the appropriate methods/functions are called

That sounds like `test_naming` unless you mean mocking functions and calling `assert_called_with(...)` (I don't suggest that b/c it's usually better practice to test a behavior rather than an underlying mechanism). I can try converting these to `test_naming` unit tests if you want.

> There is not really a good way for me to look at all these tests in mv and say with confidence that this behavior has been already tested. In such cases, the test here should ensure that the behavior is tested.

The behavior being tested is that `mv`'ing 2 files results in 2 files, not 1, which would indicate improper merging on name duplicate. I can add more clarification in code if you want.

> there is not really a reason to follow a pattern IMO

I think the pattern's purpose is to keep tests simple and short, which can be done b/c the more complex behavior is tested in `test_cp.py`.

> I don't think adding 3-4 lines of code is going to make the test heavy.

We have to combine the "count objects in bucket" logic from `AssertNObjectsInBucket` with string matching on the output paths. It will be more than 4 lines but not a lot if we implement a `assert_paths` parameter for `AssertNObjectsInBucket`. But this is unnecessary if you want the `test_mv` tests converted to `test_naming` unit tests (there are other `mv` tests in `test_naming`).

gslib/tests/test_naming.py
```diff
@@ -975,23 +975,95 @@ def testRecursiveCopyObjsAndFilesToExistingBucketSubDir(self):
       ])
       self.assertEqual(expected, actual)
 
-  # @SequentialAndParallelTransfer
+  def testRecursiveCopyFileToExistingBucketSubDirInvalidSourceParent(self):
+    """Tests recursive copy of invalid path file to existing bucket subdir."""
+    src_dir = self.CreateTempDir(test_files=[('nested', 'f0')])
+    dst_bucket_uri = self.CreateBucket(test_objects=['dst_subdir/existing_obj'])
+
+    for relative_path_symbols in ['.', '.' + os.sep]:
```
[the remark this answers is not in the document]
NickGoog: `./` is two characters. I'll change it to `relative_path_string`.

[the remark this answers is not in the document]
NickGoog: Done, good call

gslib/utils/copy_helper.py
```diff
@@ -635,26 +679,24 @@ def ConstructDstUrl(src_url,
     exp_dst_url = StorageUrlFromString(
         '%s%s' % (exp_dst_url.url_string, exp_dst_url.delim))
 
+  src_url_is_valid_parent = _IsUrlValidParentDir(src_url)
+  # raise Exception('{} {} {} {}'.format(src_url, exp_src_url, src_url_is_valid_parent, has_multiple_top_level_srcs))
```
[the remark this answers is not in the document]
NickGoog: Fixed

gslib/utils/copy_helper.py
```diff
@@ -663,58 +705,53 @@ def ConstructDstUrl(src_url,
...
-                              exp_src_url_tail.strip('/'))
-
-  elif src_url_names_container and (exp_dst_url.IsCloudUrl() or
-                                    exp_dst_url.IsDirectory()):
-    # Case 2.  Container copy to a destination other than a file.
+  # Ignore the "multiple top-level sources" rule if using double wildcard **
+  # because that treats all files as top-level, in which case the user doesn't
+  # want to preserve directories.
+  preserve_src_top_level_dirs = ('**' not in src_url.versionless_url_string and
+                                 src_url_is_valid_parent and
```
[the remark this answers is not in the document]
NickGoog: Technically you could change it to:
```
preserve_src_top_level_dirs = ('**' not in src_url.versionless_url_string and (
                                                   has_multiple_top_level_srcs or
                                                   (src_url_is_valid_parent and have_existing_dest_subdir)))
```

But this seemed needlessly less robust and harder to reason about when I wrote it.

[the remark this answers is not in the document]
dilipped: I meant dropping `src_url_is_valid_parent` altogether. Do we get the same result without it?

gslib/tests/test_mv.py
dilipped: I was suggesting `assert_called_ with`. The reason being, you mentioned that `test_cp` is already testing for the paths. So if that is the case and if we truly want to avoid the redundancy to make it lightweight, we can simply have a test that checks if mv is calling cp behavior. My point being, there is nothing here or in other tests for mv that tells me that mv calls copy_helper or other naming helpers as expected.  This is the knowledge that you and I have, but it does not reflect in the test code. Of course we can add a comment explaining this, but the more self documenting code we have, the better.

> it's usually better practice to test a behavior rather than an underlying mechanism

Testing the behavior is better, but you have to ensure that you are testing it correctly. If you say that you will partially test this because it is already tested by a dependent module, you kinda have to test that you are calling that module correctly.

But anyway, I don't want to block this PR for this. I will leave it for the second reviewer or other experts to break the tie as I feel we both have been repeating the same points and it's not leading anywhere. Will approve the PR once the other open comment is resolved.

gslib/utils/copy_helper.py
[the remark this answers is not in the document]
dilipped: Nevermind, I think it is required, else you will see .. in the destination path for single top level source.

dilipped (approved):
There is one open question related to the tests where we need an opinion from another reviewer. Otherwise, looks good.

gslib/tests/test_mv.py
NickGoog: As per my last comment, I'm fine trying to convert the `test_mv.py` E2E tests to `test_naming.py` unit tests but just wanted to check if you have objections first

dilipped: I was not suggesting converting this to `test_naming`, but instead adding a unit test in test_mv itself that verifies that naming or copy_helper is called as expected. Adding this to `test_naming` is fine, but I think we would again have the same issue, i.e nothing in mv tests that we are correctly using the "naming" behavior.

dilipped: Also, this is a shortcoming in test_mv itself. I feel in general there should be a test that ensures it calls copy_helper or the cp functionality correctly. So it's not specific to the test you have here.

NickGoog: Done, left one test as E2E b/c the unit test framework was having trouble with simulating downloads following the `test_naming.py` pattern. The one left as E2E does check output paths.

dilipped: LGTM

dilipped: Also, thanks for adding the tests!

PR #1467, opened 2022-02-18
Fix rm, rb and mb tests that started failing because of backend changes

dilipped:


gslib/tests/test_mb.py
```diff
@@ -288,8 +288,10 @@ def test_create_with_invalid_placement_flag_raises_error(self):
          suri(bucket_uri)],
         return_stderr=True,
         expected_status=1)
-    self.assertIn('BadRequestException: 400 Invalid custom placement config',
-                  stderr)
+    self.assertRegex(
+        stderr, r'.*BadRequestException: 400 (Invalid custom placement config|'
```
NickGoog, line 292: This long regex seems a little fragile if more backend changes are made

NickGoog (approved):

gslib/tests/test_mb.py
dilipped: This is going to go away anyway, because we are going to drop the placement flag, so I am not too worried about it.

PR #1472, opened 2022-03-02
Add the shim flags only if requested

dilipped:
I was adding the use_gcloud_storage=False flag everywhere. Now adding it only if it was requested from the top level. Should not have any effect on the tests, just makes it cleaner.

### DIFF
--- a/gslib/tests/testcase/integration_testcase.py
+++ b/gslib/tests/testcase/integration_testcase.py
@@ -1006,12 +1006,12 @@ class GsUtilIntegrationTestCase(base.GsUtilTestCase):
       use_gcloud_storage = False
     else:
       use_gcloud_storage = config.getbool('GSUtil', 'use_gcloud_storage', False)
-    gcloud_storage_setting = [
-        '-o',
-        'GSUtil:use_gcloud_storage={}'.format(use_gcloud_storage),
-        '-o',
-        'GSUtil:hidden_shim_mode=no_fallback',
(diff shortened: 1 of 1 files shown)


thomasmaclean:
Kokoro is now failing with a number of errors like the following:

```
1/1 finished - E[0] F[0] s[0] - TestCompose.test_compose_with_kms_encryp - E
======================================================================
ERROR: test_compose_with_kms_encryption (gslib.tests.test_compose.TestCompose)
Tests composing encrypted objects.
File "/tmpfs/src/github/src/gsutil/gslib/tests/test_compose.py", line 252, in test_compose_with_kms_encryption
self.AssertObjectUsesCMEK(obj_suri, key_fqn)
File "/tmpfs/src/github/src/gsutil/gslib/tests/testcase/integration_testcase.py", line 428, in AssertObjectUsesCMEK
force_gsutil=True)
File "/tmpfs/src/github/src/gsutil/gslib/tests/testcase/integration_testcase.py", line 1018, in RunGsUtil
] + gcloud_storage_setting + cmd
UnboundLocalError: local variable 'gcloud_storage_setting' referenced before assignment
```

thomasmaclean (changes requested):
Fix Kokoro issues (see prior comment)

dilipped:
Oops. I think I was overthinking. What I had earlier was correct and this is PR is not really needed. Closing it.

PR #1474, opened 2022-03-03
Skip test for pip package

dilipped:


gslib/tests/test_gsutil.py
```diff
@@ -18,14 +18,21 @@
 from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
+from ast import Is
```
thomasmaclean, line 21: What was Is imported for here?

dilipped: Oops. Not sure how this got added. Removed

PR #1480, opened 2022-03-10
Custom dual region - Drop placement flag as multiple locations can now be provided via -l flag

dilipped:
The --placement flag  and the PlacementConfig support will be dropped. Dual regions can be now specified via the location `-l` flag.


gslib/commands/mb.py
```diff
@@ -158,12 +157,6 @@
                          "enforced", objects in this bucket cannot be made
                          publicly accessible. Default is "inherited".
```
dilipped, line 159: @rebecca-pete Can you please take a look at the help text for -l command abobe and let me know if we need to specify anything about dual regions there. My thought was that we probably don't as the link in the help text https://cloud.google.com/storage/docs/locations  would already cover that information.

NickGoog:
Looks good, waiting for Koko

dilipped:
@NickGoog Koko is happy!

NickGoog (approved):

PR #1482, opened 2022-03-17
Run Python 3.5 tests with 3.5.2

dilipped:


test/ci/kokoro/run_integ_tests.sh
```diff
@@ -42,6 +42,14 @@ if [[ $KOKORO_JOB_NAME =~ "macos" ]]; then
 fi
 
 function latest_python_release {
```
thomasmaclean, line 44: Can we change this to "preferred_python_release" to reduce later confusion? There are comments further down clarifying that they're getting the "latest" available, so I think everything will still semantically make sense after the change.

thomasmaclean (changes requested):

test/ci/kokoro/run_integ_tests.sh
dilipped: Good point! Done.

dilipped: Oops, missed updating the calling function

dilipped: Now it's done.

thomasmaclean (approved):

PR #1485, opened 2022-03-24
Improve the comment about -a flag of du command

### LINKED ISSUE #1484: Comments about `du` is not clear
Current generated doc URL: https://cloud.google.com/storage/docs/gsutil/commands/du

Describe:
```
The du command displays the amount of space in bytes used up by the objects in a bucket, subdirectory, or project. The syntax emulates the Linux du -b command, which reports the disk usage of files and subdirectories. For example, the following command reports the total space used by all objects and subdirectories under gs://your-bucket/dir: gsutil du -s -a gs://your-bucket/dir

...
-a  Includes both live and noncurrent object versions. Also prints the generation and metageneration number for each listed object.
...
```

But it doesn't mention that without -a flag, only live objects are counted. This can surprise people if they only get live objects amount in this way and finally get charged a lot due to large amount of noncurrent object.

JamesRaynor67:
This is the AI for b/226394945

Fixes #1484


gslib/commands/du.py
```diff
@@ -62,7 +62,8 @@
               can use this to make the output machine-readable.
 
   -a          Includes both live and noncurrent object versions. Also prints the
-              generation and metageneration number for each listed object.
+              generation and metageneration number for each listed object. Without
+              specifying this flag, only live object versions are included.
```
cbonnie, line 66: nit: Could we change the language to "If this flag is not specified, only live object versions are included"? Otherwise looks great.

JamesRaynor67: Done. I accidentally pulled from upstream and had to force push to revert it. Please kindly check the final commit and file change.

NickGoog: LGTM

PR #1490, opened 2022-04-01
Add cp -v and -n flags for shim translation.

NickGoog:
Prep relevant tests for gcloud storage parity testing.

### DIFF
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -758,9 +758,11 @@ class CpCommand(Command):
       gcloud_command='alpha storage cp',
       flag_map={
           '-e': GcloudStorageFlag('--ignore-symlinks'),
+          '-n': GcloudStorageFlag('--no-clobber'),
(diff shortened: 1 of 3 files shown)


gslib/commands/cp.py
```diff
@@ -758,9 +758,11 @@ class CpCommand(Command):
       gcloud_command='alpha storage cp',
       flag_map={
           '-e': GcloudStorageFlag('--ignore-symlinks'),
+          '-n': GcloudStorageFlag('--no-clobber'),
```
dilipped, line 761: Looks good. Can you please update this test https://github.com/GoogleCloudPlatform/gsutil/blob/3e0dbf6564038dda51c4c6b065c0a45052f66c35/gslib/tests/test_cp.py#L4728  to cover the new flags, or even better, just have a separate unit test in test_cp.py that verifies this map.

NickGoog: Done

dilipped (approved):

PR #1494, opened 2022-04-08
Add shim support for defstorageclass command

dilipped:
See test parity results here: cl/440366504

### DIFF
--- a/gslib/utils/shim_util.py
+++ b/gslib/utils/shim_util.py
@@ -265,6 +265,8 @@ class GcloudStorageCommandMixin(object):
     args = []
     if isinstance(gcloud_storage_map.gcloud_command, str):
       args = gcloud_storage_map.gcloud_command.split()
+    elif isinstance(gcloud_storage_map.gcloud_command, list):
+      args.extend(gcloud_storage_map.gcloud_command)
     elif isinstance(gcloud_storage_map.gcloud_command, dict):
       # If a command has sub-commands, e.g gsutil pap set, gsutil pap get.
       # All the flags mapping must be present in the subcommand's map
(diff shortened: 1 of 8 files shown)


gslib/utils/shim_util.py
```diff
@@ -265,6 +265,8 @@ def _get_gcloud_storage_args(self, sub_opts, gsutil_args, gcloud_storage_map):
     args = []
     if isinstance(gcloud_storage_map.gcloud_command, str):
       args = gcloud_storage_map.gcloud_command.split()
+    elif isinstance(gcloud_storage_map.gcloud_command, list):
```
NickGoog, line 268: I think you have the opportunity to merge some of these blocks

dilipped: You mean get rid of str and just use a list in the map? I can do that but intentionally left it as it is because the list is really useful for bucket booleans commands only. For other commands a string command is easier to read.

NickGoog: I'm used to inputting commands and their args as lists for the `subprocess` module and how are tests work, so I don't think it's that tricky, but I'm biased. I saw this as an example of a `str` gcloud command:
https://github.com/GoogleCloudPlatform/gsutil/blob/47382bbb3b1216c66c9ab2792936fe9cce94c8cf/gslib/commands/ls.py#L335

And I don't think it would look bad as a list: `['alpha', 'storage', 'ls']`

Zooming out, I don't like using data type to decide what logic to use. "Sort by data type" can lead to passing around lots of data types with the same variable name, which increases general complexity by a lot. I would almost prefer we use a `dict` for everything where one of the `dict` values is an `enum` key saying what the format is (if we really must have a simple version for bucket booleans, `ls`, etc).

dilipped: I don't mind changing it to a list. I was kinda on a fence myself about doing it or not. However, I think using an enum is an overkill for this case. I can change the str to list everywhere.

NickGoog (approved):

PR #1497, opened 2022-04-12
Shim translation and test parity for cp manifest flag.

NickGoog:


gslib/tests/test_cp.py
```diff
@@ -2094,10 +2094,16 @@ def _ManifestUpload(self, file_name, object_name, manifest_name):
...
-        'Source Size', 'Bytes Transferred', 'Result', 'Description'
-    ]
+    if self._use_gcloud_storage:
+      expected_headers = [
+          'Source', 'Destination', 'Start', 'End', 'Md5', 'Source Size',
+          'Bytes Transferred', 'Result', 'Description'
+      ]
+    else:
+      expected_headers = [
+          'Source', 'Destination', 'Start', 'End', 'Md5', 'UploadId',
```
rrauber, line 2104: Just making sure: was dropping the UploadId column approved? I may have missed that discussion. It seems like even if it's empty, we should keep it, in case scripts are indexing on column numbers.

NickGoog: Yep, we're dropping it. Third-party scripts are the only concern. gsutil and gcloud storage both index based on column key not integer.

rrauber: I think third party scripts are a concern though right? Especially for the shim

NickGoog: Okay, I'll have to build out some infra before getting back to this.

NickGoog:
Merging b/c this is basically one non-controversial line

PR #1501, opened 2022-04-18
Include Third Party Identification as a Credential Type supporting various use cases

ScruffyProdigy:


gslib/addlhelp/creds.py
```diff
@@ -155,8 +155,17 @@
 
     In order to impersonate, your original credentials need to be granted
     roles/iam.serviceAccountTokenCreator on the target service account.
-    For more information see 
+    For more information see
     https://cloud.google.com/iam/docs/creating-short-lived-service-account-credentials
+
+  Third Party Identification:
```
lsirac, line 161: Change this to: 

External Account Credentials (Workload Identity Federation)

Using workload identity federation, you can access Google Cloud resources from Amazon Web Services (AWS), Microsoft Azure or any identity provider that supports OpenID Connect (OIDC) or SAML 2.0.

For more information see https://cloud.google.com/iam/docs/using-workload-identity-federation

lsirac (approved):

PR #1519, opened 2022-05-27
Add shim support for cp ACL flags

NickGoog:
Test parity waiting for port of "acl" command.

Manually tested.


NickGoog:
Probably doesn't need to be held up by review~

PR #1525, opened 2022-06-15
Bring back placement flag for custom dual regions

thomasmaclean:
Reverts GoogleCloudPlatform/gsutil#1480 and keeps some of the modifications to the -l flag for mb.


thomasmaclean:
Kokoro failures are expected at this point since we're waiting on production updates.

PR #1528, opened 2022-06-21
Added flush to cat_helper.py

ejeselsohn:
The gsutil cat -h flag when used on multiple objects which are small, first prints the headers of all the objects the user entered and then prints the contents of all the objects the user entered.
The reason the gsutil cat -h command with multiple small files does not work properly is because the headers are written to stderr whereas the object contents are written to stdout. Since stderr does not have a buffer and stdout does, with small files, the buffer is not cleared fast enough and both headers are printed before the first content is printed. By flushing the buffer immediately, the header of the first object is printed followed by the content, header of the second object followed by the content, etc..

### DIFF
--- a/gslib/tests/test_cat.py
+++ b/gslib/tests/test_cat.py
@@ -19,6 +19,8 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
+import sys
+
 from gslib.cs_api_map import ApiSelector
 from gslib.exception import NO_URLS_MATCHED_TARGET
 import gslib.tests.testcase as testcase
@@ -29,6 +31,9 @@ from gslib.tests.util import RUN_S3_TESTS
 from gslib.tests.util import SetBotoConfigForTest
 from gslib.tests.util import TEST_ENCRYPTION_KEY1
 from gslib.tests.util import unittest
+from gslib.utils import cat_helper
+
+from unittest import mock
 
 
 class TestCat(testcase.GsUtilIntegrationTestCase):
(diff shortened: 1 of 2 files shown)


rrauber (changes requested):
Could you add a test that covers the `flush` call? I think a unit test kind of like [this one](https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/tests/test_cp.py#L4902), which mocks out cat_out_fd and ensures `flush` is called after data is written would be a good fit!

Also, could you add a bit more information about the issue you're trying to solve to the title/description? It's helpful when looking through the history.

rrauber (approved):

gslib/tests/test_cat.py
```diff
@@ -162,3 +167,66 @@ def test_cat_encrypted_object(self):
...
+                  start_byte=0,
+                  end_byte=None,
+                  object_size=10,
+                  generation=None,
+                  decryption_tuple=None,
+                  provider='gs'),
+        mock.call()
+    ]
+    # Needed to do these two checks because the object list order differs
+    # between Windows OS and Python Version 3.5.
```
NickGoog, line 228: It doesn't really make sense to compare an OS to a Python version. This should be something like `Windows OS Python Version X` to `Other OS's Python Version Y` or `Windows OS Python 3.5 to later Python versions on Windows`

gslib/tests/test_cat.py
```diff
@@ -162,3 +167,66 @@ def test_cat_encrypted_object(self):
...
+
+  def test_cat_helper_runs_flush(self):
+    command_mock = mock.Mock()
+    cat_helper_mock = cat_helper.CatHelper(command_obj=command_mock)
+
+    object_contents = '0123456789'
+    bucket_uri = self.CreateBucket(bucket_name='bucket',
+                                   provider=self.default_provider)
+    obj = self.CreateObject(bucket_uri=bucket_uri,
+                            object_name='foo',
```
NickGoog, line 183: `foo` and `foo1` isn't as descriptive as something like `foo1` and `foo2`. If you have `foo` without a number, people may assume there's just one of them.

gslib/tests/test_cat.py
```diff
@@ -162,3 +167,66 @@ def test_cat_encrypted_object(self):
       stdout = self.RunGsUtil(
           ['cat', '-r 1-3', suri(object_uri)], return_stdout=True)
       self.assertEqual(stdout, '123')
+
+
+class TestCatHelper(testcase.GsUtilUnitTestCase):
+  """Unit tests for cat helper."""
+
+  def test_cat_helper_runs_flush(self):
+    command_mock = mock.Mock()
```
NickGoog, line 176: Might be more clear to call this `cat_command_mock`

PR #1531, opened 2022-06-23
Updated the cat -r command error behavior

ejeselsohn:
Under the old cat command, if the user entered a gsutil cat -r x-y gs://object and the starting range is out of bounds, an error was thrown. Under the new behavior, nothing is outputted and no error is thrown, just like Python's behavior.
Additionally, under the old gsutil cat code, when a user entered `gsutil cat -r - gs://object' a "CommandException" was thrown. Under the new code, the whole object is printed, just like Python's default behavior.

### DIFF
--- a/gslib/commands/cat.py
+++ b/gslib/commands/cat.py
@@ -131,7 +131,7 @@ class CatCommand(Command):
           # This if statement ensures the full object is returned
           # instead of throwing a CommandException.
           if request_range == '-':
-            break
+            continue
           range_matcher = re.compile(
               '^(?P<start>[0-9]+)-(?P<end>[0-9]*)$|^(?P<endslice>-[0-9]+)$')
           range_match = range_matcher.match(request_range)
--- a/gslib/commands/cat.py
+++ b/gslib/commands/cat.py
@@ -131,6 +131,10 @@ class CatCommand(Command):
           range_matcher = re.compile(
               '^(?P<start>[0-9]+)-(?P<end>[0-9]*)$|^(?P<endslice>-[0-9]+)$')
           range_match = range_matcher.match(request_range)
+          # This if statement enures the full object is returned
+          # instead of throwing a CommandException.
+          if request_range == '-':
+            break
           if not range_match:
             raise CommandException('Invalid range (%s)' % request_range)
           if range_match.group('start'):
--- a/gslib/commands/cat.py
+++ b/gslib/commands/cat.py
@@ -128,13 +128,13 @@ class CatCommand(Command):
           show_header = True
         elif o == '-r':
           request_range = a.strip()
-          range_matcher = re.compile(
-              '^(?P<start>[0-9]+)-(?P<end>[0-9]*)$|^(?P<endslice>-[0-9]+)$')
-          range_match = range_matcher.match(request_range)
-          # This if statement enures the full object is returned
+          # This if statement ensures the full object is returned
           # instead of throwing a CommandException.
           if request_range == '-':
             break
+          range_matcher = re.compile(
+              '^(?P<start>[0-9]+)-(?P<end>[0-9]*)$|^(?P<endslice>-[0-9]+)$')
+          range_match = range_matcher.match(request_range)
           if not range_match:
             raise CommandException('Invalid range (%s)' % request_range)
           if range_match.group('start'):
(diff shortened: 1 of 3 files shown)


gslib/commands/cat.py
```diff
@@ -131,6 +131,10 @@ def RunCommand(self):
           range_matcher = re.compile(
               '^(?P<start>[0-9]+)-(?P<end>[0-9]*)$|^(?P<endslice>-[0-9]+)$')
           range_match = range_matcher.match(request_range)
+          # This if statement enures the full object is returned
```
NickGoog, line 134: "ensures"

gslib/commands/cat.py
```diff
@@ -131,6 +131,10 @@ def RunCommand(self):
...
+          # instead of throwing a CommandException.
+          if request_range == '-':
```
NickGoog, line 136: Can this `if` block go above the `range_matcher` logic? `range_matcher` doesn't really matter if we can return earlier.

gslib/utils/cat_helper.py
```diff
@@ -137,6 +137,10 @@ def CatUrlStrings(self,
               printed_one = True
             cat_object = blr.root_object
             storage_url = StorageUrlFromString(blr.url_string)
```
NickGoog, line 139: This line can be moved below the new `if` block because it doesn't matter if the `if` block triggers a `return`.

rrauber (approved):

gslib/commands/cat.py
```diff
@@ -128,6 +128,10 @@ def RunCommand(self):
           show_header = True
         elif o == '-r':
           request_range = a.strip()
+          # This if statement ensures the full object is returned
+          # instead of throwing a CommandException.
+          if request_range == '-':
+            break
```
rrauber, line 134: Did you mean `continue` instead of `break` here? What happens if both the -h and -r flags are used?

gslib/utils/cat_helper.py
```diff
@@ -136,6 +136,10 @@ def CatUrlStrings(self,
               print('==> %s <==' % blr)
               printed_one = True
             cat_object = blr.root_object
+            # This if statement ensures nothing is ouputted and no error
+            # is thrown if the user enters an out of bounds range for the object.
+            if start_byte >= cat_object.size:
+              return 0
```
rrauber, line 142: There's a `finally` clause here which sets `sys.stdout` to `old_stdout`, which would typically run before returning. I wonder if that could be related to some of the timeouts you're seeing.

rrauber (changes requested):

gslib/commands/cat.py
ejeselsohn: yup, you are right. nice find!

gslib/utils/cat_helper.py
NickGoog: Nice catch!

gslib/utils/cat_helper.py
```diff
@@ -136,6 +136,13 @@ def CatUrlStrings(self,
               print('==> %s <==' % blr)
               printed_one = True
             cat_object = blr.root_object
+            # This if statement ensures nothing is ouputted and no error
+            # is thrown if the user enters an out of bounds range for the object.
+            if getattr(
+                cat_object, 'size', None
```
NickGoog, line 142: What should the correct behavior here be if `cat_object.size` is `None`? In what scenario would that happen?

You might be able to replace this if statement with something like:
```
if getattr(cat_object, 'size', float('inf')) <= start_byte:
```

gslib/utils/cat_helper.py
NickGoog: Wait, I think resetting `sys.stdout` isn't necessary:
```
def f():
    try:
        return 'returning'
    finally:
        print('finally')

print(f())

-> finally
-> returning
```

ejeselsohn: Done.

gslib/utils/cat_helper.py
ejeselsohn: done.

rrauber (approved):

NickGoog:
Restarting Kokoro

NickGoog (approved):

PR #1537, opened 2022-06-29
Shim rewrite command.

NickGoog:


NickGoog:
Went through one XML and one JSON test for every OS in Kokoro and confirmed all errors are flakes or KMS 409-related.

PR #1539, opened 2022-07-01
Achieve custom header test parity.

rrauber:


gslib/tests/test_cp.py
```diff
@@ -1153,14 +1153,21 @@ def test_other_headers(self):
 
     stdout = self.RunGsUtil(['ls', '-L', dst_uri], return_stdout=True)
     self.assertRegex(stdout, r'Cache-Control\s*:\s*public,max-age=12')
-    self.assertRegex(stdout, r'Metadata:\s*1:\s*abcd')
+
+    if self._use_gcloud_storage:
+      self.assertRegex(stdout, r'"key":\s* "1",\n\s*"value":\s*"abcd"')
```
NickGoog, line 1158: This part may not be necessary because the output of `ls -L` should match now.

NickGoog (commented):
Overall, looks good. Let's submit the 409 fix first and then see how Kokoro reacts to it.

gslib/tests/test_cp.py
```diff
@@ -1189,14 +1200,17 @@ def test_request_reason_header_persists_multiple_requests_json(self):
...
-    self.assertRegex(stderr, reason_regex)
+    if self._use_gcloud_storage:
+      reason_regex = r'X-Goog-Request-Reason\': b\'b/this_is_env_reason'
+    else:
+      reason_regex = r'x-goog-request-reason\': \'b/this_is_env_reason'
+
+    self.assertRegex(
+        stderr,
+        # PUT follows GET request. Both need the request-reason header.
+        r'GET[\s\S]*' + reason_regex + r'[\s\S]*PUT[\s\S]*' + reason_regex)
```
NickGoog, line 1213: Be careful that there are no other `GET` or `PUT` requests in `stderr` that might cause this to pass

rrauber: Definitely! There aren't any others as it stands. It's a bit of a risk that this might change, but I think really what this test has to cover is that multiple requests use the reason header.

gslib/tests/test_cp.py
rrauber: Tried removing it and got the following error. It looks like `ls -L` is still using the old style. Do you know off hand if there's anything extra I need to enable to get gsutil style formatting? Just pulled from master and got gcloud at head.

AssertionError: Regex didn't match: 'Metadata:\\s*1:\\s*abcd' not found in '
gs://gsutil-test-test-other-headers-bucket-552b913a/foo:
  Creation time:               Fri, 05 Aug 2022 15:25:23 GMT
  Update time:                 Fri, 05 Aug 2022 15:25:23 GMT
  Storage class update time:   Fri, 05 Aug 2022 15:25:23 GMT
  Storage class:               STANDARD
  Cache-Control:               public,max-age=12
  Content-Language:            en
  Content-Length:              810
  Content-Type:                image/gif
  Metadata:
  [
    {
      "key": "1",
      "value": "abcd"
    }
  ]

NickGoog: I think you need both the latest gcloud and the latest gsutil commit

rrauber: Seems to work now!

NickGoog (approved):

PR #1541, opened 2022-07-11
Add retries with exponential backoff to the flow that authorizes the service agent to use CMEKs

rrauber:
This PR took a long journey to basically get back to where we started.

Recently we've been having an uptick in 409 errors in tests due to concurrent changes to permissions on KMS keys. These were occuring because gsutil's testing framework attempts to create keys and grant the service agent permission to use them in each test that makes use of those KMS keys. This is to ensure portability -- any person on any project can run a gsutil test that relies on the presence/appropriate configuration of certain keys and they will be set up correctly because of this. 

I considered three ways to cut down on permissions changes without sacrificing portability, each of which doesn't work due to how the test command is implemented, and how the tests are run in our kokoro jobs:
- Parallel tests are run by generating a list of tests, then creating subprocesses running the gsutil test command for each individual test: https://github.com/GoogleCloudPlatform/gsutil/blob/651253515666d324bd78fc0ddb73c977fe150d89/gslib/commands/test.py#L343.
- For our MacOS and Linux Kokoro builds, a single `gsutil test` command runs most of our test suite, but for Windows, each subcommand is tested by a different invocation of `gsutil test`, e.g. `gsutil test cp`, presumably due to parallel processing differences on that platform. Here is the loop that runs these various test commands: https://github.com/GoogleCloudPlatform/gsutil/blob/master/test/ci/kokoro/windows/run_integ_tests.ps1#L106

The three solutions were:
1. Caching the results of the authorize service agent helper function in the hope that doing so would prevent further calls if a key had already been authorized. This does not work because the actual calls happen in different subprocesses, so they are not able to share a cache.
2. Moving key authorization out of individual tests and into the logic for the test command itself. This does not work because of the fact that windows runs the `gsutil test` command many times, resulting in overall more calls to the permissions setting methods than the original solution.
3. Moving the authorization logic to setUpClass methods for only the commands that use them. This did not work because individual tests are run in different subprocesses, and setUpClass runs once per process.

What I ultimately ended up with was adding retries with exponential backoff to the pipeline which authorizes the service agent, and some light refactors to consolidate the existing code. These changes make things more stable in the meantime and prepare us for a better solution: adding functionality to the test command so that we can add something like `gsutil test setup-resources` to the beginning of our kokoro scripts and provide people an easy way to set up new projects. This will essentially achieve what (2) was trying to, but will avoid differences in how Windows runs the test suite by moving the authorization calls out of typical test commands.


NickGoog (approved):
Love it

gslib/commands/test.py
```diff
@@ -554,6 +554,16 @@ def RunCommand(self):
         elif o == '-u':
           tests.util.RUN_INTEGRATION_TESTS = False
 
+    if tests.util.RUN_INTEGRATION_TESTS:
+      try:
+        tests.util.AuthorizeProjectToUseTestingKmsKeys()
+      except Exception as error:
+        # Authorized keys are static resources which typically already have
+        # appropriate permissions. Most errors can be safely ignored.
+        logging.warning(
```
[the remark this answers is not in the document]
rrauber: Unfortunately it's hard to get it to repeat at every test. I added some logic to repeat it after everything has run though.

gslib/commands/test.py
```diff
@@ -554,6 +554,16 @@ def RunCommand(self):
         elif o == '-u':
           tests.util.RUN_INTEGRATION_TESTS = False
 
+    if tests.util.RUN_INTEGRATION_TESTS:
+      try:
+        tests.util.AuthorizeProjectToUseTestingKmsKeys()
+      except Exception as error:
+        # Authorized keys are static resources which typically already have
```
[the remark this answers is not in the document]
rrauber: Thanks! Changed the comment a bit to reflect what I was really after: transient errors would block the entire suite from running, which could get really annoying. It is also possible for someone running these tests to not have permission to give permission to their service agent to use kms keys, in which case throwing an exception here would prevent them from running any tests. There are a lot of tests that can be run without touching kms keys so this could block legitimate uses of the test suite.

gslib/commands/test.py
```diff
@@ -554,6 +554,19 @@ def RunCommand(self):
...
+    if tests.util.RUN_INTEGRATION_TESTS:
+      try:
+        tests.util.AuthorizeProjectToUseTestingKmsKeys()
+        service_agent_authorized_successfully = True
+      except Exception as error:
+        # Authorized keys are static resources that typically already have
+        # appropriate permissions. Demoting errors to warnings means transient
+        # errors will not block running the whole suite of tests.
+        logging.warning(
+            'Failed to authorize the service agent to use KMS keys.')
```
NickGoog, line 566: This warning still sounds like a non-transient error. Do we even need it with the new logging below?

rrauber: Removed the warning as part of a change in approach.

gslib/tests/testcase/integration_testcase.py
```diff
@@ -1001,79 +1001,13 @@ def RunGsUtil(self,
...
-
-    if six.PY3:
-      # TODO(b/135936279): Make this number configurable in .boto
-      comm_kwargs['timeout'] = 180
-    else:
-      timer = threading.Timer(180, Kill)
-      timer.start()
-
-    c_out = p.communicate(**comm_kwargs)
+    processed_command = util.GetGsutilCommand(cmd, force_gsutil=force_gsutil)
```
NickGoog, line 1004: What does "processed" mean in this case? If it just means some string generation has happened, might be a bit of a confusing word to put next to actual processes

gslib/tests/util.py
```diff
@@ -172,6 +269,45 @@ def GetUserGroups():
 POSIX_INSUFFICIENT_ACCESS_ERROR = 'Insufficient access with uid/gid/mode for %s'
 
 
+class KmsTestingResources(object):
```
NickGoog, line 272: If we put this and the below `Authorize` util in their own file, it might be more organized and isolate the dependencies of a dedicated command for test resource setup more (if we don't just use the `kms` command).

gslib/tests/util.py
```diff
@@ -172,6 +269,45 @@ def GetUserGroups():
...
+  # deleted). Tests should fill in the %d entries with a digit between 0 and 9.
+  MUTABLE_KEY_NAME_TEMPLATE = 'cryptokey-for-gsutil-integration-tests-%d%d%d'
+
+
+def AuthorizeProjectToUseTestingKmsKey(
+    key_name=KmsTestingResources.CONSTANT_KEY_NAME):
+  """Ensures test keys exist and that the service agent is authorized."""
+  kms_api = KmsApi(logging.getLogger())
+
+  keyring_fqn = kms_api.CreateKeyRing(
```
NickGoog, line 299: Might want to explain what `fqn` means somewhere

NickGoog (approved):

gslib/tests/util.py
rrauber: fqn is now fully qualified.

gslib/tests/util.py
rrauber: I think I'll save this for the next stage (implementing the subcommand/flag), just because the implementation of that command will probably suggest a certain location for the file / other things that could be moved there.

gslib/tests/testcase/integration_testcase.py
rrauber: Changed to `full_gsutil_command`!

PR #1545, opened 2022-07-18
Fix typo in cat documentation.

rrauber:


rrauber:
Addressed in a different PR.

PR #1549, opened 2022-07-21
Shim gsutil rb command.

ejeselsohn:
Before, there was no translation mapping between gsutil and gcloud.

After, there is a translation mapping using the GcloudStorageMap class.


NickGoog (commented):
Waiting for Koko

NickGoog:
All test failures appear KMS or CRD related

PR #1552, opened 2022-07-25
Shim requesterpays command.

NickGoog:


gslib/tests/test_requester_pays.py
```diff
@@ -260,15 +260,18 @@ def test_mv(self):
 
   def test_rewrite(self):
     object_uri = self.CreateObject(contents=b'bar')
+    storage_class = 'nearline' if self._use_gcloud_storage else 'dra'
```
thomasmaclean, line 263: What does 'dra' mean here?

thomasmaclean (changes requested):

gslib/tests/test_requester_pays.py
NickGoog: The deprecated (or at least no longer documented) "DURABLED_REDUCED_AVAILABILITY" class, which gsutil has an alias hardcoded for.

rrauber (approved):

NickGoog:
Discussed offline: Thomas approves

PR #1553, opened 2022-07-25
Allow custom storage class flag for PCU

ejeselsohn:
Before, chunks were being uploaded as the default storage class.
After, chunks are being uploaded as the custom storage class.

### DIFF
--- a/gslib/tests/test_copy_helper_funcs.py
+++ b/gslib/tests/test_copy_helper_funcs.py
@@ -111,13 +111,13 @@ class TestCpFuncs(GsUtilUnitTestCase):
     self.assertEqual(50, component_size)
 
   def testFilterExistingComponentsNonVersioned(self):
-    """Tests upload with a variety of component states, including custom storage class."""
+    """Tests upload with a variety of component states."""
     mock_api = MockCloudApi()
     bucket_name = self.MakeTempName('bucket')
     tracker_file = self.CreateTempFile(file_name='foo', contents=b'asdf')
     tracker_file_lock = parallelism_framework_util.CreateLock()
 
-    # dst_obj_metadata used for passing content-type and storage-class.
+    # dst_obj_metadata used for passing content-type and storage class.
     content_type = 'ContentType'
     storage_class = 'StorageClass'
 
@@ -223,13 +223,13 @@ class TestCpFuncs(GsUtilUnitTestCase):
                      existing_objects_to_delete[0].url_string)
 
   def testFilterExistingComponentsVersioned(self):
-    """Tests upload with versionined parallel components and custom storage class."""
+    """Tests upload with versionined parallel components."""
 
     mock_api = MockCloudApi()
     bucket_name = self.MakeTempName('bucket')
     mock_api.MockCreateVersionedBucket(bucket_name)
 
-    # dst_obj_metadata used for passing content-type and storage-class.
+    # dst_obj_metadata used for passing content-type and storage class.
     content_type = 'ContentType'
     storage_class = 'StorageClass'
 
--- a/gslib/tests/test_copy_helper_funcs.py
+++ b/gslib/tests/test_copy_helper_funcs.py
@@ -111,14 +111,15 @@ class TestCpFuncs(GsUtilUnitTestCase):
     self.assertEqual(50, component_size)
 
   def testFilterExistingComponentsNonVersioned(self):
-    """Tests upload with a variety of component states."""
+    """Tests upload with a variety of component states, including custom storage class."""
     mock_api = MockCloudApi()
     bucket_name = self.MakeTempName('bucket')
     tracker_file = self.CreateTempFile(file_name='foo', contents=b'asdf')
     tracker_file_lock = parallelism_framework_util.CreateLock()
 
-    # dst_obj_metadata used for passing content-type.
-    empty_object = apitools_messages.Object()
+    # dst_obj_metadata used for passing content-type and storage-class.
+    content_type = 'ContentType'
+    storage_class = 'StorageClass'
 
     # Already uploaded, contents still match, component still used.
     fpath_uploaded_correctly = self.CreateTempFile(file_name='foo1',
@@ -138,8 +139,8 @@ class TestCpFuncs(GsUtilUnitTestCase):
 
     args_uploaded_correctly = PerformParallelUploadFileToObjectArgs(
         fpath_uploaded_correctly, 0, 1, fpath_uploaded_correctly_url,
-        object_uploaded_correctly_url, '', empty_object, tracker_file,
-        tracker_file_lock, None, False)
+        object_uploaded_correctly_url, '', content_type, storage_class,
+        tracker_file, tracker_file_lock, None, False)
 
     # Not yet uploaded, but needed.
     fpath_not_uploaded = self.CreateTempFile(file_name='foo2', contents=b'2')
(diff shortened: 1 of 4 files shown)


gslib/tests/test_copy_helper_funcs.py
```diff
@@ -111,14 +111,15 @@ def testGetPartitionInfo(self):
     self.assertEqual(50, component_size)
 
   def testFilterExistingComponentsNonVersioned(self):
```
NickGoog, line 113: Taking a closer look, too much is happening in these tests to isolate the desired behavior.
Please create a separate test for just the bug you're fixing.

gslib/utils/copy_helper.py
```diff
@@ -335,7 +335,8 @@ def _PerformParallelUploadFileToObject(cls, args, thread_state=None):
     dst_object_metadata = apitools_messages.Object(
         name=args.dst_url.object_name,
         bucket=args.dst_url.bucket_name,
-        contentType=args.content_type)
+        contentType=args.content_type,
+        storageClass=args.storage_class)
```
NickGoog, line 339: Storage class isn't a consideration for sliced downloads because the file system destination doesn't have a concept of storage classes.

gslib/utils/copy_helper.py
```diff
@@ -1001,6 +1002,7 @@ def _PartitionFile(fp,
                    file_size,
                    src_url,
                    content_type,
+                   storage_class,
```
NickGoog, line 1005: Alphabetical order

gslib/utils/copy_helper.py
```diff
@@ -335,7 +335,8 @@ def _PerformParallelUploadFileToObject(cls, args, thread_state=None):
     dst_object_metadata = apitools_messages.Object(
```
NickGoog, line 335: Update the code comment on line 334 please

gslib/utils/copy_helper.py
```diff
@@ -1020,8 +1022,9 @@ def _PartitionFile(fp,
     file_size: The size of fp, in bytes.
     src_url: Source FileUrl from the original command.
     content_type: content type for the component and final objects.
+    storage_class: storage class for the component and final objects.
     canned_acl: The user-provided canned_acl, if applicable.
-    dst_bucket_url: CloudUrl for the destination bucket
+    dst_bucket_url: CloudUrl for the destination bucket.
```
NickGoog, line 1027: Nice

NickGoog (changes requested):
I don't see where this gets custom storage classes from user flags. Try tracing the path of the storage class flag down from `cp.py`.

gslib/tests/test_copy_helper_funcs.py
ejeselsohn: I created a separate test. Gcloud storage does not have the same bug. Do you want me to do test parity for the new test?

gslib/utils/copy_helper.py
ejeselsohn: Done.

ejeselsohn:
> I don't see where this gets custom storage classes from user flags. Try tracing the path of the storage class flag down from `cp.py`.

I believe it happens here:
https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/utils/copy_helper.py#L3866

gslib/tests/test_copy_helper_funcs.py
NickGoog: If you just add it to `gsutil_test_parity.yaml` and find a relevant test for coverage, it should be fine

gslib/tests/test_copy_helper_funcs.py
```diff
@@ -111,14 +111,15 @@ def testGetPartitionInfo(self):
...
-    """Tests upload with a variety of component states."""
+    """Tests upload with a variety of component states, including custom storage class."""
     mock_api = MockCloudApi()
     bucket_name = self.MakeTempName('bucket')
     tracker_file = self.CreateTempFile(file_name='foo', contents=b'asdf')
     tracker_file_lock = parallelism_framework_util.CreateLock()
 
-    # dst_obj_metadata used for passing content-type.
-    empty_object = apitools_messages.Object()
+    # dst_obj_metadata used for passing content-type and storage-class.
```
NickGoog, line 120: No hyphen needed for "storage class"

gslib/tests/test_copy_helper_funcs.py
```diff
@@ -111,14 +111,15 @@ def testGetPartitionInfo(self):
     self.assertEqual(50, component_size)
 
   def testFilterExistingComponentsNonVersioned(self):
-    """Tests upload with a variety of component states."""
+    """Tests upload with a variety of component states, including custom storage class."""
```
NickGoog, line 114: Is this docstring change still necessary?

gslib/tests/test_cp.py
```diff
@@ -3325,6 +3325,22 @@ def test_kms_key_correctly_applied_to_composite_upload(self):
...
+    obj_suri = suri(bucket_uri, 'composed')
+
+    with SetBotoConfigForTest([
+        ('GSUtil', 'parallel_composite_upload_threshold', '1'),
+        ('GSUtil', 'parallel_composite_upload_component_size', '1')
+    ]):
+      self.RunGsUtil(['cp', '-s', 'nearline', fpath, obj_suri])
+    stdout = self.RunGsUtil(['ls', '-L', obj_suri], return_stdout=True)
+    self.assertRegexpMatchesWithFlags(stdout,
+                                      r'Storage class:\s+NEARLINE',
```
NickGoog, line 3341: Why use `\s` instead of a regular white space?

gslib/tests/test_cp.py
```diff
@@ -3325,6 +3325,22 @@ def test_kms_key_correctly_applied_to_composite_upload(self):
     with SetBotoConfigForTest([('GSUtil', 'prefer_api', 'json')]):
       self.AssertObjectUsesCMEK(obj_suri, key_fqn)
 
+  @SkipForS3('No composite upload support for S3.')
+  def test_nearline_applied_to_parallel_composite_upload(self):
+    bucket_uri = self.CreateBucket()
```
NickGoog, line 3330: To emphasize the class difference, can you set the default class to `standard` here?

gslib/tests/test_cp.py
ejeselsohn: Becuase the way ls -L works is it tabs NEARLINE over so one space will cause the test to fail.

gslib/tests/test_copy_helper_funcs.py
ejeselsohn: Sounds good. So after this is merged I will add it to gsutil_test_parity.yaml.

gslib/tests/test_copy_helper_funcs.py
ejeselsohn: Done.

gslib/tests/test_copy_helper_funcs.py
ejeselsohn: No, I kept it to be more specific but I will delete it.

gslib/tests/test_cp.py
ejeselsohn: Done.

gslib/tests/test_cp.py
NickGoog: Like inserting the correct number of white spaces. Something like `Storage class:           NEARLINE`

ejeselsohn: After speaking with Nick, changed it to have the correct number of whitespaces.

PR #1578, opened 2022-08-25
Backfill cl/470082513

starsandskies:
Centralize the discussion of sliced object downloads in the webpage https://cloud.google.com/storage/docs/sliced-object-downloads

### DIFF
--- a/gslib/commands/cp.py
+++ b/gslib/commands/cp.py
@@ -363,26 +363,11 @@ _STREAMING_TRANSFERS_TEXT = """
 
 _SLICED_OBJECT_DOWNLOADS_TEXT = """
 <B>SLICED OBJECT DOWNLOADS</B>
-  gsutil uses HTTP Range GET requests to perform "sliced" downloads in parallel
-  when downloading large objects from Cloud Storage. This means that disk
-  space for the temporary download destination file is pre-allocated and
-  byte ranges (slices) within the file are downloaded in parallel. Once all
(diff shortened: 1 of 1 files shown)


rrauber (approved):

gslib/commands/cp.py
```diff
@@ -363,26 +363,11 @@
...
-  NOTE: Since sliced object downloads cause multiple writes to occur at various
-  locations on disk, this mechanism can degrade performance for disks with slow
-  seek times, especially for large numbers of slices. While the default number
-  of slices is set small to avoid this problem, you can disable sliced object
-  download if necessary by setting the "sliced_object_download_threshold"
-  variable in the ``.boto`` config file to 0.
+  gsutil can automatically use ranged ``GET`` requests to perform downloads in
+  parallel for large files being downloaded from Cloud Storage. See `sliced object
+  download documentation
+  <https://cloud.google.com/storage/docs/sliced-object-downloads">`_
```
rrauber, line 369: Is the underscore at the end of the line here intentional? It also seems like the " character is unmatched if that matters.

starsandskies: The underscore is intentional - I'm not exactly sure how it works, but it's the style for every URL link in this repository. I'm assuming it's somehow important for identifying the need to turn the string into a hyperlink.

The " was a typo, however. Thanks for catching, fixed!

PR #1584, opened 2022-09-06
Adding line in the version command about shim usage

thomasmaclean:


gslib/commands/version.py
```diff
@@ -117,7 +117,8 @@ def RunCommand(self):
           'gsutil path: {gsutil_path}\n'
           'compiled crcmod: {compiled_crcmod}\n'
           'installed via package manager: {is_package_install}\n'
-          'editable install: {is_editable_install}\n')
+          'editable install: {is_editable_install}\n'
+          'using gcloud storage: {is_shim_enabled}\n')
```
dilipped, line 121: I wonder if the field title should be "is shim enabled". Because enabling the shim does not guarantee that gcloud storage will be used, so will the user be misguided that the `gcloud storage` is being used for all the commands?

thomasmaclean: Good call. I'll update.

dilipped (approved):

PR #1585, opened 2022-09-06
Shim outputs metrics environment for Cloud SDK.

thomasmaclean:


gslib/utils/shim_util.py
```diff
@@ -387,7 +387,10 @@ def _translate_top_level_flags(self):
         variables that can be set for the gcloud storage command execution.
     """
     top_level_flags = []
-    env_variables = {'CLOUDSDK_STORAGE_RUN_BY_GSUTIL_SHIM': 'True'}
+    env_variables = {
+        'CLOUDSDK_METRICS_ENVIRONMENT': 'gsutil_shim',
```
dilipped, line 391: Do we need a separate environment var? Can we not rely on the RUN_BY_GSUTIL_SHIM?

thomasmaclean: This environment variable will be injected directly into the user agent.

dilipped (approved):

PR #1591, opened 2022-09-07
Shim cp continue-on-error flag

NickGoog:


thomasmaclean (commented):
Looks like you've got a few lint errors but otherwise LGTM.

PR #1592, opened 2022-09-09
Mb test parity minus retention.

NickGoog:
Also fixes pap output issue and improves shim error message.

### DIFF
--- a/gslib/tests/test_mb.py
+++ b/gslib/tests/test_mb.py
@@ -70,13 +70,21 @@ class TestMb(testcase.GsUtilIntegrationTestCase):
     stderr = self.RunGsUtil(['mb', suri(bucket_uri)],
                             expected_status=1,
                             return_stderr=True)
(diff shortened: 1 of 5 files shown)


gslib/tests/test_mb.py
```diff
@@ -123,7 +131,11 @@ def test_create_with_retention_invalid_arg(self):
                              suri(bucket_uri)],
                             expected_status=1,
                             return_stderr=True)
-    self.assertRegexpMatches(stderr, r'Incorrect retention period specified')
+    if self._use_gcloud_storage:
+      self.assertIn("Duration unit 'econd' must be preceded by a number",
```
rrauber, line 135: second?

rrauber (approved):

gslib/tests/test_mb.py
NickGoog: The "s" is actually a valid unit, so it gets cut. Will add a comment.

PR #1593, opened 2022-09-09
Translate retention flag output for mb shim

NickGoog:


rrauber (approved):
Code looks good, may be worth waiting on kokoro a bit longer before merging.

PR #1600, opened 2022-10-12
Ensure arbitrary headers are included in every rsync request.

rrauber:
Tested with S3 SSE-C headers, which required updating the boto submodule to include relevant fixes.


gslib/command.py
```diff
@@ -641,6 +641,9 @@ def __init__(self,
       for log_filter in logging_filters:
         self.logger.addFilter(log_filter)
 
+    if self.headers is not None:
+      self.non_metadata_headers = GetNonMetadataHeaders(self.headers)
```
NickGoog, line 645: What is the default value of this?
Seems like it should have:
```
else:
  self.non_metadata_headers = None
```

gslib/tests/test_rsync.py
```diff
@@ -3130,3 +3131,52 @@ def test_kms_key_specified_will_not_prevent_non_kms_copy_to_s3(self):
...
+    tmp_dir = self.CreateTempDir()
+    tmp_file = self.CreateTempFile(tmpdir=tmp_dir, contents=b'foo')
+    bucket_uri1 = self.CreateBucket()
+    bucket_uri2 = self.CreateBucket()
+
+    header_flags = [
+        '-h',
+        '"x-amz-server-side-encryption-customer-algorithm:AES256"',
+        '-h',
+        '"x-amz-server-side-encryption-customer-key:MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTI="',
```
NickGoog, line 3165: Are there constants for these somewhere?

gslib/tests/test_rsync.py
```diff
@@ -3130,3 +3131,52 @@ def test_kms_key_specified_will_not_prevent_non_kms_copy_to_s3(self):
...
+        '-h',
+        '"x-amz-server-side-encryption-customer-key-md5:dnF5x6K/8ZZRzpfSlMMM+w=="',
+    ]
+
+    with SetBotoConfigForTest([('GSUtil', 'check_hashes', 'never')]):
+      self.RunGsUtil(header_flags +
+                     ['-m', 'cp', tmp_file,
```
NickGoog, line 3172: Does this actually ensure more than one thread is used?

rrauber: Not necessary for the test, removed.

gslib/tests/test_rsync.py
rrauber: There are now!

gslib/tests/test_rsync.py
NickGoog: Then how are you testing headers get to worker threads?

gslib/command.py
NickGoog: How did this even work before?

gslib/tests/test_rsync.py
rrauber: Rsync runs in parallel by default (that's why the customer was running into issues with that command and not others): https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/commands/rsync.py#L1004

gslib/tests/test_rsync.py
NickGoog: Where did you get these values?

gslib/command.py
rrauber: Headers always has a non-none value -- it's provided an empty dict by the command initializer. This is more defensive though, so I like it better.

gslib/tests/test_rsync.py
NickGoog: Do we not have more direct tests for the `Command`, `WorkerPool`, and `WorkerThread`?

rrauber: Not really, none that are set up to catch cases like this, or need to be updated after the API change (other than the ones that I have changed)

gslib/command.py
NickGoog: You were passing `self.non_metadata_headers` around while it was potentially undeclared, so there probably should have been something like:
```
AttributeError: 'X' object has no attribute 'hi'
```

rrauber: Previously I didn't have non_metadata_headers -- I just made the GetNonMetadataHeaders call right in the CloudApiDelegator call that needed them. Now that we needed to pass them to more places an attribute seemed like a better fit, so it's new. 

The only place where complaints of the sort were raised were in unit tests that didn't instantiate Command through the command runner. This is because of this line in command runner which ensures headers always have a non-null value for user-runnable commands: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/command_runner.py#L354

That's why none of the E2E tests were having problems. As I said, adding the else as you suggested is great and better than what I had, though in practice it's not really necessary outside of tests given how commands are executed.

rrauber: With those unit tests the issue wasn't an attribute error, it was more that GetNonMetadataHeaders was getting called with None, which it didn't expect.

NickGoog: I think `AttributeError` should have happened anyways if someone managed a call where `self.headers` was `None`

NickGoog (approved):

Issue #353, opened 2016-04-13
Can we have an option to ignore errors when listing files for rsync?


poizan42:
Just having a broken symlink makes it stop. It would be nice if you could tell it to keep going and just ignore the file it couldn't list.

```
poizan@dy:~$ sudo gsutil/gsutil rsync -C -r /mnt/ext/PoiDisk/Backup/ gs://last-line-of-defence/Backup
Building synchronization state...
At source listing 10000...
At source listing 20000...
At source listing 30000...
Caught non-retryable exception while listing file:///mnt/ext/PoiDisk/Backup/: [Errno 2] No such file or directory: '/mnt/ext/PoiDisk/Backup/dybackup/poizanhome/poizan/.pftp'
At destination listing 10000...
At destination listing 20000...
At destination listing 30000...
CommandException: Caught non-retryable exception - aborting rsync
```

mfschwartz:
Hi - do you need to follow symlinks, or would it work to use gsutil rsync -e?

And if you do need to follow symlinks, can you please explain how it is useful to be able to perform an rsync that skips over invalid symlinks?

houglum:
It looks like `rsync -r --copy-links` will skip over broken symlinks, but inform users when it finished that it did so:

```
$ rsync -r -vv --copy-links rdir1/ rdir2/
sending incremental file list
symlink has no referent: "/[...]/rdir1/broken_link_file"
delta-transmission disabled for local transfer or --whole-file
valid_file1
valid_file2
valid_file2

pewfile
pewfile2
pewfile3
total: matches=0  hash_hits=0  false_alarms=0 data=16

sent 320 bytes  received 162 bytes  964.00 bytes/sec
total size is 16  speedup is 0.03
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1183) [sender=3.1.0]
```

There error here comes from `_FieldedListingIterator` creating a `FileWildcardIterator` -- this calls `_GetFileObject`, which attempts to call `os.path.getsize` and thus tries to stat an invalid file path.

While detecting the raised OSError somewhere along this path is doable, I'm not sure what implications this might have on other places where we use such iterators; that will require a bit more investigation.

VikramTiwari:
Any way this could be prioritized. The only for now is to know which symlinks are broken and skip those. No way to get this working if you don't know what those are.

PR #1602, opened 2022-10-13
Wildcard iterator should exclude filepaths from rsync -x

thomasmaclean:


gslib/wildcard_iterator.py
```diff
@@ -583,16 +583,23 @@ class FileWildcardIterator(WildcardIterator):
...
     Args:
       wildcard_url: FileUrl that contains the wildcard to iterate.
+      exclude_pattern: A regex of paths to ignore during iteration.
       ignore_symlinks: If True, ignore symlinks during iteration.
       logger: logging.Logger used for outputting debug messages during
               iteration. If None, the root logger will be used.
     """
     self.wildcard_url = wildcard_url
+    self.exclude_pattern = re.compile(
+        exclude_pattern) if exclude_pattern else None
```
dilipped, line 602: Recommended to perform explicit checks for None. i.e if exclude_pattern is not None else None

gslib/wildcard_iterator.py
```diff
@@ -583,16 +583,23 @@ class FileWildcardIterator(WildcardIterator):
...
+  def __init__(self,
+               wildcard_url,
+               exclude_pattern=None,
+               ignore_symlinks=False,
+               logger=None):
     """Instantiates an iterator over BucketListingRefs matching wildcard URL.
 
     Args:
       wildcard_url: FileUrl that contains the wildcard to iterate.
+      exclude_pattern: A regex of paths to ignore during iteration.
```
dilipped, line 595: Might we worth mentioning here that it is ignored for Cloud URLs

gslib/wildcard_iterator.py
```diff
@@ -643,6 +650,10 @@ def __iter__(self, bucket_listing_fields=None):
     for filepath in filepaths:
       expanded_url = StorageUrlFromString(filepath)
       try:
+        if self.exclude_pattern and self.exclude_pattern.match(filepath):
```
dilipped, line 653: The help text mentions that the pattern is relative to the source https://cloud.google.com/storage/docs/gsutil/commands/rsync#options

But I guess in this case ,let's say I have a pattern `data/.*` and the source is `dir`, I think `the expanded_url` will be `dir/data/a.txt` but the pattern will be `data/.*` and it won't match?

thomasmaclean: I saw this too... I refactored the code to match the logic in rsync 1:1 so that skips happen in the wildcard iterator first.

gslib/wildcard_iterator.py
thomasmaclean: This description is specifically for `FileWildcardIterator` which doesn't work with Cloud URLs.

gslib/wildcard_iterator.py
thomasmaclean: Just realized the exclude_pattern is already a RegEx when it's passed in, so this whole line is being simplified to a simple assignment.

gslib/wildcard_iterator.py
```diff
@@ -643,6 +649,15 @@ def __iter__(self, bucket_listing_fields=None):
     for filepath in filepaths:
       expanded_url = StorageUrlFromString(filepath)
       try:
+        if self.exclude_pattern:
+          base_url_str, _ = os.path.split(wildcard)
```
dilipped, line 653: What if the wildcard look something like `f*/b*` that matches `foo/bar`. Won't the base_url_str be incorrect? I am guessing this won't be an issue for rsync as it looks like rsync already expands the wildcard before passing down the string to the WildcardIterator (it calls wildcardIterator twice), but given that this is generic code used for non-rsync commands as well, I think we should address it.

thomasmaclean: So `wildcard` is actually made up of `base_url_str` and the suffix I split off here if you follow the logic back to `rsync.py`, but I do see your point. One way to side-step this entirely is to require an `exclusion_tuple` that contains the original `exclusion_pattern` and the original `base_url_str`, so that way we're not trying to reconstitute it using `wildcard` in some dubious way.

dilipped: Alternatively, we can restrict the exclusion pattern to work only if the `base_url_str` does not have a wildcard, i.e we say that if you want to filter out any patterns, you should not pass in a wildcard string other than the one with format `<wildcard-less path>/**`.  Given that rsync already runs a wildcard iterator on the source to determine that the [source is a single directory](https://github.com/GoogleCloudPlatform/gsutil/blob/bf9394b01bb29d33653e93aad1cd143532d905d6/gslib/commands/rsync.py#L1648), and given the fact that the -x option is currently used only for rsync, it shouldn't be an issue.

It would look something like this
```
base_url_str, _ = os.path.split(wildcard)
if storage_url.WILDCARD_REGEX.match(base_url_str):
  # should never happen for rsync,
  # but just to protect any other command accidentally using this feature in the future.
  raise Error('-x pattern is only allowed for non-wildcard paths')
...
```

For the bug that lead to this fix, one other solution could be to just expand the `continue-on-error` to work for wildcard iterator as well. But this has greater side effect, although we might end up fixing another issue if we do that https://github.com/GoogleCloudPlatform/gsutil/issues/353. Having said that, I think we should still go ahead with fixing -x.

dilipped: Thinking a bit more about this, I think my above proposal is not the best as it does make the logic only relevant to the cases where source is presented as a directory. It will work because rsync always deals with directory, but since wildcard iterator is a generic module, I feel it's not the best. 

So an alternative could be to pass in the exclude pattern with the base_url_str appended. This can be done in the rsync code. So for `gsutil rsync -x "data./.*\.txt$" dir gs://my-bucket`, the exclude pattern will be `dir/data./.*\.txt$` instead of `data./.*\.txt$`. So the entire logic of extracting `base_url_str` can stay in `rsync` and the code in wildcard iterator can stay generic in true sense. I think the logic can go somewhere here https://github.com/GoogleCloudPlatform/gsutil/blob/bf9394b01bb29d33653e93aad1cd143532d905d6/gslib/commands/rsync.py#L734 .

thomasmaclean: I have a version of this completed, I'll be adding it here in a second.

gslib/wildcard_iterator.py
```diff
@@ -649,12 +651,12 @@ def __iter__(self, bucket_listing_fields=None):
     for filepath in filepaths:
       expanded_url = StorageUrlFromString(filepath)
       try:
-        if self.exclude_pattern:
-          base_url_str, _ = os.path.split(wildcard)
+        if self.exclude_tuple is not None:
+          (base_url_str, exclude_pattern) = self.exclude_tuple
```
dilipped, line 655: This works , but how about doing what I was suggesting here https://github.com/GoogleCloudPlatform/gsutil/pull/1602#discussion_r995250227. It will let us avoid all this logic involving file path slicing. This means, instead of passing the exclude_tuple, you can just pass the complete exclude pattern. i.e `os.path.join(base_url_str, exclude_pattern) if exclude_pattern else None` and then you will just have to do
```
if self.exclude_pattern is not None:
  if self.exclude_pattern.match(filepath):
    # log message
    continue
```

This also helps avoid passing in a tuple when a single string would suffice.

dilipped: We can do `exclude_pattern.pattern` to get the string back. I tried pulling your PR and making local changes, and it works. Pasting the relevant diff here.

```
-        expanded_exclude_pattern=re.compile(os.path.join(base_url_str, cls.exclude_pattern.pattern)),
+        exclude_tuple=(base_url_str, cls.exclude_pattern),

...

       try:
-        if self.expanded_exclude_pattern is not None:
-          if self.expanded_exclude_pattern.match(str(expanded_url)):
+        if self.exclude_tuple is not None:
+          (base_url_str, exclude_pattern) = self.exclude_tuple
+          str_to_check = filepath[len(base_url_str):]
+          if str_to_check.startswith(self.wildcard_url.delim):
+            str_to_check = str_to_check[1:]
+          if exclude_pattern.match(str_to_check):
             if self.logger:
               self.logger.info('Skipping excluded path %s...', filepath)
             continue

...

@@ -807,7 +811,7 @@ def CreateWildcardIterator(url_str,
                            gsutil_api,
                            all_versions=False,
                            project_id=None,
-                           expanded_exclude_pattern=None,
+                           exclude_tuple=(None, None),
```



The reason I feel like doing that is because the notion of a `base_url_str` is very rsync specific as rsync deals with a single base dir. What if we have source pattern that resolves to multiple dirs, so my suggestion is to do is try to avoid having rsync specific logic in wildcard iteartor. If you think of wildcard iterator on its own, what should we pass in as `base_url_str` is not clear. Let's say we want to use it for `cp` command in the future, then what should we pass? Note that the logic relies on the fact that the `base_url_str` should not have any wildcard, this is possible only in `rsync` since the base is already expanded. If we simply have a contract that says `send a regex pattern that should be used for filtering the listing`, then it's up to the caller on what pattern to pass down and hence the suggestion. For rsync it might look like`base_url_str + exclude_pattern`, for cp it might look like something else, and wildcard iterator should not have to worry about it. It will just use the pattern to filter out the results.

Regarding matching the logic in rsync, the current logic in rsync is based off of certain assumptions that rsync already has the base wildcard expanded and it only works on a single dir. Additionally it also does not call wildcard iterator for non-recursive cases (it lists the directory directly using `_LocalDirIterator`), so for that reason,  I feel we shouldn't try to match that logic in wildcard iterator.

thomasmaclean: Thanks for the in-depth explanation... I just committed a version of your suggestion.

gslib/tests/test_wildcard_iterator.py
```diff
@@ -499,3 +501,14 @@ def testExistingDirNoFileMatch(self):
         self._test_wildcard_iterator(uri).IterAll(
             expand_top_level_buckets=True))
     self.assertEqual(0, len(res))
+
+  def testExcludeDir(self):
+    """Tests that the exclude regex will omit a nested directory."""
+    exp_uri_strs = self.root_files_uri_strs
+    uri = self._test_storage_uri(suri(self.test_dir, '*'))
+    et = (self.test_dir, re.compile('dir1'))
```
dilipped, line 509: Looks like this test was not updated?

gslib/wildcard_iterator.py
dilipped: Thanks for accepting the suggestion! Anything that involves wildcard iterator scares me, haha, so I'm doing extra scrutiny here :) 

 I have left few more comments that should be straight-forward to address.

gslib/tests/test_wildcard_iterator.py
thomasmaclean: Still working on updating the test.

thomasmaclean: OK the tests are now passing and everything is good.

gslib/tests/test_wildcard_iterator.py
```diff
@@ -499,3 +500,14 @@ def testExistingDirNoFileMatch(self):
         self._test_wildcard_iterator(uri).IterAll(
             expand_top_level_buckets=True))
     self.assertEqual(0, len(res))
+
+  def testExcludeDir(self):
+    """Tests that the exclude regex will omit a nested directory."""
+    exp_uri_strs = self.root_files_uri_strs
+    uri = self._test_storage_uri(suri(self.test_dir, '*'))
+    exclude_pattern = re.compile('file://{}/dir1'.format(self.test_dir))
```
dilipped, line 508: Make sure to check Kokoro for Windows given that forward slash is involved. 

Nit: `suri(self.test_dir, 'dir1')` should work as well if you want to follow the `suri` pattern everywhere else.

dilipped (approved):
LGTM

gslib/commands/rsync.py
```diff
@@ -731,10 +731,14 @@ def _FieldedListingIterator(cls, gsutil_api, base_url_str, desc):
           'metadata/%s' % GID_ATTR,
           'metadata/%s' % UID_ATTR,
       ])
+    expanded_exclude_pattern = re.compile('{}/{}'.format(
+        base_url_str.rstrip('/\\'), cls.exclude_pattern.pattern
```
dilipped, line 735: Can we also add a test to ensure that rstrip was called? You can also modify the existing test to have a slash in the end of the base dir.

dilipped (commented):
Does rsync have unit tests? If yes, then can you please consider adding a unit test to ensure that wildcard iterator gets called with exclude pattern.

gslib/tests/test_wildcard_iterator.py
thomasmaclean: Done.

gslib/commands/rsync.py
thomasmaclean: Just added an integration test to rsync and included a slash in the base directory.

dilipped: Excellent! LGTM!

dilipped (approved):

NickGoog:
Hey, this broke master. Please check Kokoro before submitting and approving next time http://sponge2/ed7c055b-a3ee-4720-8580-885a55952405

PR #1607, opened 2022-10-18
Fixed possible TypeError on some versions of Python

thomasmaclean:


gslib/boto_translation.py
```diff
@@ -651,13 +652,15 @@ def _PerformSimpleDownload(self,
                              num_progress_callbacks=XML_PROGRESS_CALLBACKS,
                              headers=None,
                              hash_algs=None):
-    try:
+    argspec = inspect.getfullargspec(key.get_contents_to_file)
```
NickGoog, line 655: Using `inspect` is generally an anti-pattern. Can you do a type check with `isinstance(key, SOME_TYPE)`?
If not, can you put a comment here explaining why `inspect` is necessary?

thomasmaclean: Switched to isinstance...

gslib/boto_translation.py
```diff
@@ -651,13 +652,14 @@ def _PerformSimpleDownload(self,
                              num_progress_callbacks=XML_PROGRESS_CALLBACKS,
                              headers=None,
                              hash_algs=None):
-    try:
+    if isinstance(key, GsKey):
```
dilipped, line 655: I feel we should ask more clarifying questions to the user as I do not see any reason why the existing code would not catch TypeError.

dilipped: Plus, we are not even sure that this fix is going to resolve the issue. It might be some other issue like I have mentioned offline in the chat.

PR #1608, opened 2022-10-20
Shim bucketpolicy only and ubla commands

NickGoog:
Spoilers: The two commands are the same except for some changed display strings.

### DIFF
--- a/gslib/commands/ubla.py
+++ b/gslib/commands/ubla.py
@@ -29,6 +29,7 @@ from gslib.exception import NO_URLS_MATCHED_TARGET
 from gslib.help_provider import CreateHelpText
 from gslib.third_party.storage_apitools import storage_v1_messages as apitools_messages
 from gslib.utils.constants import NO_MAX
+from gslib.utils.shim_util import GcloudStorageMap
 from gslib.utils.text_util import InsistOnOrOff
 
 _SET_SYNOPSIS = """
@@ -83,6 +84,15 @@ _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 IamConfigurationValue = apitools_messages.Bucket.IamConfigurationValue
 uniformBucketLevelAccessValue = IamConfigurationValue.BucketPolicyOnlyValue
 
+GCLOUD_FORMAT_STRING = (
+    '--format='
+    'multi[terminator="\n"](name:format="value(format(\'Uniform bucket-level'
+    ' access setting for gs://{}:\'))",'
+    ' iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no="False")'
+    ':format="value[terminator=\'\n\'](format(\'  Enabled: {}\'))",'
+    ' iamConfiguration.uniformBucketLevelAccess.lockedTime.sub("T", " ")'
+    ':format="value(format(\'  LockedTime: {}\'))")')
+
 
 class UblaCommand(Command):
   """Implements the gsutil ubla command."""
@@ -117,6 +127,48 @@ class UblaCommand(Command):
           'get': _get_help_text,
           'set': _set_help_text,
       },
+  )
+
+  gcloud_storage_map = GcloudStorageMap(
+      gcloud_command={
+          'get':
+              GcloudStorageMap(
+                  gcloud_command=[
+                      'alpha', 'storage', 'buckets', 'list',
+                      GCLOUD_FORMAT_STRING
+                  ],
+                  flag_map={},
+              ),
+          'set':
+              GcloudStorageMap(
+                  gcloud_command={
+                      'on':
+                          GcloudStorageMap(
+                              gcloud_command=[
+                                  'alpha',
+                                  'storage',
+                                  'buckets',
+                                  'update',
+                                  '--uniform-bucket-level-access',
+                              ],
+                              flag_map={},
+                          ),
+                      'off':
+                          GcloudStorageMap(
+                              gcloud_command=[
+                                  'alpha',
+                                  'storage',
+                                  'buckets',
+                                  'update',
+                                  '--no-uniform-bucket-level-access',
+                              ],
+                              flag_map={},
+                          ),
+                  },
+                  flag_map={},
+              )
+      },
+      flag_map={},
   )
 
   def _ValidateBucketListingRefAndReturnBucketName(self, blr):
--- a/gslib/commands/bucketpolicyonly.py
+++ b/gslib/commands/bucketpolicyonly.py
@@ -23,6 +23,7 @@ import textwrap
 from gslib import metrics
 from gslib.command import Command
 from gslib.command_argument import CommandArgument
+from gslib.commands import ubla
 from gslib.cs_api_map import ApiSelector
 from gslib.exception import CommandException
 from gslib.exception import NO_URLS_MATCHED_TARGET
@@ -121,6 +122,10 @@ class BucketPolicyOnlyCommand(Command):
           'set': _set_help_text,
       },
   )
+
+  # Get command will display "Uniform bucket-level access" display string, but
+  # contents are otherwise identical.
+  gcloud_storage_map = ubla.UblaCommand.gcloud_storage_map
 
   def _ValidateBucketListingRefAndReturnBucketName(self, blr):
     if blr.storage_url.scheme != 'gs':


gslib/commands/ubla.py
```diff
@@ -119,6 +120,48 @@ class UblaCommand(Command):
       },
   )
 
+  gcloud_storage_map = GcloudStorageMap(
+      gcloud_command={
+          'get':
+              GcloudStorageMap(
+                  gcloud_command=[
+                      'alpha', 'storage', 'buckets', 'list',
+                      '--format=json(iamConfiguration.uniformBucketLevelAccess)'
```
dilipped, line 129: Here is the format string that matches with gsutil. `gcloud alpha storage buckets list --format='value[separator="\n", terminator="\n\n"](format("Uniform bucket-level access setting for {}", url_string.sub("/$", ":")), format("  Enabled: {}", bucket_policy_only.enabled), format("  LockedTime: {}", bucket_policy_only.lockedTime))' `. 

With gsutil
```
$ gsutil ubla get  gs://test_bucket_booleans/ gs://gsutil_test_bucket-4
Uniform bucket-level access setting for gs://test_bucket_booleans:
  Enabled: True
  LockedTime: 2022-07-07 19:53:21.403000+00:00

Uniform bucket-level access setting for gs://gsutil_test_bucket-4:
  Enabled: False
```

With the above formatted string
```
$ gcloud alpha storage buckets list --format='value[separator="\n", terminator="\n\n"](format("Uniform bucket-level access setting for {}", url_string.sub("/$", ":")), format("  Enabled: {}", bucket_policy_only.enabled), format("  LockedTime: {}", bucket_policy_only.lockedTime))' gs://test_bucket_booleans/  gs://gsutil_test_bucket-4
Uniform bucket-level access setting for gs://test_bucket_booleans:
  Enabled: True
  LockedTime: 2022-07-07T19:53:21.403000+00:00

Uniform bucket-level access setting for gs://gsutil_test_bucket-4:
  Enabled: False
  LockedTime: None

```

This was based off of the DisplayableResource object so you might have to update the field names. Let me know offline if you'd like to take a look at the DD.

dilipped: Sure. I'm fine if we have decided that matching gsutil formatting for all commands is not needed. In fact, that was my first proposal, but later we decided to match all the formatting and I came up with the above formatting string.

BTW, might be better to have `yaml` format instead of `json`, slightly easier to read and concise.

NickGoog:
I lean toward JSON b/c commands like `lifecycle get` output it in gsutil

gslib/commands/ubla.py
dilipped: I am replying here to the comment below to keep the conversation in one place. I think we use JSON for lifecycle because it is meant to be written to a file and re-used. For all other bucket booleans, we focus more on human-readability, so would suggest using yaml. Also, it will be more consistent with other gsutil get commands.

NickGoog:
Hmm? I'm doing JSON to be consistent with the other gsutil get commands
(not YAML)

dilipped:
What do you mean? are we doing json output for all other bucket booleans get command? I am not aware of this decision. Also, please avoid replying via the email as it adds your comment as a stand-alone comment instead of chaining it to the existing thread.

NickGoog:
Ok, will reply on the main thread

gslib/commands/ubla.py
NickGoog: Not aware of any gsutil commands that return YAML. Commands that print JSON (non-exhaustive):

https://github.com/GoogleCloudPlatform/gsutil/blob/578f56cb51879d7f5e40dbb14dcf7eb64bc651c7/gslib/commands/cors.py#L205

https://github.com/GoogleCloudPlatform/gsutil/blob/578f56cb51879d7f5e40dbb14dcf7eb64bc651c7/gslib/commands/iam.py#L362

https://github.com/GoogleCloudPlatform/gsutil/blob/578f56cb51879d7f5e40dbb14dcf7eb64bc651c7/gslib/commands/label.py#L357

https://github.com/GoogleCloudPlatform/gsutil/blob/578f56cb51879d7f5e40dbb14dcf7eb64bc651c7/gslib/commands/lifecycle.py#L189

https://github.com/GoogleCloudPlatform/gsutil/blob/578f56cb51879d7f5e40dbb14dcf7eb64bc651c7/gslib/commands/logging.py#L233

This may partially be for file-reuse, but I personally think JSON is as readable than YAML and switching to YAML for a single command would be bad UX.

NickGoog: Won't `lockedTime.sub('T', '')` error if `lockedTime` is null? And what is `T`?
Also, if `lockedTime` is null, won't `format(...)` yield `LockedTime: null` instead of an empty string?

dilipped: Discussed offline but adding here for future reference. The `multi` format will discard the field completely if it's `None`, so in this case, `sub` will never be triggered if lockedTime is None. Also the second result there shows a case of LockedTime being `None`.

Regarding `T`, The ISO format for datetime looks like this `2022-07-07T19:53:21.403000+00:00` but gsutil prints it without `T`, so we are just replacing it with a space.

NickGoog: Embedding in Python strings doesn't work:
```
GCLOUD_FORMAT = (
    '--format="multi[terminator=\'\n\']('
    'name:format=\'value(format(\"'
    'Uniform bucket-level access setting for gs://{}:\"))\','
    ' iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no=\'False\')'
    ':format=\'value[terminator=\"\n\"](format(\" Enabled: {}\"))\','
    ' iamConfiguration.uniformBucketLevelAccess.lockedTime.sub(\'T\', \' \')'
    ':format=\'value(format(\" LockedTime: {}\"))\')"'
)
```
->
```
ERROR: (gcloud.alpha.storage.buckets.list) Name expected [default *HERE* "multi[terminator='
'](name:format='value(format("Uniform bucket-level access setting for gs://{}:"))', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False'):format='value[terminator="
"](format(" Enabled: {}"))', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' '):format='value(format(" LockedTime: {}"))')"].
```

Same result for replacing all single quotation marks with escaped double quotation marks and embedding the format string in a block string:
```
"""\
--format="multi[terminator='\n'](name:format='value(format\
(\"Uniform bucket-level access setting for gs://{}:\"))',\
 iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False'):format=\
'value[terminator=\"\n\"](format(\" Enabled: {}\"))',\
 iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' ')\
:format='value(format(\" LockedTime: {}\"))')"'"""
```

dilipped: maybe try using raw string `r'....'`

NickGoog: Yeah, same error:
```
GCLOUD_DESCRIBE_FORMAT = r'--format="multi[terminator=\'\n\'](name:format=\'value(format(\"Uniform bucket-level access setting for gs://{}:\"))\', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no=\'False\'):format=\'value[terminator=\"\n\"](format(\" Enabled: {}\"))\', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub(\'T\', \' \'):format=\'value(format(\" LockedTime: {}\"))\')"'
```

Still had to escape inner-single quotation marks to avoid syntax errors. We got this working for notifications by avoiding the outer-double quotation marks but that causes errors for this string.

NickGoog: I got the the generated flags to be identical by escaping newlines and double quotation marks, but it still fails
Code:
```
gcloud_command=[
                      'alpha', 'storage', 'buckets', 'list',
                      '--format="multi[terminator=\'\\n\'](name:format=\'value(format(\\"Uniform bucket-level access setting for gs://{}:\\"))\', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no=\'False\'):format=\'value[terminator=\\"\\n\\"](format(\\" Enabled: {}\\"))\', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub(\'T\', \' \'):format=\'value(format(\\" LockedTime: {}\\"))\')"'
                  ],
```
Execution:
```
Gcloud Storage Command: ... gcloud alpha storage buckets list --format="multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level access setting for gs://{}:\"))', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False'):format='value[terminator=\"\n\"](format(\" Enabled: {}\"))', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' '):format='value(format(\" LockedTime: {}\"))')" gs://bucket --verbosity debug
...
ERROR: (gcloud.alpha.storage.buckets.list) Name expected [default *HERE* "multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level access setting for gs://{}:\"))', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False'):format='value[terminator=\"\n\"](format(\" Enabled: {}\"))', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' '):format='value(format(\" LockedTime: {}\"))')"].
```

NickGoog: `repr()` of the executed command seems to be more accurate than the `Gcloud Storage Command:` debug log. Shows escaping is still happening with all combinations tried so far.

NickGoog: The only solution I haven't tried is concacting strings with "+" and switching between outer double quotation and single quotation marks every time a quotation mark has to be embedded. However, overall, I think it isn't worth the effort given the earlier points and time wasted.

dilipped: This worked for me

```
GCLOUD_FORMAT_STRING = (
  "multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level"
  " access setting for gs://{}:\"))',"
  " iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False')"
  ":format='value[terminator=\"\n\"](format(\"  Enabled: {}\"))',"
  " iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' ')"
  ":format='value(format(\"  LockedTime: {}\"))')")
```
Then in your map
```
          'get':
              GcloudStorageMap(
                  gcloud_command=[
                      'alpha', 'storage', 'buckets', 'list',
                      '--format', GCLOUD_FORMAT_STRING
                  ],
                  flag_map={},
              ),
```

NickGoog: That works! It's the equivalent of not putting the outer quotation marks after `--format=`, which is strange b/c they're required for input on the terminal:
```
$ gcloud storage buckets describe --format=multi...
bash: syntax error near unexpected token `('
```
`repr` agrees and shows adding even more escaping around the inner single quotation marks:
```
--format=multi[terminator=\'\n\'](name:format=\'value(format("Uniform bucket-level access setting for gs://{}:"))\', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no=\'False\'):format=\'value[terminator="\n"](format("  Enabled: {}"))\', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub(\'T\', \' \'):format=\'value(format("  LockedTime: {}"))\')'
```
Your string continues to work if I combine it with `--format=` (also if I invert the double quotations and single quotations but that's irrelevant to the upcoming point):
```
GCLOUD_FORMAT_STRING = (
  "--format="
  "multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level"
  " access setting for gs://{}:\"))',"
  " iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False')"
  ":format='value[terminator=\"\n\"](format(\"  Enabled: {}\"))',"
  " iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' ')"
  ":format='value(format(\"  LockedTime: {}\"))')")
```

Compare this to Shyam's string with newlines added for display purposes only:
```
--format=
"multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level
 access setting for gs://{}:\"))',
iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False')
:format='value[terminator=\"\n\"](format(\" Enabled: {}\"))', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' ')
:format='value(format(\" LockedTime: {}\"))')"
```
I suppose, the takeaway here is that for *some* reason, we don't need wrapping quotation marks after `--format=` in Python strings, but we do need them in terminal strings. Nothing stood out to me in the `arparse` source or the shim code that might be causing this aside from the fact that outer quotation marks in Python triggers more escaping of internal quotation marks. I didn't look for super long though because we've already spent a while on this.

dilipped: When running the command via the terminal, you need the quotes so that the shell does not try parsing the string. It's basically why we recommend using double quotes around the wildcard string for ls/cp command so that the string is passed as it is to the code instead of the shell trying to expand it. The same principles apply here. `--format="mystring"` on a command line would just mean that `mystring` gets passed as it is. When we do it via the code, there is not shell in place to expand the string and hence quotes is not needed. In fact, if we add quotes, it gets passed as it is, so a string `--format="mystring"` will result in a value of `"mystring"` instead of `mystring`.

Here is a quick demo
```
$ cat test_argparse.py
import sys
print(sys.argv)

$ python3 test_argparse.py --foo "a(val)"
['test_argparse.py', '--foo', 'a(val)']
$ python3 test_argparse.py --foo="a(val)"
['test_argparse.py', '--foo=a(val)']
```

For the second example `--foo="a(val)"`, argparse just splits the string on `=`. https://github.com/python/cpython/blob/0e4e058602d93b88256ff90bbef501ba20be9dd3/Lib/argparse.py#L2204

dilipped (approved):

gslib/commands/ubla.py
NickGoog: Looks like the terminal swallows all unescaped quotation marks (single and double):

```
$ python3 test_argparse.py """""--format=\"hi\""""""
['test_argparse.py', '--format="hi"']
```
I tried running Shyam's string without the unescaped single quotes, and it escaped single quotation marks (must be b/c they were nested inside double quotation marks):
```
$ python3 test_argparse.py --format='hi'
['test_argparse.py', '--format=hi']
hartunian@hcloud2:~/Desktop$ python3 test_argparse.py --format="'hi'"
['test_argparse.py', "--format='hi'"]

$ python3 test_argparse.py --format="multi[terminator='\n'](name:format='value(format(\"Uniform bucket-level access setting for gs://{}:\"))', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no='False'):format='value[terminator=\"\n\"](format(\" Enabled: {}\"))', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub('T', ' '):format='value(format(\" LockedTime: {}\"))')"
['test_argparse.py', '--format=multi[terminator=\'\\n\'](name:format=\'value(format("Uniform bucket-level access setting for gs://{}:"))\', iamConfiguration.uniformBucketLevelAccess.enabled.yesno(no=\'False\'):format=\'value[terminator="\\n"](format(" Enabled: {}"))\', iamConfiguration.uniformBucketLevelAccess.lockedTime.sub(\'T\', \' \'):format=\'value(format(" LockedTime: {}"))\')']
```

Looks Python `subprocess.run` may have more formatting logic to add to the equation xD
```
$ cat test_test_argparse.py 
import subprocess

flags1 = ['--format', '"json"']
flags2 = ['--format="json"']
flags3 = ["--format", "'json'"]
flags4 = ["--format='json'"]
flags5 = ["--format", "json"]
flags6 = ["--format=json"]
for flags in (flags1, flags2, flags3, flags4, flags5, flags6):
  print(subprocess.run(
      ['python3', './test_argparse.py']
      + flags1, stdout=subprocess.PIPE).stdout)

$ python3 test_test_argparse.py 
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
b'[\'./test_argparse.py\', \'--format\', \'"json"\']\n'
```

PR #1610, opened 2022-10-20
Shim lifecycle command.

NickGoog:
Shimmed "get" subcommand output should be similar to gsutil, except for the case where the lifecycle doesn't exist. In that case, the shim will print "null" or a mostly empty JSON object instead of "%s has no lifecycle configuration.\n".

Do not submit until corresponding gcloud CLs are in to support the lifecycle file format.

### DIFF
--- a/gslib/commands/lifecycle.py
+++ b/gslib/commands/lifecycle.py
@@ -31,6 +31,7 @@ from gslib.help_provider import CreateHelpText
 from gslib.storage_url import UrlsAreForSingleProvider
 from gslib.third_party.storage_apitools import storage_v1_messages as apitools_messages
 from gslib.utils.constants import NO_MAX
+from gslib.utils.shim_util import GcloudStorageMap
 from gslib.utils.translation_helper import LifecycleTranslation
 
 _GET_SYNOPSIS = """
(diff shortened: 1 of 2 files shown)


gslib/commands/lifecycle.py
```diff
@@ -137,6 +138,45 @@ class LifecycleCommand(Command):
...
+                      ] + self.args[2:],  # Adds target bucket URLs.
+                      flag_map={},
+                  ),
+          },
+          flag_map={},
+      )
+      # Don't trigger unneeded translation now that complete command is above.
+      self.args = self.args[:1]
+    else:
+      gcloud_storage_map = LifecycleCommand.gcloud_storage_map
```
dilipped, line 176: Nit: You can probably add the map directly here, will be consistent with the set section above.

gslib/tests/test_lifecycle.py
```diff
@@ -90,23 +99,34 @@ def test_default_lifecycle(self):
...
 
   def test_valid_lifecycle(self):
     bucket_uri = self.CreateBucket()
     fpath = self.CreateTempFile(contents=self.lifecycle_doc.encode('ascii'))
     self.RunGsUtil(['lifecycle', 'set', fpath, suri(bucket_uri)])
     stdout = self.RunGsUtil(
         ['lifecycle', 'get', suri(bucket_uri)], return_stdout=True)
-    self.assertEqual(json.loads(stdout), self.lifecycle_json_obj)
+
+    if self._use_gcloud_storage:
```
dilipped, line 125: I think we should match the gsutil output here, else customers might have to modify their existing JSON.

NickGoog: Why would people have to modify existing JSON?

dilipped: If I understand correctly based off of the PR, gcloud storage's output will be in the form `{'Lifecylce': {k:v}}`, whereas gsutil's output will be in the form `{k:v}`. This means that if we have to keep the input format same for both gcloud storage and gsutil, the users will have to modify the output from gcloud storage (i.e from the shim) and make it look like the one returned by pure gsutil command? I might be missing something here but this is what I got from the PR.

dilipped: Alternatively, if we are accepting the format `{'Lifecylce': {k:v}}` in gcloud storage, the existing gsutil format might not work.

NickGoog: Both gcloud and gsutil accept `{'Lifecycle': {k:v}}` or `{k:v}`

Might be able to remove the `Lifecycle` outer key with a more complex format string using `multi`, but don't see a need given the outer key is accepted by both CLIs.

gslib/commands/lifecycle.py
NickGoog: Done

gslib/tests/test_lifecycle.py
dilipped: I see. That good that both CLIs accept both format. Even so, I would say, let's just keep the same format using the shim. I would not want to break any other scripts that the users might have that rely on the format outputted by gsutil. Ideally for all get commands, we shouldn't have any modification in the test code accept for stderr messages.

NickGoog: Done

dilipped (approved):

dilipped:
Please check kokoro failures are unrelated. The new interface does not show all the platforms together for a particular PR. Is there a quick way to look at it?

NickGoog:
Filed feedback to the dashboard maintainers. Also, the tests looked fine, but I just retriggered to be sure.

PR #1614, opened 2022-10-27
Update google auth dependency for interactive mode, and url validation

rrauber:


NickGoog:
Looks good, waiting for Kokoro

NickGoog:
There's an access denied error that's a little troubling, given that this is an auth change:
`test_skip_parent_directory_symbol_object_with_symlink_destination`
http://sponge2/a1f37d33-4e8b-4618-9898-1035525fb6a4

Looks like other builds succeeded with that same test, though. 
Will approve, but can you give it a second look?

NickGoog (approved):

rrauber:
Seems to be a flake given that other MacOS tests pass, as you pointed out. Other platforms succeed too. I've re-triggered the MacOS build, but will proceed for now.

PR #1616, opened 2022-11-01
rpo shim support and test parity

pradipm90:
rpo shim support and test parity

### DIFF
--- a/gslib/commands/rpo.py
+++ b/gslib/commands/rpo.py
@@ -28,6 +28,8 @@ from gslib.exception import NO_URLS_MATCHED_TARGET
 from gslib.help_provider import CreateHelpText
 from gslib.third_party.storage_apitools import storage_v1_messages as apitools_messages
 from gslib.utils.constants import NO_MAX
+from gslib.utils.shim_util import GcloudStorageFlag
+from gslib.utils.shim_util import GcloudStorageMap
 
 VALID_RPO_VALUES = ('ASYNC_TURBO', 'DEFAULT')
 VALID_RPO_VALUES_STRING = '({})'.format('|'.join(VALID_RPO_VALUES))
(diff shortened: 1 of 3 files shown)


NickGoog:
LGTM, will wait for Dilip b/c he has more familiarity with the feature

gslib/commands/rpo.py
```diff
@@ -116,6 +121,48 @@ class RpoCommand(Command):
...
+              GcloudStorageMap(
+                  gcloud_command={
+                      'enforced':
+                          GcloudStorageMap(
+                              gcloud_command=[
+                                  'alpha',
+                                  'storage',
+                                  'buckets',
+                                  'update',
+                                  '--rpo',
```
dilipped, line 144: I will have to take a look at the gcloud CL for rpo. I don't think we should have boolean flags for rpo as the set command accepts strings like ASYNC_TURBO and DEFAULT. Please ping me offline the CL for rpo.

gslib/commands/rpo.py
```diff
@@ -116,6 +121,48 @@ class RpoCommand(Command):
...
+                  gcloud_command=[
+                      'alpha', 'storage', 'buckets', 'list', _GCLOUD_LIST_FORMAT
+                  ],
+                  flag_map={},
+                  supports_output_translation=True,
+              ),
+          'set':
+              GcloudStorageMap(
+                  gcloud_command={
+                      'enforced':
```
dilipped, line 137: Did you try running the command with use_gcloud_storage=True in boto config? I think it should error out.

dilipped: Things like these would also be caught by adding a shim translation test like this: https://github.com/GoogleCloudPlatform/gsutil/blob/8cbf3362f34becf358f2fcd60dc3f20bf08a9a17/gslib/tests/test_ls.py#L194

dilipped (changes requested):
I think the implementation is incorrect. Feel free to reach out if you have any questions.

dilipped:
This is being implemented in a separate PR. Closing this one.

PR #1622, opened 2022-11-15
 Update the service accounts links

nique905:
This fix updates the service accounts link as the current compute/docs/authentication redirect link is being updated to follow the new authentication template and would break this reference point. Also fixes a few descriptions and product name for Compute Engine.

### DIFF
--- a/gslib/addlhelp/creds.py
+++ b/gslib/addlhelp/creds.py
@@ -109,8 +109,8 @@ _DETAILED_HELP_TEXT = ("""
   Compute Engine Internal Service Account:
     This is the type of service account used for accounts hosted by App Engine
     or Compute Engine. Such credentials are created automatically for
-    you on Compute Engine when you run the "gcloud compute instances
-    create" command and the credentials can be controlled with the "--scopes"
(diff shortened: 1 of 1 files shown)


gslib/addlhelp/creds.py
```diff
@@ -106,15 +106,16 @@
...
     https://developers.google.com/console/help/#DifferentRoles
 
-  Google Compute Engine Internal Service Account:
+  Compute Engine Internal Service Account:
     This is the type of service account used for accounts hosted by App Engine
-    or Google Compute Engine. Such credentials are created automatically for
-    you on Google Compute Engine when you run the gcloud compute instances
-    creates command and the credentials can be controlled with the --scopes
+    or Compute Engine. Such credentials are created automatically for
+    you on Compute Engine when you run the "gcloud compute instances
```
starsandskies, line 112: For code blocks, use double backticks (``) to begin and end the block

nique905: Thanks, updated.

thomasmaclean (approved):

Issue #1615, opened 2022-10-28
gsutil regex - not the same behaviour since last version (v5.16)


4sushi:
Hello, since yesterday (2022-10-27) with the upgrade of gsutil to version **5.16**, we have some issues with gsutil command. That impact most of our CICD jobs, on Github Actions.

This command with regex expression do not have the same behaviour anymore:
```shell
gsutil -m rsync -c -d -r -x "\..*|.*/\.[^/]*$|.*/\..*/.*$|_.*" . ${{ secrets.BUCKET }}/project
```

Result with gsutil **5.16** (latest):
```
Building synchronization state...
Skipping excluded path ./packages.yml...
Skipping excluded path ./e0859b964bdbc629ad7f2f98...
Skipping excluded path ./makefile...
```

Result with gsutil **5.15**:
```
Building synchronization state...
Starting synchronization...
Copying file://./8c6353775efb45e0657f2a11 [Content-Type=application/octet-stream]...
/ [0 files][    0.0 B/  3.8 KiB]                                                
/ [1 files][  3.8 KiB/  3.8 KiB]     
```

Do you have any idea of the reason of this change?

Thanks.

navi86:
@4sushi 
I have the same issue, except that it's working on version 5.14 and it doesn't work on version 5.15 in my case

tomonacci:
I just ran into this as well using 5.16. I think my particular problem is due to https://github.com/GoogleCloudPlatform/gsutil/pull/1602, which I got to by blaming https://github.com/GoogleCloudPlatform/gsutil/blob/2fd97591681a51ca0541d04b865e7d67a54efad4/gslib/commands/rsync.py#L734-L736

The merge commit of this PR is in https://github.com/GoogleCloudPlatform/gsutil/compare/v5.14...v5.15 so I think I will get the same problem with 5.15 (unverified).

I suspect that the format string should be something like `{}/(?:{})`. Otherwise, if alternation (`|`) is used as in @4sushi's example, the precedence dictates that the first component (`base_url_str`) binds with the LHS of `|`, which usually doesn't make sense (a subtle point is that if your RHS expression is unanchored, like `.*/node_modules/.*`, it _does_ work). Also, the use of `^` doesn't make sense anymore either, which was another source of breakage for me (although at this point I'm not sure if my use of `^` was ever legitimate... anyways, before it was working as I intended).

So the current workaround seems to be

[...]

PR #1623, opened 2022-11-16
Fixes -x flag to address issue #1615

thomasmaclean:
#1615 for context.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -2766,32 +2766,28 @@ class TestRsync(testcase.GsUtilIntegrationTestCase):
 
     # Use @Retry as hedge against bucket listing eventual consistency.
     @Retry(AssertionError, tries=3, timeout_secs=1)
-    def _Check1():
+    def _check_exclude_regex(regex, want):
       """Tests rsync skips the excluded pattern."""
       bucket_uri = self.CreateBucket()
-      regex = 'data.[/\\\\].*\\.txt$'
       # Add a trailing slash to the source directory to ensure its removed.
       local = tmpdir + ('\\' if IS_WINDOWS else '/')
       self.RunGsUtil(['rsync', '-r', '-x', regex, local, suri(bucket_uri)])
 
-      listing1 = TailSet(tmpdir, self.FlatListDir(tmpdir))
-      listing2 = TailSet(suri(bucket_uri), self.FlatListBucket(bucket_uri))
+      listing = TailSet(tmpdir, self.FlatListDir(tmpdir))
       self.assertEquals(
-          listing1,
+          listing,
           set(['/a', '/b', '/c', '/data1/a.txt', '/data1/ok', '/data2/b.txt']))
-      self.assertEquals(listing2, set(['/a', '/b', '/c', '/data1/ok']))
-
-    # Use @Retry as hedge against bucket listing eventual consistency.
-    @Retry(AssertionError, tries=3, timeout_secs=1)
+      got = TailSet(suri(bucket_uri), self.FlatListBucket(bucket_uri))
+      self.assertEquals(got, want)
+    
+    def _Check1():
+      _check_exclude_regex('data.[/\\\\].*\\.txt$', set(['/a', '/b', '/c', '/data1/ok']))
     def _Check2():
-      """Tests that a regex with a pipe works as expected."""
-      bucket_uri = self.CreateBucket()
-      regex = '.*\\.txt$|^[abc]'
-      self.RunGsUtil(['rsync', '-r', '-x', regex, tmpdir, suri(bucket_uri)])
-      listing = TailSet(suri(bucket_uri), self.FlatListBucket(bucket_uri))
-      self.assertEquals(listing, set(['/data1/ok']))
+      _check_exclude_regex('^data|[bc]$', set(['/a']))
 
+    # Ensure the example exclude pattern from documentation works as expected.
     _Check1()
+    # Tests that a regex with a pipe works as expected.
     _Check2()
 
   @unittest.skipIf(IS_WINDOWS,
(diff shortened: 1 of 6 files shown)


gslib/wildcard_iterator.py
```diff
@@ -649,8 +651,12 @@ def __iter__(self, bucket_listing_fields=None):
     for filepath in filepaths:
       expanded_url = StorageUrlFromString(filepath)
       try:
-        if self.exclude_pattern is not None:
-          if self.exclude_pattern.match(str(expanded_url)):
+        if self.exclude_tuple is not None:
```
dilipped, line 654: Should the code from here be removed, now that we are filtering out things at iterator level? https://github.com/GoogleCloudPlatform/gsutil/blob/2fd97591681a51ca0541d04b865e7d67a54efad4/gslib/commands/rsync.py#L764

gslib/tests/test_rsync.py
```diff
@@ -2781,7 +2781,18 @@ def _Check1():
           set(['/a', '/b', '/c', '/data1/a.txt', '/data1/ok', '/data2/b.txt']))
       self.assertEquals(listing2, set(['/a', '/b', '/c', '/data1/ok']))
 
+    # Use @Retry as hedge against bucket listing eventual consistency.
+    @Retry(AssertionError, tries=3, timeout_secs=1)
+    def _Check2():
+      """Tests that a regex with a pipe works as expected."""
+      bucket_uri = self.CreateBucket()
+      regex = '.*\\.txt$|ok$'
```
dilipped, line 2789: Shouldn't it be `'.*\\.txt$|.*ok$'`? I'm very confused why it works currently, any idea?

 Maybe also add a test for `^`, because that's the main reason we had to go back to the tuple solution.

thomasmaclean: It works because there's no `^`...! `^ok$` wouldn't work, but here I'm just saying exclude anything that ends in "ok". But yes, we should include a test for `^`, but I would argue one that doesn't follow with a `.*` so we can confirm it's not checking the base URL.

gslib/wildcard_iterator.py
thomasmaclean: Yes it should...

thomasmaclean: It turns out we need both. It turns out `wildcard_iterator` is excluding directories while `rsync` is excluding files. I've added a comment to this effect to make it more clear.

gslib/wildcard_iterator.py
```diff
@@ -649,8 +651,12 @@ def __iter__(self, bucket_listing_fields=None):
...
+          (base_url_str, exclude_pattern) = self.exclude_tuple
+          str_to_check = filepath[len(base_url_str):]
```
dilipped, line 656: It seems like base_url_str is a StorageUrl.url_string, which means it will have the scheme prefix i.e `file://`. The `filepath` here is probably just the output of glob, so will be without any scheme, and hence the str_to_check will be incorrect. You might want to use the `expanded_url` instead.

I'm guessing this was probably why the `.*txt$|ok` was working before.

gslib/wildcard_iterator.py
dilipped: Perhaps the bug mentioned in the comment below was the reason it was needed in rsync. After the fix, I think it should be fine to avoid doing the filtering in `rsync.py`? However, given that we are only doing this in wildcard iterator for file urls, we might still want to use the exclude patterns for cloud urls. 

I think at some point we should move the filtering for cloud urls inside wildcard iterator as well, but that might not be worth doing in gsutil, maybe for gcloud storage. 

For now, I think we can do this in `rsync.py`?
```
if cls.exclude_pattern and not url.IsFileUrl():
  ... 
   ...
```

gslib/wildcard_iterator.py
thomasmaclean: Good catch!

thomasmaclean: I remember this was copying the previous implementation very closely so I'll want to test this a bunch before I make further changes.

gslib/tests/test_rsync.py
```diff
@@ -2760,28 +2760,37 @@ def _Check2():
...
-      self.assertEquals(listing2, set(['/a', '/b', '/c', '/data1/ok']))
+      got = TailSet(suri(bucket_uri), self.FlatListBucket(bucket_uri))
+      self.assertEquals(got, want)
+
+    def _Check1():
+      _check_exclude_regex('data.[/\\\\].*\\.txt$',
+                           set(['/a', '/b', '/c', '/data1/ok']))
 
+    def _Check2():
+      _check_exclude_regex('^data|[bc]$', set(['/a']))
```
dilipped, line 2788: I think even if you do not have any of the wildcard iterator filtering logic, this test will still pass? That's because rsync.py will always do the filtering. That made me realize that we never had a test case that can reproduce the symlink issue from the original bug. That's because as long as the files are valid, the existing logic present in rsync.py will always work. One way to verify would be to check for the `Skipping excluded path` string in the logs to ensure that the directory that is excluded never gets expanded (i.e the files underneath the dirs are never acted upon). I would still remove the duplicate filtering logic though, i.e skip the filtering in rsync.py if it is FileUrl.

dilipped (approved):

gslib/wildcard_iterator.py
```diff
@@ -697,12 +693,26 @@ def _IterDir(self, directory, wildcard):
...
+                                                topdown=True):
+      filtered_dirnames = []
+
+      for dirname in dirnames:
+        full_dir_path = os.path.join(dirpath, dirname)
+        # Removes directories in place to prevent them and their children from
+        # being iterated. See https://docs.python.org/3/library/os.html#os.walk
+        if not self._ExcludeDir(full_dir_path):
+          filtered_dirnames.append(dirname)
+        # If a symlink is excluded above we don't want to print 2 messages.
```
dilipped, line 706: Nice! nit: maybe move the comment under the else block, makes it easier to associate.

dilipped (approved):

dilipped (approved):

PR #1627, opened 2022-11-29
Shim iam command

NickGoog:


gslib/commands/iam.py
```diff
@@ -309,6 +316,55 @@ class IamCommand(Command):
       },
   )
 
+  def get_gcloud_storage_args(self):
+    sub_command = self.args.pop(0)
```
thomasmaclean, line 320: Not familiar with how sub-args are parsed by the shim, but it looks like the idea is to mutate `self.args` here so that `get_gcloud_storage_args` doesn't see the sub command?

gslib/tests/test_iam.py
```diff
@@ -1393,3 +1422,136 @@ def test_set_valid_iam_multithreaded_single_object(self):
...
+  @mock.patch.object(iam.IamCommand, 'RunCommand', new=mock.Mock())
+  def test_shim_translates_iam_set_object(self):
+    with SetBotoConfigForTest([('GSUtil', 'use_gcloud_storage', 'True'),
+                               ('GSUtil', 'hidden_shim_mode', 'dry_run')]):
+      with SetEnvironmentForTest({
+          'CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL': 'True',
+          'CLOUDSDK_ROOT_DIR': 'fake_dir',
+      }):
+        mock_log_handler = self.RunCommand(
+            'iam', ['set', 'policy-file', 'gs://b/o1', 'gs://b/o2'],
```
thomasmaclean, line 1470: Might be good to throw in a test with a mix of buckets and objects here since there is logic in the shim mapping function that throws that error.

thomasmaclean (approved):

gslib/commands/iam.py
NickGoog: Yes!

gslib/tests/test_iam.py
NickGoog: It's a few tests down

gslib/commands/iam.py
```diff
@@ -309,6 +316,55 @@ class IamCommand(Command):
...
+          flag_map={
+              '-a': GcloudStorageFlag('--all-versions'),
+              '-e': GcloudStorageFlag('--etag'),
+              # Hack to make gcloud recognize empty string value.
+              '_empty_etag': GcloudStorageFlag('--etag='),
+              '-f': GcloudStorageFlag('--continue-on-error'),
+              '-R': GcloudStorageFlag('--recursive'),
+              '-r': GcloudStorageFlag('--recursive'),
+          })
+      # Flags must be at the start of self.args to get parsed.
```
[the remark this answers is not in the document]
NickGoog: No, this isn't referring to that bug. Flags must be at the start of `gsutil iam` b/c there's a variable number of cloud URLs. The structure is `gsutil iam set [FLAGS] [FILE URL] [CLOUD URLS...]`

If a flag isn't directly after "set" it will be incorrectly parsed as a URL.

```
$ gsutil iam set ~/Desktop/iam.json gs://bucket -r
gs://0noogler-kingdom/
file://-r
'_FileUrl' object has no attribute 'IsBucket'
$ gsutil iam set ~/Desktop/iam.json -r gs://bucket
file://-r
'_FileUrl' object has no attribute 'IsBucket'
```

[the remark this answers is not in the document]
dilipped: Got it, thanks!

dilipped (approved):

dilipped:
@NickGoog  This PR is causing `gslib.tests.test_iam.TestIamSet.test_set_force_error` to fail.

NickGoog:
@dilipped , fix looks simple. Replace `BucketNotFoundError` with `CommandError`.

NickGoog:
https://github.com/GoogleCloudPlatform/gsutil/pull/1652

PR #1628, opened 2022-11-30
Pin httplib2 to 0.20.4 version

dilipped:
Tested on my virtual env

```
$ pip freeze | grep httplib2
# nothing

$ pip install .

$ pip freeze | grep httplib2
httplib2==0.20.4
```


rrauber (approved):

dilipped:
The kokoro failures are because of the ongoing issues with kms test.
There is an rsync test failure for Windows and we have a separate PR with the fix. Seems like an issue with the test itself.
There is a MacOS issue where python 3.5 cannot be installed.

PR #1634, opened 2022-12-07
Shim support for additional headers.

rrauber:


gslib/tests/test_shim_util.py
```diff
@@ -993,7 +1015,6 @@ def test_get_flag_from_header_with_unset_true_for_metadata_headers(self):
     headers_to_expected_flag_map = {
         'x-goog-meta-foo': '--remove-custom-metadata=foo',
         'x-amz-meta-foo': '--remove-custom-metadata=foo',
-        'x-amz-foo': '--remove-custom-headers=x-amz-foo',
```
dilipped, line 996: How are we handing the unset part ? Do we have any `--remov-custom-headers` equivalent?

rrauber: Now I'm wondering if this is actually something we have to support -- let me know if I'm missing something. From what I could tell, the only real place where separate logic for removing headers is used is in the `setmeta` command, which should already be handled with the --remove-custom-metadata flag, since doing this requires specific support for specific fields anyway (e.g., for XML APIs: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/boto_translation.py#L822). Also, just adding gsutil -h 'Content-Type' cp src dest didn't seem to change any content type headers.

dilipped: I think this go/gcloud-storage-object-metadata-flags proposed it, and that's why we had it in the boto translation. But I do not see it being implemented so my guess is that we later decided to not have a flag for custom headers. Maybe check with Nick, but looks like we should be good.

dilipped (approved):

PR #1638, opened 2022-12-14
ACL shim translator

LuisFGarciaM:
Added shim tranlsator for GCS acl-file and test.

### DIFF
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -21,6 +21,7 @@ from __future__ import unicode_literals
 
 from apitools.base.py import encoding
 from gslib import metrics
+
 from gslib.cloud_api import AccessDeniedException
 from gslib.cloud_api import BadRequestException
 from gslib.cloud_api import PreconditionException
@@ -295,14 +296,6 @@ _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _ch_help_text = CreateHelpText(_CH_SYNOPSIS, _CH_DESCRIPTION)
 
-_GET_COMMAND_BUCKET = GcloudStorageMap(
-  gcloud_command = ['alpha', 'storage', 'bucket', '--format=json'],
-  flag_map = GcloudStorageFlag('--acl-file'))
-
-_GET_COMMAND_OBJECT = GcloudStorageMap(
-  gcloud_command = ['alpha', 'storage', 'object', '--format=json'],
-  flag_map = GcloudStorageFlag('--acl-file'))
-
 def _ApplyExceptionHandler(cls, exception):
   cls.logger.error('Encountered a problem: %s', exception)
   cls.everything_set_okay = False
@@ -354,38 +347,35 @@ class AclCommand(Command):
     _CannedGcsAcl = ['private', 'public-read', 'project-private',
                     'public-read-write', 'authenticated-read',
                     'bucket-owner-read', 'bucket-owner-full-control']
-    
-    sub_command = self.args.pop(0)
-    if sub_command == 'set':
-      if StorageUrlFromString(self.args[0]).IsObject():
+    if self.args[0] == 'get':
+      if StorageUrlFromString(self.args[1]).IsObject():
         command_group = 'objects'
-      else:
+      elif StorageUrlFromString(self.args[1]).IsBucket():
         command_group = 'buckets'
-      if StorageUrlFromString(self.args[1]) in _CannedGcsAcl:
-        predef_acl= '--predefined-acl='+ str(self.args[1])
-      else: 
-        predef_acl='--acl-file='+ str(self.args[1])
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+            'alpha', 'storage', command_group, 'describe', '--format=json'
+        ],
+                                              flag_map={})
+    elif self.args[0] == 'set':
+      if StorageUrlFromString(self.args[2]).IsBucket():
+        command_group = 'buckets'
+      elif StorageUrlFromString(self.args[2]).IsObject():
+        command_group = 'objects'
+      if self.args[1] in _CannedGcsAcl:
+        url_with_acl= '--predefined-acl=' + str(self.args[1])
+      elif self.args[1] not in _CannedGcsAcl:
+        url_with_acl='--acl-file=' + str(self.args[1])
       gcloud_storage_map = GcloudStorageMap(gcloud_command={
         'set': GcloudStorageMap(
-          gcloud_command=['alpha', 'storage', command_group, 'update', predef_acl],
+          gcloud_command=['alpha', 'storage', command_group, 'update', url_with_acl],
           flag_map={
             'a': GcloudStorageFlag('--all-versions'),
             'f': GcloudStorageFlag('--continue-on-error'),
             'r': GcloudStorageFlag('--recursive'),
             'R': GcloudStorageFlag('--recursive'),
           })
-      }, flag_map={})
-    elif sub_command =='get':
-      if StorageUrlFromString(self.args[0]).IsObject():
-        command_group = 'objects'
-      else:
-        command_group = 'buckets'
-      gcloud_storage_map = GcloudStorageMap(gcloud_command={
-        'get': GcloudStorageMap(
-          gcloud_command=['alpha', 'storage', command_group, 'describe', '--format=json[acl]'],
-          flag_map={}
-          )}, flag_map={})
-          
+        }, 
+                                              flag_map={})
     return super().get_gcloud_storage_args(gcloud_storage_map)
 
   def _CalculateUrlsStartArg(self):
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -296,6 +296,10 @@ _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _ch_help_text = CreateHelpText(_CH_SYNOPSIS, _CH_DESCRIPTION)
 
+CannedGcsAcl = set(('private', 'public-read', 'project-private',
+                    'public-read-write', 'authenticated-read',
+                    'bucket-owner-read', 'bucket-owner-full-control'))
+
 def _ApplyExceptionHandler(cls, exception):
   cls.logger.error('Encountered a problem: %s', exception)
   cls.everything_set_okay = False
@@ -342,14 +346,12 @@ class AclCommand(Command):
           'ch': _ch_help_text
       },
   )
-  CannedGcsAcl = ['private', 'public-read', 'project-private',
-                    'public-read-write', 'authenticated-read',
-                    'bucket-owner-read', 'bucket-owner-full-control']
+
   def get_gcloud_storage_args(self):
     if self.args[0] == 'get':
       url = self.args[1]
-      object_or_bucket = StorageUrlFromString(url)
-      if object_or_bucket.IsObject():
+      object_or_bucket_url = StorageUrlFromString(url)
+      if object_or_bucket_url.IsObject():
         command_group = 'objects'
       else:
         command_group = 'buckets'
@@ -360,15 +362,15 @@ class AclCommand(Command):
    
     elif self.args[0] == 'set':
       url = self.args[2]
-      object_or_bucket = StorageUrlFromString(url)
-      if object_or_bucket.IsObject():
+      object_or_bucket_url = StorageUrlFromString(url)
+      if object_or_bucket_url.IsObject():
         command_group = 'objects'
       else:
         command_group = 'buckets'
-      if self.args[1] in AclCommand.CannedGcsAcl:
-        acl_flag= '--predefined-acl=' + str(self.args[1])
+      if self.args[1] in CannedGcsAcl:
+        acl_flag= '--predefined-acl=' + self.args[1]
       else:
-        acl_flag='--acl-file=' + str(self.args[1])
+        acl_flag='--acl-file=' + self.args[1]
       gcloud_storage_map = GcloudStorageMap(gcloud_command={
         'set': GcloudStorageMap(
           gcloud_command=['alpha', 'storage', command_group, 'update', url, acl_flag],
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
+
 import os
 
 from apitools.base.py import encoding
@@ -156,7 +157,7 @@ _CH_DESCRIPTION = """
   such objects. For help doing this, see "gsutil help setmeta".
 
   Grant anyone on the internet WRITE access to the bucket example-bucket:
-  
+
   WARNING: this is not recommended as you will be responsible for the content
 
     gsutil acl ch -u AllUsers:W gs://example-bucket
@@ -296,7 +297,6 @@ _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _ch_help_text = CreateHelpText(_CH_SYNOPSIS, _CH_DESCRIPTION)
 
-
 def _ApplyExceptionHandler(cls, exception):
   cls.logger.error('Encountered a problem: %s', exception)
   cls.everything_set_okay = False
@@ -371,7 +371,7 @@ class AclCommand(Command):
           recurse = True
           break
       RaiseErrorIfUrlsAreMixOfBucketsAndObjects(object_or_bucket_urls, recurse)
-      
+
       if object_or_bucket_urls[0].IsBucket() and not recurse:
         command_group = 'buckets'
       else:
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -363,7 +363,7 @@ class AclCommand(Command):
     elif self.args[0] == 'set':
       if len(self.args[2:])>1:
         url_list = self.args[2:]
-        object_or_bucket_url = [StorageUrlFromString(i) for i in url]
+        object_or_bucket_url = [StorageUrlFromString(i) for i in url_list]
       else:
         url_list =[self.args[2]]
         object_or_bucket_url = [StorageUrlFromString(self.args[2])]
--- a/gslib/commands/acl.py
+++ b/gslib/commands/acl.py
@@ -342,32 +342,36 @@ class AclCommand(Command):
           'ch': _ch_help_text
       },
   )
-
-  def get_gcloud_storage_args(self):
-    _CannedGcsAcl = ['private', 'public-read', 'project-private',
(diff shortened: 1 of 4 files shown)


gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
       },
   )
 
+  def get_gcloud_storage_args(self):
+    _CannedGcsAcl = ['private', 'public-read', 'project-private',
+                    'public-read-write', 'authenticated-read',
+                    'bucket-owner-read', 'bucket-owner-full-control']
+    if self.args[0] == 'get':
+      if StorageUrlFromString(self.args[1]).IsObject():
```
NickGoog, line 351: Can you create one `StorageUrl` instead of generating two? Ditto below

gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
       },
   )
 
+  def get_gcloud_storage_args(self):
+    _CannedGcsAcl = ['private', 'public-read', 'project-private',
```
NickGoog, line 347: Is there already a constant for this somewhere? Why redeclare the constant in each run of this function?

gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
...
+                    'public-read-write', 'authenticated-read',
+                    'bucket-owner-read', 'bucket-owner-full-control']
+    if self.args[0] == 'get':
+      if StorageUrlFromString(self.args[1]).IsObject():
+        command_group = 'objects'
+      elif StorageUrlFromString(self.args[1]).IsBucket():
+        command_group = 'buckets'
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+            'alpha', 'storage', command_group, 'describe', '--format=json'
```
NickGoog, line 356: This format string will return every detail of the object or bucket, not just the ACL

gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
...
+        ],
+                                              flag_map={})
+    elif self.args[0] == 'set':
+      if StorageUrlFromString(self.args[2]).IsBucket():
+        command_group = 'buckets'
+      elif StorageUrlFromString(self.args[2]).IsObject():
+        command_group = 'objects'
+      if self.args[1] in _CannedGcsAcl:
+        url_with_acl= '--predefined-acl=' + str(self.args[1])
+      elif self.args[1] not in _CannedGcsAcl:
```
NickGoog, line 366: This looks like it can just be `else`

gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
...
   )
 
+  def get_gcloud_storage_args(self):
+    _CannedGcsAcl = ['private', 'public-read', 'project-private',
+                    'public-read-write', 'authenticated-read',
+                    'bucket-owner-read', 'bucket-owner-full-control']
+    if self.args[0] == 'get':
+      if StorageUrlFromString(self.args[1]).IsObject():
+        command_group = 'objects'
+      elif StorageUrlFromString(self.args[1]).IsBucket():
```
NickGoog, line 353: `command_group` will be trigger an undeclared variable error if the arg isn't a bucket or object

gslib/commands/acl.py
```diff
@@ -340,6 +343,41 @@ class AclCommand(Command):
...
+        command_group = 'buckets'
+      elif StorageUrlFromString(self.args[2]).IsObject():
+        command_group = 'objects'
+      if self.args[1] in _CannedGcsAcl:
+        url_with_acl= '--predefined-acl=' + str(self.args[1])
+      elif self.args[1] not in _CannedGcsAcl:
+        url_with_acl='--acl-file=' + str(self.args[1])
+      gcloud_storage_map = GcloudStorageMap(gcloud_command={
+        'set': GcloudStorageMap(
+          gcloud_command=['alpha', 'storage', command_group, 'update', url_with_acl],
```
NickGoog, line 370: Please re-examine the syntax of the `buckets|objects update` command

gslib/tests/test_acl.py
```diff
@@ -18,24 +18,36 @@
 from __future__ import print_function
```
NickGoog, line 18: I'll look at the tests after the core logic is fixed

gslib/commands/acl.py
```diff
@@ -339,6 +342,45 @@ class AclCommand(Command):
...
+  def get_gcloud_storage_args(self):
+    if self.args[0] == 'get':
+      url = self.args[1]
+      object_or_bucket = StorageUrlFromString(url)
+      if object_or_bucket.IsObject():
+        command_group = 'objects'
+      else:
+        command_group = 'buckets'
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+            'alpha', 'storage', command_group, 'describe', url, '--format=json[acl]'
```
NickGoog, line 357: Try testing this command in the terminal to see if the shim output matches the gsutil output without the shim

gslib/commands/acl.py
```diff
@@ -339,6 +342,45 @@ class AclCommand(Command):
...
+   
+    elif self.args[0] == 'set':
+      url = self.args[2]
+      object_or_bucket = StorageUrlFromString(url)
+      if object_or_bucket.IsObject():
+        command_group = 'objects'
+      else:
+        command_group = 'buckets'
+      if self.args[1] in AclCommand.CannedGcsAcl:
+        acl_flag= '--predefined-acl=' + str(self.args[1])
```
NickGoog, line 369: Is `str(...)` necessary?

gslib/commands/acl.py
```diff
@@ -339,6 +342,45 @@ class AclCommand(Command):
           'ch': _ch_help_text
       },
   )
+  CannedGcsAcl = ['private', 'public-read', 'project-private',
```
NickGoog, line 345: Why make a static variable on the class?

gslib/commands/acl.py
```diff
@@ -339,6 +342,45 @@ class AclCommand(Command):
...
+        command_group = 'objects'
+      else:
+        command_group = 'buckets'
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+            'alpha', 'storage', command_group, 'describe', url, '--format=json[acl]'
+        ],
+                                              flag_map={})
+   
+    elif self.args[0] == 'set':
+      url = self.args[2]
```
NickGoog, line 362: There can be more than one URL.
https://cloud.google.com/storage/docs/gsutil/commands/acl

gslib/commands/acl.py
```diff
@@ -339,6 +342,45 @@ class AclCommand(Command):
...
+      object_or_bucket = StorageUrlFromString(url)
```
NickGoog, line 363: Might want to rename this `object_or_bucket_url`

gslib/commands/acl.py
LuisFGarciaM: Here is the output when I try both commands. https://docs.google.com/document/d/16cLV99FBR3Wymv74Aaj0iIVeY_HOyYUE3YUOg7kew1I/edit?resourcekey=0-2vih9Q9RWQgVjg6I5LP-zg

gslib/commands/acl.py
NickGoog: Use the immutable version of `set` please. Also, look at how constants are normally named.

gslib/commands/acl.py
```diff
@@ -340,6 +346,50 @@ class AclCommand(Command):
...
+    if sub_command == 'get':
+      if StorageUrlFromString(self.args[0]).IsObject():
+        command_group = 'objects'
+      else:
+        command_group = 'buckets'
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+          'alpha', 'storage', command_group, 'describe', self.args[0], '--format=json[acl]'
+      ],
+                                            flag_map={})
+      self.args = self.args[:-1]
```
NickGoog, line 360: I think we can keep the last argument, which is `url` according to http://cloud/storage/docs/gsutil/commands/acl

```
 gcloud_storage_map = GcloudStorageMap(gcloud_command=[
          'alpha', 'storage', command_group, 'describe', '--format=json[acl]'
      ],
                                            flag_map={})
```

Removing `self.args[0]` from the current translation should cause the `url` to get translated and added to the end of the command, resulting in `... describe --format=json[acl] url`

gslib/commands/acl.py
```diff
@@ -340,6 +346,50 @@ class AclCommand(Command):
...
+    
+    elif sub_command == 'set':
+      if len(self.args[1:])>1:
+        url_list = self.args[1:]
+        object_or_bucket_url = [StorageUrlFromString(i) for i in url_list]
```
NickGoog, line 365: It doesn't matter if there are one or more arguments. This code should work for both cases.

gslib/commands/acl.py
```diff
@@ -340,6 +346,50 @@ class AclCommand(Command):
...
+      else:
+        url_list =[self.args[1]]
+        object_or_bucket_url = [StorageUrlFromString(self.args[1])]
+      if object_or_bucket_url[0].IsBucket():
+        command_group = 'buckets'
+      else:
+        command_group = 'objects'
+      if self.args[0] in CANNED_GCS_ACLS:
+        acl_flag= '--predefined-acl=' + self.args[0]
```
NickGoog, line 374: Looks like you need a space before `=`
Ditto below

gslib/commands/acl.py
```diff
@@ -340,6 +346,50 @@ class AclCommand(Command):
...
+              'alpha', 'storage', command_group, 'update'] + url_list +  [acl_flag],
+          flag_map={
+              '-a': GcloudStorageFlag('--all-versions'),
+              '-f': GcloudStorageFlag('--continue-on-error'),
+              '-R': GcloudStorageFlag('--recursive'),
+              '-r': GcloudStorageFlag('--recursive'),
+          })
+      
+      args_to_be_removed = len(url_list) + 1 
+      self.args = self.args[:-args_to_be_removed]
```
NickGoog, line 388: Does this just leaves the flags? Might want to add a code comment, so people know what's left in `self.args` that the shim translates

gslib/commands/acl.py
NickGoog: Looks like you switched to a set, but didn't make it immutable with `frozenset`

gslib/commands/acl.py
NickGoog: Since it's now plural, I'd make this `object_or_bucket_urls`

gslib/tests/test_acl.py
```diff
@@ -760,3 +769,181 @@ class TestAclOldAlias(TestAcl):
...
+  def test_shim_translates_acl_set_multiple_buckets_urls_recursive_all_versions(
+      self):
+    with SetBotoConfigForTest([('GSUtil', 'use_gcloud_storage', 'True'),
+                               ('GSUtil', 'hidden_shim_mode', 'dry_run')]):
+      with SetEnvironmentForTest({
+          'CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL': 'True',
+          'CLOUDSDK_ROOT_DIR': 'fake_dir',
+      }):
+        mock_log_handler = self.RunCommand('acl', [
+            'set', '-r', '-a', 'acl-file', 'gs://bucket', 'gs://bucket1',
```
NickGoog, line 926: Change one of these to an object URL to test that mix of buckets and objects is allowed if recursion is enabled.

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,38 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
     self.assertIsNone(storage_url.UrlsAreMixOfBucketsAndObjects(urls))
+
+  def test_urls_raise_error_if_mix_of_objects_and_buckets(self):
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b1', 'gs://b/o']))
+    with self.assertRaisesRegex(
+            CommandException,
+            'Cannot operate on a mix of buckets and objects.'):
+      storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False)
+
+  def test_urls_raise_error_if_objects_and_buckets_are_mixed_different(self):
```
NickGoog, line 147: Instead of "_different", you can call the first test "bucket_followed_by_object" and the second "object_followed_by_bucket"

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,38 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
+            CommandException,
+            'Cannot operate on a mix of buckets and objects.'):
+      storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False)
+
+  def test_accepts_mix_of_objects_and_buckets_if_recursion_requested(self):
+    # No error raised
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b1', 'gs://b/o']))
+    storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=True)
+  
+  def test_not_raising_error_if_mix_objects_with_recursion(self):
```
NickGoog, line 159: Can remove this. It's not testing any code paths different from the above.

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,38 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b/p']))
+    self.assertFalse(storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=True))
+
+  def test_not_raising_error_if_mix_objects_without_recursion(self):
```
NickGoog, line 163: Would call this "multiple_objects" instead of "mix" b/c we're using "mix" to refer to a combination of buckets and objects in all the other tests. Ditto for the buckets test below.

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,38 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b/p']))
+    self.assertFalse(storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False))
+  
+  def test_not_raising_error_if_mix_buckets_with_recursion(self):
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b/p']))
+    self.assertFalse(storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=True))
+  
+  def test_not_raising_error_if_mix_objects_with_recursion(self):
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b/p']))
+    self.assertFalse(storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False))
```
NickGoog, line 173: Should `recursion_requested=True` here?

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,34 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
+      storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False)
+
+  def test_accepts_mix_of_objects_and_buckets_if_recursion_requested(self):
+    # No error raised
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b1', 'gs://b/o']))
+    storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=True)
+
+  def test_not_raising_error_if_multiple_objects_without_recursion(self):
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b/p']))
+    self.assertFalse(storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False))
```
NickGoog, line 161: Don't need to `assertFalse` on these. We're implicitly testing an error wasn't raised.

gslib/tests/test_storage_url.py
```diff
@@ -135,3 +136,34 @@ def test_urls_are_mix_of_objects_and_buckets_is_true_for_a_mix(self):
...
+
+  def test_urls_raise_error_if_object_followed_by_bucket(self):
+    urls = list(map(storage_url.StorageUrlFromString, ['gs://b/o', 'gs://b']))
+    with self.assertRaisesRegex(
+            CommandException,
+            'Cannot operate on a mix of buckets and objects.'):
+      storage_url.RaiseErrorIfUrlsAreMixOfBucketsAndObjects(urls, recursion_requested=False)
+
+  def test_accepts_mix_of_objects_and_buckets_if_recursion_requested(self):
+    # No error raised
```
NickGoog, line 155: Add period

gslib/tests/test_storage_url.py
NickGoog: Looks like you missed some

gslib/commands/acl.py
```diff
@@ -340,6 +350,51 @@ class AclCommand(Command):
...
+      gcloud_storage_map = GcloudStorageMap(gcloud_command=[
+          'alpha', 'storage', command_group, 'describe',
+          '--format=multi(acl:format=json)'
+      ],
+                                            flag_map={})
+
+    elif sub_command == 'set':
+      # Flags must be at the start of self.args to get parsed.
+      self.ParseSubOpts()
+      if self.args[0] in CANNED_GCS_ACLS:
```
dilipped, line 369: looks like the gsutil acl command relies on checking if the argument is a file, which is much easier and also safer (prevents the failure for cases where a file name is one of the canned acl names) I think .
https://github.com/GoogleCloudPlatform/gsutil/blob/6fee45bb59c6124524d3930a5daca88a701ccd9c/gslib/command.py#L1101

dilipped (commented):
LGTM overall. Just the check for predefined acl can be changed to just check if it is a file.

dilipped:
Please make sure to run the parity_check.py script before merging.

gslib/tests/test_acl.py
```diff
@@ -760,3 +769,181 @@ class TestAclOldAlias(TestAcl):
...
+                                           return_log_handler=True)
+        info_lines = '\n'.join(mock_log_handler.messages['info'])
+        self.assertIn(
+            ('Gcloud Storage Command: {} alpha storage objects update'
+             ' --acl-file=acl-file --recursive --all-versions gs://bucket'
+             ' gs://bucket1/o gs://bucket2').format(
+                 os.path.join('fake_dir', 'bin', 'gcloud')), info_lines)
+
+  @mock.patch.object(acl.AclCommand, 'RunCommand', new=mock.Mock())
+  def test_shim_translates_acl_set_mix_buckets_and_objects_raises_error(self):
```
dilipped, line 938: Are we missing a test where there are mix of objects and buckets and recursion is requested?

LuisFGarciaM: I think this test case is in the test_storage_url.py file, line 154. Since that was done through a util I added the test case in that file.

NickGoog: It exists `test_shim_translates_acl_set_multiple_buckets_urls_recursive_all_versions`

dilipped (approved):

PR #1645, opened 2022-12-22
Adding shim support for custom dual regions

thomasmaclean:
The mb tests have been modified so that they pass with the shim turned on, and I also fixed a few tests not related to CDR.

### DIFF
--- a/gslib/tests/test_mb.py
+++ b/gslib/tests/test_mb.py
@@ -308,7 +308,7 @@ class TestMb(testcase.GsUtilIntegrationTestCase):
                                   suppress_consec_slashes=False)
     self.RunGsUtil(['mb', '-l', 'us-east1+us-east4', suri(bucket_uri)])
     stdout = self.RunGsUtil(['ls', '-Lb', suri(bucket_uri)], return_stdout=True)
-    self.assertRegex(stdout, r"locations:\t\t\[\'US-EAST1\', \'US-EAST4")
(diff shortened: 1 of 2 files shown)


dilipped (approved):
LGTM. Make sure to run the parity_check.py script before merging.

gslib/tests/test_mb.py
```diff
@@ -308,7 +303,7 @@ def test_create_with_custom_dual_regions_via_l_flag(self):
                                   suppress_consec_slashes=False)
     self.RunGsUtil(['mb', '-l', 'us-east1+us-east4', suri(bucket_uri)])
     stdout = self.RunGsUtil(['ls', '-Lb', suri(bucket_uri)], return_stdout=True)
-    self.assertRegex(stdout, r"locations:\t\t\[\'US-EAST1\', \'US-EAST4")
+    self.assertRegex(stdout, r"ocations:\s*\[\s*.US-EAST1.,\s*.US-EAST4")
```
dilipped, line 306: So I just realized that we have not added --format=gsutil to the ls command yet https://github.com/GoogleCloudPlatform/gsutil/blob/6fee45bb59c6124524d3930a5daca88a701ccd9c/gslib/commands/ls.py#L337 , because of which we need to make this adjustments. I

I  feel that if we just add `--format=gsutil` the formatting should be exactly like gsutil. Can you try that first?

PR #1646, opened 2022-12-22
Adjust test_list_public_access_prevention to work with shim

thomasmaclean:


gslib/tests/test_ls.py
```diff
@@ -1155,11 +1155,11 @@ def test_list_public_access_prevention(self):
     bucket_uri = self.CreateBucket()
     stdout = self.RunGsUtil(['ls', '-Lb', suri(bucket_uri)], return_stdout=True)
     self.assertRegex(stdout,
-                     r'Public access prevention:\t*(unspecified|inherited)')
+                     r'Public access prevention:\s*(unspecified|inherited)')
```
dilipped, line 1158: Shouldn't the gsutil formatter used in gcloud storage ensure that the formatting is same as gsutil.

dilipped: I think the flag was never added to the ls command. See my comment on the other PR: https://github.com/GoogleCloudPlatform/gsutil/pull/1645#pullrequestreview-1236663259

thomasmaclean: Even once this flag was added these tests still failed in the same way, so I'll keep these changes here.

PR #1649, opened 2023-01-04
Add shim translator for rpo flag in mb command

carlosmerec:
- Add rpo flag shim translation for mb command
- Add unit test to test shim translation
- Add rpo get shim translation for rpo get command
- Fixed e2e tests

### DIFF
--- a/gslib/commands/rpo.py
+++ b/gslib/commands/rpo.py
@@ -86,8 +86,7 @@ GET_COMMAND = GcloudStorageMap(gcloud_command=[
     '--format=value[separator=": "](name.sub("^", "gs://"),'
     'rpo.yesno(no="None"))'
 ],
-                               flag_map={},
-                               supports_output_translation=True)
+                               flag_map={})
 
 
 class RpoCommand(Command):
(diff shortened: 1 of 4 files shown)


gslib/commands/rpo.py
```diff
@@ -80,6 +81,14 @@
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 
+GET_COMMAND = GcloudStorageMap(gcloud_command=[
+    'storage', 'buckets', 'list',
+    '--format=value[separator=": "](name.sub("^", "gs://"),'
+    'rpo.yesno(no="None"))'
+],
+                               flag_map={},
+                               supports_output_translation=True)
```
dilipped, line 90: Maybe remove this. I think this argument should be removed from the definition itself, as it is not used anywhere. There was a plan to use it at one point, but later we decided that we are not going to support that level of granularity. Maybe file a bug (internally) to remove this value ?

dilipped (commented):
LGTM. Did you run the parity_check.py script for this already. If yes then can you please share the results here or offline. Thanks!

gslib/commands/rpo.py
carlosmerec: Done!

PR #1650, opened 2023-01-05
Shim translator rpo set

carlosmerec:
- Add unit test to test shim translation
   - Fix incorrect test test_get_returns_default_for_dual_region_bucket

Depends on merging this [pr](https://github.com/GoogleCloudPlatform/gsutil/pull/1649) first

### DIFF
--- a/gslib/commands/rpo.py
+++ b/gslib/commands/rpo.py
@@ -88,6 +88,13 @@ GET_COMMAND = GcloudStorageMap(gcloud_command=[
 ],
                                flag_map={})
 
+SET_COMMAND = GcloudStorageMap(
+    gcloud_command=[
+        'storage', 'buckets', 'update', '--recovery-point-objective'
+    ],
+    flag_map={},
+)
+
 
 class RpoCommand(Command):
   """Implements the gsutil rpo command."""
@@ -127,6 +134,7 @@ class RpoCommand(Command):
   gcloud_storage_map = GcloudStorageMap(
       gcloud_command={
           'get': GET_COMMAND,
+          'set': SET_COMMAND,
       },
       flag_map={},
   )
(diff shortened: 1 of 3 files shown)


gslib/tests/test_rpo.py
```diff
@@ -102,7 +144,7 @@ def _verify_get_returns_default_or_none(self, bucket_uri):
 
   @SkipForXML('RPO only runs on GCS JSON API.')
   def test_get_returns_default_for_dual_region_bucket(self):
-    bucket_uri = self.CreateBucket(location='us')
+    bucket_uri = self.CreateBucket(location='nam4')
```
dilipped, line 147: Curious about this change.

gslib/commands/rpo.py
```diff
@@ -80,6 +81,21 @@
 _set_help_text = CreateHelpText(_SET_SYNOPSIS, _SET_DESCRIPTION)
 _get_help_text = CreateHelpText(_GET_SYNOPSIS, _GET_DESCRIPTION)
 
+GET_COMMAND = GcloudStorageMap(gcloud_command=[
```
dilipped, line 84: I saw another PR covering this. Do you have chain of PRs. That's gonna be tricky in git.

dilipped (commented):
Please run the parity_check.py script and share the output

gslib/tests/test_rpo.py
carlosmerec: Test tries to create dual region bucket but it is passing the zone `us`  which creates a `multi-region` bucket, the test was passing due to only rpo config for `multi-region` bucket is `DEFAULT`, [reference](https://cloud.google.com/storage/docs/json_api/v1/buckets/patch), so when the assertion is made, returns `DEFAULT`, so I changed the code to actually create `dual-region` bucket using a [predefined dual-region](https://cloud.google.com/storage/docs/locations#predefined-dr)

gslib/commands/rpo.py
carlosmerec: Yes, I do, I'll have to rebase my first branch 😅

gslib/tests/test_rpo.py
dilipped: got it, thanks

PR #1653, opened 2023-01-09
ACL command shim fixes for test parity.

NickGoog:
Running tests via gcloud storage revealed a few issues.

These are all the fixes I could make without stripping the extra ACL object fields gcloud storage displays. Mostly just translating XML to JSON predefined ACLs.

Lots of lines edited by the yapf formatter.

### DIFF
--- a/gslib/tests/test_acl.py
+++ b/gslib/tests/test_acl.py
@@ -82,7 +82,11 @@ class TestAcl(TestAclBase):
     stderr = self.RunGsUtil(self._set_acl_prefix + [inpath, obj_uri],
                             return_stderr=True,
                             expected_status=1)
-    self.assertIn('ArgumentException', stderr)
+    if self._use_gcloud_storage:
+      error_text = 'JSONDecodeError'
+    else:
+      error_text = 'ArgumentException'
+    self.assertIn(error_text, stderr)
 
   def test_set_invalid_acl_bucket(self):
     """Ensures that invalid content returns a bad request error."""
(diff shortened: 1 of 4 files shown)


gslib/tests/test_acl.py
```diff
@@ -850,11 +870,11 @@ def test_shim_translates_predefined_acl_set_object(self):
           'CLOUDSDK_ROOT_DIR': 'fake_dir',
       }):
         mock_log_handler = self.RunCommand(
-            'acl', ['set', 'private', 'gs://bucket/object'],
+            'acl', ['set', 'publicRead', 'gs://bucket/object'],
```
dilipped, line 873: Any reason for changing this to publicRead? Does not matter much but I think private should have been fine as well?

NickGoog: `private` is the same string for XML and JSON, whereas hyphenated strings (e.g. `public-read`) show translation to JSON for gcloud storage.

gslib/tests/test_acl.py
```diff
@@ -100,8 +108,13 @@ def test_set_xml_acl_json_api_object(self):
     stderr = self.RunGsUtil(self._set_acl_prefix + [inpath, obj_uri],
                             return_stderr=True,
                             expected_status=1)
-    self.assertIn('ArgumentException', stderr)
-    self.assertIn('XML ACL data provided', stderr)
+
+    if self._use_gcloud_storage:
+      self.assertIn('JSONDecodeError', stderr)
+      # XML not currently supported in gcloud storage.
```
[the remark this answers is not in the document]
NickGoog: I think the error makes it clear they should be using JSON. Don't think it's worth it to catch the JSON error and try an XML parser to see if the user provided XML and raise a different error.

NickGoog:
Still discussing whether or not to make UX bugs for the error messages, but Dilip said he's fine merging other than that offline

Issue #1639, opened 2022-12-15
`gsutil -m` breaks "Reauthentication required"


thirdpwais3:
Let's suppose you want to copy a bunch of files to gcloud.  So you `gsutil cp -r my_local_dir gs://my-bucket/my_path`.  If you have  a lot of files, `gsutil` will print a big message telling you that you should actually be doing `gsutil -m cp` because of some such faster transfers etc.  (Note that `aws` will just use multi-upload mode automatically without nagging you).  Anyways, so then you get into a habit of using `gsutil -m cp` all the time because otherwise `gsutil` will nag you.  

Ok so one day log in and now do a `gsutil -m cp -r my_local_dir gs://my-bucket/my_path`.  Since it's been a day and your authorization needs to be refreshed, gsutil will print `"Reauthentication required"` and ask you for a password.  But since you did `gsutil -m cp`, the `gsutil` command clobbers itself and doesn't let you enter a password, instead filling your terminal with errors for each file, and leaving you wondering what went wrong.  

So it turns out if you just `gsutil cp -r my_local_dir gs://my-bucket/my_path`  (no `-m`) then you'll get single-threaded experience and `"Reauthentication required"` re-auth flow will work as expected.

But there are two issues here:
1) (this issue) `gsutil -m` will break the "Reauthentication required" flow, and it's really confusing what the user should do to fix.
2) (some future issue) `gsutil -m` should probably be the default, and not nag w/out it; `aws` just does the right thing and does not nag.

NickGoog:
What type of credentials are you using that require reauth? Trying to reproduce.

Did this start the day you posted 12/15/22?

In that case, this change was part of the last update (based on change log dates): https://github.com/GoogleCloudPlatform/gsutil/pull/1617

The reauth flow takes place mostly inside the dependency package so not sure how much we can do here.

thirdpwais3:
@NickGoog this issue has been happening forever

@thomasmaclean is gsutil just deprecated then?  i guess i should move to `gcloud storage` ?

NickGoog:
Hi, can you provide a stack trace by using the `-d` flag? `gsutil -d ...`? Be sure to remove any personal information before posting.

Migration to `gcloud storage` is encouraged by we're still missing some things in the new CLI like `rsync` and XML support.

thirdpwais3:
`This command is available in one or more alternate release tracks.  Try:
  gcloud alpha storage`

please see above, gcloud storage is a non-starter

[...]

PR #1655, opened 2023-01-10
Handle reauth challenges gracefully in transfer commands.

rrauber:
Should address #1639 by performing at least one API call in the main process. 

Some uploads do not interact with the destination API until the transfer call itself, which must happen in child processes if parallelism is enabled. For users with reauth enabled, this results in a poor experience, since reauth challenges cannot be answered if they're triggered in child processes. This PR adds an API call in the main process to trigger the reauth flow for those users.

### DIFF
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -1698,7 +1698,6 @@ class RsyncCommand(Command):
 
     copy_helper.TriggerReauthForDestinationProviderIfNecessary(
         dst_url,
-        src_url,
         self.gsutil_api,
         parallelism_requested=True,  # rsync uses parallel_operations_override.
     )
--- a/gslib/commands/rsync.py
+++ b/gslib/commands/rsync.py
@@ -1697,7 +1697,10 @@ class RsyncCommand(Command):
       RegisterSignalHandler(signal_num, _HandleSignals)
 
     copy_helper.TriggerReauthForDestinationProviderIfNecessary(
-        self.gsutil_api, dst_url)
+        self.gsutil_api,
+        dst_url,
+        parallelism_requested=True,  # rsync uses parallel_operations_override.
+    )
 
     # Perform sync requests in parallel (-m) mode, if requested, using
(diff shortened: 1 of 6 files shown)


dilipped (commented):
No tests?

dilipped (approved):

gslib/utils/copy_helper.py
```diff
@@ -1486,6 +1486,51 @@ def ExpandUrlToSingleBlr(url_str,
...
+  # Destination wildcards are expanded by an API call in the main process.
+  if ContainsWildcard(destination_url.url_string):
+    return
+
+  # If gsutil executes sequentially, all calls will occur in the main process.
+  if not parallelism_requested:
+    return
+
+  # The specific API call is not important, but one must occur.
+  gsutil_api.GetBucket(
```
tejohnso, line 1528: Note that this breaks systems which have adhered to the principle of least privilege. 

So a system set up with a credential that has **Storage Object Admin** access (`storage.objects.*`) is now going to fail because a bucket listing requires additional privilege (`storage.buckets.*`).

The failure output is `[credential id] does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied on resource (or it may not exist).`

Also it is not intuitive that a copy request would require read permission on an entire bucket. The error leads to confusion.

dilipped: Thanks for bringing this to our attention! We are going to work on a fix soon. I will reopen the original bug

gslib/commands/rsync.py
```diff
@@ -1696,6 +1696,12 @@ def RunCommand(self):
     for signal_num in GetCaughtSignals():
       RegisterSignalHandler(signal_num, _HandleSignals)
 
+    copy_helper.TriggerReauthForDestinationProviderIfNecessary(
+        dst_url,
+        self.gsutil_api,
+        parallelism_requested=True,  # rsync uses parallel_operations_override.
```
NickGoog, line 1702: What if the user hardcoded single process and single thread?

PR #1656, opened 2023-01-10
Adding notes about shim updates for gcloud storage 1.2

thomasmaclean:


gslib/addlhelp/shim.py
```diff
@@ -96,6 +106,12 @@
     are ignored for cloud objects. This is fixed for the shim and gcloud
     storage.
 
+  iam
```
NickGoog, line 109: See the design doc's second item in the caveats section (not posting links externally)

NickGoog (approved):

gslib/addlhelp/shim.py
thomasmaclean: I just added a note about it.

PR #1659, opened 2023-01-12
Fix rpo get shimming command

carlosmerec:
There is a failing test for rpo command:  [test](https://github.com/GoogleCloudPlatform/gsutil/blob/a238e7e3500461dcd8ab15fcb6e7523a2b0f1e51/gslib/tests/test_rpo.py#L238)

The problem is that the [shimming_command_for_rpo_get](https://github.com/GoogleCloudPlatform/gsutil/blob/a238e7e3500461dcd8ab15fcb6e7523a2b0f1e51/gslib/commands/rpo.py#L86) crashes when an s3 bucket is passed due to `sub`  [transform_function](https://cloud.google.com/sdk/gcloud/reference/topic/projections) is expecting a string value, but instead is getting a `None` value, this PR fixes this issue by replacing the `sub` function for the `format` function

### DIFF
--- a/gslib/tests/test_rpo.py
+++ b/gslib/tests/test_rpo.py
@@ -102,8 +102,8 @@ class TestRpoUnit(testcase.GsUtilUnitTestCase):
         info_lines = '\n'.join(mock_log_handler.messages['info'])
         self.assertIn(('Gcloud Storage Command: {} storage'
                        ' buckets list --format=value[separator=": "]'
-                       '(name.sub("^", "gs://"),rpo.yesno(no="None"))').format(
-                           os.path.join('fake_dir', 'bin', 'gcloud')),
+                       '(format("gs://{}", name),rpo.yesno(no="None"))').format(
+                           os.path.join('fake_dir', 'bin', 'gcloud'), r'{}'),
                       info_lines)
 
   def test_shim_translates_recovery_point_objective_set_command(self):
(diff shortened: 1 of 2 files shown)


gslib/tests/test_rpo.py
```diff
@@ -237,7 +237,16 @@ def test_s3_fails_for_set(self):
...
-                            expected_status=1)
-    self.assertIn('command can only be used for GCS buckets', stderr)
+    expected_status = 0 if self._use_gcloud_storage else 1
+    stdout, stderr = self.RunGsUtil(
+        ['rpo', 'get', suri(bucket_uri)],
+        return_stderr=True,
+        return_stdout=True,
+        expected_status=expected_status)
+    if self._use_gcloud_storage:
+      # TODO gcloud storage buckets list command doesn't raise an error when a
```
dilipped, line 247: I would say remove this TODO. The comment is slightly incorrect. gcloud storage shouldn't be raising an error for a list call as it is able to access the metadata of the biucket correctly. We are getting None here because of the format flag. The format flag is trying to access a field that is not present. In that case, returning None is the correct behavior. One solution can be that we populate the field's value as "Not available for S3" for the S3 Resource. You can create an internal bug. for this or add this to the bug (I can add it if you can send me the bug offline).

dilipped (approved):

gslib/tests/test_rpo.py
carlosmerec: Done!

PR #1667, opened 2023-01-26
Small help updates to acl examples and the cp -j flag

starsandskies:


gslib/commands/cp.py
```diff
@@ -484,7 +484,7 @@
                  uploaded objects retain the ``Content-Type`` and name of the
                  original files.
 
-                 Note that if you want to use `the top-level ``-m`` option
+                 Note that if you want to use the ``-m`` `top-level option
```
NickGoog, line 487: Should there be a backtick before "top-level"?

NickGoog (approved):

gslib/commands/cp.py
starsandskies: Yes, while I don't know the exact code logic, this is how to get hyperlinks to appear in the generated docs (see the -v flag below for a similar example). My strong suspicion is that in the original text here, the backticks for -m being embedded in the larger hyperlink backticks were what was messing with the last generation of the docs.

PR #1671, opened 2023-02-13
Fix error message for external account authorized user credentials

aeitzman:
Fix error handling for the new external account authorized user credential type and added tests.

The external account authorized user credential type was added previously in #1617 but the type was missed in hasConfiguredCredentials. This results in the CLI returning misleading error messages when a failure occurs when using the new type.


NickGoog:
Kokoro failures appear unrelated

PR #1672, opened 2023-02-14
Handles OAuthException from google-auth

lsirac:
gsutil does not handle OAuthException's from google-auth, which results in the entire stack trace being printed when an error occurs when using external account credentials. We should prompt the user to sign-in again in this case.

### DIFF
--- a/gslib/__main__.py
+++ b/gslib/__main__.py
@@ -36,7 +36,7 @@ import six
 from six.moves import configparser
 from six.moves import range
 
-from google.auth import exceptions
+from google.auth import exceptions as google_auth_exceptions
 import gslib.exception
 from gslib.exception import CommandException
 from gslib.exception import ControlCException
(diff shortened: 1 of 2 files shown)


gslib/__main__.py
```diff
@@ -760,6 +761,18 @@ def _RunNamedCommandAndHandleExceptions(command_runner,
     else:
       _OutputAndExit('You must re-authenticate with your SAML IdP. '
                      'Please run\n$ gsutil config')
+  except exceptions.OAuthError as e:
```
dilipped, line 764: Any reason for not combining this with line 706. You can do something like


except (oauth2client.client.HttpAccessTokenRefreshError,  exceptions.OAuthError) as e:
  ...

gslib/__main__.py
```diff
@@ -36,6 +36,7 @@
 from six.moves import configparser
 from six.moves import range
 
+from google.auth import exceptions
```
dilipped, line 39: `exceptions` is pretty generic, maybe you can add `as google_auth_exceptions`

PR #1674, opened 2023-02-17
Fix external_account_authorized_user implementation in wrapped_credentials.py + update google-auth dependency

aeitzman:
Changing wrapped_credentials to store the client_secret and refresh_token from external account authorized user credentials. External account credentials did not have these fields so they were never added to wrapped credentials in the previous implementation. This led to a caching bug where the library did not recognize a newer version of the external account authorized user credential since the caching key did not include the refresh token.

### DIFF
--- a/gslib/tests/test_wrapped_credentials.py
+++ b/gslib/tests/test_wrapped_credentials.py
@@ -197,6 +197,7 @@ class TestWrappedCredentials(testcase.GsUtilUnitTestCase):
     creds.access_token = ACCESS_TOKEN
     creds.token_expiry = datetime.datetime(2001, 12, 5, 0, 0)
     creds_json = creds.to_json()
+    self.maxDiff = None
     json_values = json.loads(creds_json)
     expected_json_values = {
         "_class":
@@ -204,7 +205,7 @@ class TestWrappedCredentials(testcase.GsUtilUnitTestCase):
         "_module":
             "gslib.utils.wrapped_credentials",
         "client_id":
-            "//iam.googleapis.com/locations/global/workforcePools/$WORKFORCE_POOL_ID/providers/$PROVIDER_ID",
+            "clientId",
         "access_token":
             ACCESS_TOKEN,
         "token_expiry":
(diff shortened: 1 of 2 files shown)


gslib/tests/test_wrapped_credentials.py
```diff
@@ -197,22 +197,23 @@ def testWrappedCredentialSerializationExternalAccountAuthorizedUser(self):
     creds.access_token = ACCESS_TOKEN
     creds.token_expiry = datetime.datetime(2001, 12, 5, 0, 0)
     creds_json = creds.to_json()
+    self.maxDiff = None
```
dilipped, line 200: Remove this

gslib/utils/wrapped_credentials.py
```diff
@@ -58,18 +58,26 @@ def __init__(self, base_creds):
...
-        raise TypeError("Invalid Credentials")
     self._base = base_creds
-    super(WrappedCredentials, self).__init__(access_token=self._base.token,
-                                             client_id=self._base._audience,
-                                             client_secret=None,
-                                             refresh_token=None,
-                                             token_expiry=self._base.expiry,
-                                             token_uri=None,
-                                             user_agent=None)
+    if isinstance(base_creds, external_account.Credentials):
```
dilipped, line 62: Nit: Maybe you can do
```
    if isinstance(base_creds, external_account.Credentials):
        client_secret = None
        refresh_token = None
    elif ...authotized_user.Credentials:
        client_secret = self._....
         ...
    else:
         raise..
    super(WrappedCredential).__init__(
        ...
        client_secret=client_secret, 
        ....
   )

```

dilipped:
@aeitzman looks like some of the dependencies within google-auth have been updated. Have you already tried installing the package locally via `pip install .` to ensure that there are no dependency conflicts. Unfortunately the kokoro tests don't use `pip` and wouldn't be able to point out any issues with depdency conflicts when trying to install via pip. It's only during the release process that we try running tests for Pip, so I would recommend trying to install the package locally on a fresh virtual env and ensure that there are no dependency conflicts.

aeitzman:
> 

Ran pip install . from fresh vritualenv, no depedency issues.

Issue #1517, opened 2022-05-24
gsutil assumes read/write access to: /usr/lib/google-cloud-sdk/platform/gsutil


ulisesL:
gsutil assumes read/write access to: **/usr/lib/google-cloud-sdk/platform/gsutil**

The issue becomes apparent when the gsutil's mtls configuration is enabled. The temporary client cert will not be able to be created in the path above, which leads to a file not found exception. The file is not found because creation fails due to write restrictions in **/usr/lib/google-cloud-sdk/platform/gsutil**.

PR #1680, opened 2023-03-14
Change state directory for mTLS.

rrauber:
Temporary certificate files used to be in the gsutil install directory. Now they're in the directory that contains the tracker file directory, which is controlled by the `state_dir` boto config option. Addresses #1517.

It looks like a recent PR didn't run the formatter, so there's some noise in this PR. The novel changes are in `context_config.py` and `test_context_config.py`.

### DIFF
--- a/gslib/context_config.py
+++ b/gslib/context_config.py
@@ -27,6 +27,7 @@ from boto import config
 
 import gslib
 from gslib import exception
+from gslib.utils import boto_util
 from gslib.utils import execution_util
 
 # Maintain a single context configuration.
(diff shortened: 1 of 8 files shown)


NickGoog:
>> they're in the tracker file directory
I think mTLS should be in the *parent* of the tracker directory. When I think something is wrong with tracker files, I personally just delete the entire tracker file directory and this operation should have no connection to mTLS.

rrauber:
Updated the description, the state directory actually does contain a separate directory for tracker files: https://github.com/GoogleCloudPlatform/gsutil/blob/c43122ce3515cae034c28c0e2c0f1f2aafece0f7/gslib/tracker_file.py#L89

NickGoog (approved):

PR #1683, opened 2023-03-23
Fix invalid value for property api_endpoint_overrides/storage

carlosmerec:
When the `shim` is used the following error is thrown by `gcloud`:

```
Invalid value for property [api_endpoint_overrides/storage]: The endpoint_overrides property must be an absolute URI beginning with http:// or https:// and ending with a trailing '/'. [https://storage.googleapis.com/storage/v1] is not a valid endpoint override.
```

### DIFF
--- a/gslib/tests/test_shim_util.py
+++ b/gslib/tests/test_shim_util.py
@@ -1082,7 +1082,7 @@ class TestBotoTranslation(testcase.GsUtilUnitTestCase):
       self.assertEqual(
           env_vars, {
               'CLOUDSDK_API_ENDPOINT_OVERRIDES_STORAGE':
-                  'https://foo_host:1234/storage/v2',
+                  'https://foo_host:1234/storage/v2/',
           })
 
   def test_gcs_json_endpoint_translation_with_missing_port(self):
@@ -1094,10 +1094,11 @@ class TestBotoTranslation(testcase.GsUtilUnitTestCase):
     }):
       flags, env_vars = self._fake_command._translate_boto_config()
       self.assertEqual(flags, [])
-      self.assertEqual(env_vars, {
-          'CLOUDSDK_API_ENDPOINT_OVERRIDES_STORAGE':
-              'https://foo_host/storage/v2',
-      })
+      self.assertEqual(
+          env_vars, {
+              'CLOUDSDK_API_ENDPOINT_OVERRIDES_STORAGE':
+                  'https://foo_host/storage/v2/',
+          })
 
   def test_gcs_json_endpoint_translation_usees_default_version_v1(self):
     with _mock_boto_config(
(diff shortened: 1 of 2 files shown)


gslib/tests/test_shim_util.py
```diff
@@ -1110,9 +1111,24 @@ def test_gcs_json_endpoint_translation_usees_default_version_v1(self):
       self.assertEqual(
           env_vars, {
               'CLOUDSDK_API_ENDPOINT_OVERRIDES_STORAGE':
-                  'https://foo_host:1234/storage/v1'
+                  'https://foo_host:1234/storage/v1/'
           })
 
+  def test_gcs_json_endpoint_translation_should_start_with_https(self):
```
dilipped, line 1117: Isn't the test above taking care of ensuring `https` is present? Do we really need this additional test?

gslib/tests/test_shim_util.py
```diff
@@ -1110,9 +1111,24 @@ def test_gcs_json_endpoint_translation_usees_default_version_v1(self):
...
+    with _mock_boto_config({'Credentials': {'gs_json_host': 'foo_host',}}):
+      flags, env_vars = self._fake_command._translate_boto_config()
+      self.assertEqual(flags, [])
+      self.assertTrue(
+          env_vars['CLOUDSDK_API_ENDPOINT_OVERRIDES_STORAGE'].startswith(
+              'https://'))
+
+  def test_gcs_json_endpoint_translation_should_end_with_trailing_slash(self):
```
dilipped, line 1125: Ditto. Feels like test_gcs_json_endpoint_translation is already taking care of ensuring the endpoint ends with  a slash?

gslib/tests/test_shim_util.py
carlosmerec: Yeah, implicitly is testing that, I should probably delete this and add a comment?, I wanted to be clear that the `https` string must be present at the beginning of the string when the translation is made.

gslib/tests/test_shim_util.py
carlosmerec: It is the same reason as above, be explicit about the requirement that the string should end with a trailing `/`, but probably a comment is enough?

gslib/tests/test_shim_util.py
dilipped: I see. Yeah, a comment should be enough. We can skip the extra test.

gslib/tests/test_shim_util.py
dilipped: Same here, comment should be enough.

carlosmerec:
For the Kokoro tests, the linux ones fail due to `BUILD FAILED` and the mac os one is because of a test failure but it seems it is unrelated.

PR #1684, opened 2023-03-23
Update help doc references to gsutil config

starsandskies:
Notably, this doesn't update the core `gsutil config` reference doc itself. My understanding is that update is being taken care of separately, but I'd be happy to fold it into this PR!


gslib/addlhelp/creds.py
```diff
@@ -40,23 +40,24 @@
...
-  are located at ~/.boto unless a different path is specified in the BOTO_CONFIG
-  environment variable). However, gsutil will still look for credentials in the
-  boto config file if a type of non-GCS credential is needed that's not stored
-  in the gcloud credential store (e.g., an HMAC credential for an S3 account).
+  Once credentials have been configured via ``gcloud auth``, those credentials
+  are used regardless of whether the user has any boto configuration files
+  (which are located at ~/.boto unless a different path is specified in the
+  BOTO_CONFIG environment variable). However, gsutil still looks for credentials
+  in the boto config file if a type of non-GCS credential is needed that's not
+  stored in the gcloud credential store (e.g., an HMAC credential for an S3
```
NickGoog, line 48: This might be impacted by the fact that we're adding HMAC support to gcloud. Not sure if we plan on adding it to the wrapper script and passing HMAC from gcloud to gsutil @thomasmaclean

gslib/addlhelp/dev.py
```diff
@@ -96,20 +96,19 @@
        ./gsutil test -u
 
      If you made changes to boto, please run the boto tests. For these tests you
-     need to use HMAC credentials (from gsutil config -a), because the current
-     boto test suite doesn't import the OAuth2 handler. You'll also need to
-     install some python modules. Change directories into the boto root
-     directory at third_party/boto and run:
+     need to use HMAC credentials (from gsutil config -a), because the boto test
+     suite doesn't import the OAuth2 handler. You'll also need to install some
+     python modules. Change directories into the boto root directory at
```
NickGoog, line 101: Typo: `Python`

starsandskies: Done, though I assumed the crux of the comment was capitalization and not code font. Let me know if I'm mistaken

gslib/addlhelp/creds.py
starsandskies: If it is impacted, would it be possible to update the wording here separately? I only made superficial changes to this paragraph - it will continue to say basically the same thing it has been saying up to now.

NickGoog: @thomasmaclean should probably approve b/c he did the HMAC design

thomasmaclean: Although you will be able to specify the HMAC keys through properties in gcloud, the shim will have the ability pull in HMAC keys from the boto file, so this logic still stands.

Syam1388 (changes requested):
Hi

PR #1682, opened 2023-03-21
Post OOB OAuth2 deprecation changes to config help text

thomasmaclean:

PR #1685, opened 2023-03-27
Minor text changes to OAuth2 OOB changes

thomasmaclean:
Response to late-breaking suggestions from @starsandskies. See #1682.


starsandskies (approved):
LGTM!

PR #1691, opened 2023-04-24
Update broken Kokoro E2E tests.

NickGoog:
The underlying VMs running our tests changed.

* Added a "safe directory" for git. Git now blocks submodules in locations not owned by the current user. More info: https://stackoverflow.com/questions/72978485/git-submodule-update-failed-with-fatal-detected-dubious-ownership-in-repositor
* Removed pyenv update command. This isn't a pyenv command unless you install the pyenv-update package.
* Removed pyenv version fix that's no longer needed according to the bug that had been linked.

WIP: This does not yet fix the SSL issue for Python 3.5


test/ci/kokoro/run_integ_tests.sh
```diff
@@ -68,10 +68,6 @@ function install_pyenv {
     export PATH="$PYENV_ROOT/bin:$PATH"
     eval "$(pyenv init --path)"
   fi
-  pyenv update
```
dilipped, line 71: Maybe leave a comment saying that if we start seeing kokoro failing because of a certain python version missing, we can do `pyenv update`.  That way we don't have to spend time investigating why the version is not found.

IIRC, the update was added at that time because python 3.8 was missing. Looks like the new pool probably ships with pyenv that is aware of 3.8, but it might happen that certain future Python versions are not available with the available pyenv.

dilipped (approved):

test/ci/kokoro/run_integ_tests.sh
NickGoog: `pyenv update` just isn't a command unless we also want to install the `pyenv-update` package

https://github.com/pyenv/pyenv#upgrading-with-installer-or-git-checkout

https://github.com/pyenv/pyenv#upgrading-with-installer-or-git-checkout

PR #1689, opened 2023-04-18
Update google auth library

BigTailWolf:
Update to latest release of Auth library that can support TPC features.

Our older version of Auth SDK is validating the external credentials' iam endpoints, sts endpoints pattern at constructing, this will for sure to break our TPC(two party control) features.

Basically the TPC is some customer deploying a clause that using full Google stack but have their own management, such like the different domain name.

All these changes has already been pulled into `gcloud` changes.

PR #1692, opened 2023-04-24
Fix SSL missing error by updating Linux Python 3.5 version.

NickGoog:
See https://github.com/pyenv/pyenv/wiki/Common-build-problems#2-your-openssl-version-is-incompatible-with-the-python-version-youre-trying-to-install

### DIFF
--- a/test/ci/kokoro/run_integ_tests.sh
+++ b/test/ci/kokoro/run_integ_tests.sh
@@ -36,9 +36,14 @@ BOTO_CONFIG="/tmpfs/src/.boto_$API"
 # https://cloud.google.com/storage/docs/boto-gsutil
 export BOTO_PATH="$BOTO_CONFIG"
 
-# Set the locale to utf-8 for macos b/154863917
-if [[ $KOKORO_JOB_NAME =~ "macos" ]]; then
+# Set the locale to utf-8 for macos b/154863917 and linux
+# https://github.com/GoogleCloudPlatform/gsutil/pull/1692
+if [[ $KOKORO_JOB_NAME =~ "linux" ]]; then
+  export LANG=C.UTF-8
+  export LC_ALL=C.UTF-8
+elif [[ $KOKORO_JOB_NAME =~ "macos" ]]; then
   export LANG=en_US.UTF-8
+  export LC_ALL=en_US.UTF-8
 fi
 
 function preferred_python_release {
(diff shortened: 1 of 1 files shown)


NickGoog:
The Python 3.5 tests are running, but seem to have more errors than the other Python version tests (`WeakValueDictionary`). Since 3.5 is being deprecated soon'ish. The failure is likely this Python bug in 3.5.3: https://bugs.python.org/issue29519

Switching to 3.5.4 to see if it helps as suggested here (didn't actually see evidence in the link in the post but will test): https://github.com/ansible/ansible/issues/21982#issuecomment-424303758

NickGoog:
Python 3.5.4 solved the `WeakValueDictionary` issue

dilipped:
All linux tests seem to be failing. Especially something like 
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb1 in position 21118: invalid start byte
```
Has this started happening after the pool change? Can you try doing https://github.com/GoogleCloudPlatform/gsutil/blob/09257c0e83df44641d56ba3c3b1fc806d541a413/test/ci/kokoro/run_integ_tests.sh#L41 for linux as well?

NickGoog:
```
if [[ $KOKORO_JOB_NAME =~ "macos" || $KOKORO_JOB_NAME =~ "linux" ]]; then
  export LANG=en_US.UTF-8
fi
```
Didn't seem to work (not posting logs externally).

I also tried:
```
if [[ $KOKORO_JOB_NAME =~ "macos" || $KOKORO_JOB_NAME =~ "linux" ]]; then
  export LANG=en_US.UTF-8
  export PYTHONIOENCODING=utf-8
fi
```

The ASCII error appears on Linux on Python 3.5 and 3.6 (not 3.7 and 3.8).

Playing around with some other env vars

NickGoog:
@dilipped , I set different env vars for Linux, and it's looking good!

dilipped:
> @dilipped , I set different env vars for Linux, and it's looking good!

Nice! Good job fixing the utf-8 issue. Looks like all tests are still failing for unrelated changes. Do we have a fix for it or can this be assigned to the current on-duty?

dilipped (approved):

NickGoog:
@dilipped , yeah, the failures related to POSIX and unwritable files have been happening so consistently I don't think they're flakes. Did not investigate a fix, but I'm going to do some light research into all the current bugs for https://github.com/GoogleCloudPlatform/gsutil/pull/1689

NickGoog:
The Windows Kokoro VM is now broken, seemingly as a result of an internal CL changing the VM platform. Handing off to the next OnDuty @rrauber

PR #1696, opened 2023-05-03
Add shim support for boto rsync_buffer_lines

NickGoog:
My yapf formatting changes appear to have picked up other files.

I edited:
* `gslib/tests/test_shim_util.py`
* `gslib/utils/shim_util.py`


dilipped:
Looks good. Some rsync and cp tests are failing in kokoro, are those related? 

also, was the parity check script executed for all these flags? Did you confirm that they work as intended?

NickGoog:
No parity done yet. The PR will get huge if I start that.

dilipped:
Can we add one flag at a time in that case. It's risky to add all those flags to the shim without verifying that they work as intended. To keep the PR small, maybe we can add one flag, update the tests if needed, check it with paritycheck script, push, and repeat this for other flags. This might mean a lot more PRs, but we will avoid submitted something without testing.

NickGoog:
Isn't this all in alpha? Parity probably won't be ready for weeks

NickGoog:
Isn't this all in alpha? Parity probably won't be ready for weeks

NickGoog:
This might take a while

gslib/utils/shim_util.py
```diff
@@ -401,7 +403,10 @@ def _translate_top_level_flags(self):
       top_level_flags.append('--impersonate-service-account=' +
                              constants.IMPERSONATE_SERVICE_ACCOUNT)
     # TODO(b/208294509) Add --perf-trace-token translation.
-    if not self.parallel_operations:
+    rsync_parallelism_disabled = self.command_name != 'rsync' or (
```
rrauber, line 406: With rsync in the variable name it's a bit counterintuitive for this to be true when the command is not rsync. How about something like this?
 
```
should_use_rsync_override = self.command_name == 'rsync' and not (process and thread count check)
if not self.parallel_operations and not should_use_rsync_override:
  ...
```

gslib/tests/test_rsync.py
```diff
@@ -257,13 +293,31 @@ def _Check1():
...
            suri(bucket2_uri)], return_stderr=True)
-      self.assertIn('obj1 has an invalid mtime in its metadata', stderr)
-      self.assertNotIn('obj2 has an invalid mtime in its metadata', stderr)
-      self.assertIn(
-          'obj3 has an mtime more than 1 day from current system '
-          'time', stderr)
-      self.assertIn('obj4 has a negative mtime in its metadata', stderr)
-      self.assertIn('obj5 has a negative mtime in its metadata', stderr)
+      if self._use_gcloud_storage:
+        self.assertTrue(
```
rrauber, line 297: Does self.assertRegex work here?

gslib/tests/test_rsync.py
```diff
@@ -2463,12 +2590,12 @@ def _Check1():
     @Retry(AssertionError, tries=3, timeout_secs=1)
     def _Check2():
       # Check that re-running the same rsync command causes no more changes.
-      self.assertEquals(
-          NO_CHANGES,
-          self.RunGsUtil(['rsync', '-d',
-                          suri(bucket1_uri),
-                          suri(bucket2_uri)],
-                         return_stderr=True))
+      # Check that re-running the same rsync command causes no more changes.
```
rrauber, line 2593: Looks like this comment is duplicated.

gslib/commands/cp.py
```diff
@@ -639,11 +640,35 @@
...
+
+  Args:
+    sub_opts: List of pairs representing flag keys and values, e.g.
+      [('a', 'public-read')]
+  """
+  predefined_acl_idx = None
+  for i, (k, _) in enumerate(sub_opts):
+    if k == '-a':
+      predefined_acl_idx = i
+      break
```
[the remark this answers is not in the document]
NickGoog: Not mutating `sub_opts` might be infeasible b/c `self.sub_opts` is attached to the `Command` object and appears at different points in execution. In other words, I don't think there's a clean place where we can pass a long a clean copy of `sub_opts` with the translations made.

This transform function is specific to `cp`, `mv`, and `rsync` (not `acl` or `defacl`, which use the same translation map but with different syntax). Moving it to a more generic location didn't make sense to me for that reason.

gslib/tests/test_rsync.py
NickGoog: Good catch

gslib/utils/shim_util.py
NickGoog: Good idea

gslib/tests/test_rsync.py
NickGoog: Done

gslib/tests/test_rsync.py
```diff
@@ -465,18 +532,18 @@ def test_bucket_to_bucket_dst_mtime(self):
     # matches dst.
     src_bucket = self.CreateBucket()
     dst_bucket = self.CreateBucket()
-    self.CreateObject(bucket_uri=src_bucket,
-                      object_name='obj1',
-                      contents=b'OBJ1')
-    self.CreateObject(bucket_uri=src_bucket,
-                      object_name='subdir/obj2',
-                      contents=b'subdir/obj2')
+    source_obj1 = self.CreateObject(bucket_uri=src_bucket,
```
rrauber, line 535: Are these `source_obj` variables used?

gslib/tests/test_rsync.py
```diff
@@ -201,16 +232,21 @@ def test_invalid_args(self):
     self.RunGsUtil(['rsync', suri(bucket_uri), suri(obj1)], expected_status=1)
     # rsync bucket to non-existent bucket.
     self.RunGsUtil(
-        ['rsync', suri(bucket_uri), self.nonexistent_bucket_name],
+        ['rsync',
+         suri(bucket_uri), 'gs://' + self.nonexistent_bucket_name],
```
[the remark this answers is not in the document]
rrauber: Think it'd be worth adding a bucket to directory case so we don't lose coverage?

gslib/tests/test_rsync.py
```diff
@@ -1837,25 +1928,52 @@ def _Check1():
...
+        self.assertRegex(stderr, uid_regex.format('obj13'))
+        self.assertRegex(stderr, gid_regex.format('obj14'))
+        self.assertRegex(stderr, gid_regex.format('obj15'))
+        self.assertRegex(stderr, read_regex.format('obj16'))
+        self.assertRegex(stderr, read_regex.format('obj17'))
+        self.assertRegex(stderr, read_regex.format('obj18'))
+        self.assertRegex(stderr, read_regex.format('obj19'))
+      else:
+        self.assertIn(ORPHANED_FILE, stderr)
+        self.assertTrue(BuildErrorRegex(obj1, POSIX_MODE_ERROR).search(stderr))
```
rrauber, line 1959: These could also probably be self.assertRegex (though I see that this was gsutil code)

gslib/tests/test_rsync.py
```diff
@@ -3291,11 +3421,13 @@ def test_bucket_to_bucket_includes_arbitrary_headers(self):
         suri(bucket2_uri)
     ],
                             return_stderr=True)
-
-    headers_for_all_requests = re.findall(r"Headers: \{([\s\S]*?)\}", stderr)
-    self.assertTrue(headers_for_all_requests)
-    for headers in headers_for_all_requests:
-      self.assertIn("'arbitrary': 'header'", headers)
+    if self._use_gcloud_storage:
+      self.assertIn('--additional-headers=arbitrary=header', stderr)
```
rrauber, line 3425: I think this test was intended to ensure that all requests contained the additional header, so this assertion is a bit weaker than the gsutil one. Is it possible to do a similar check here?

rrauber (approved):

gslib/tests/test_rsync.py
NickGoog: Good catch

gslib/tests/test_rsync.py
NickGoog: Done

gslib/tests/test_rsync.py
NickGoog: Good call. Done.

PR #1699, opened 2023-05-11
Du shim support and test parity.

rrauber:
All tests pass as they are with the current gcloud implementation.


dilipped (commented):
Was the parity script ran for all these flags?

dilipped (commented):
Was the parity script ran for all these flags?

rrauber:
It was! As I mentioned in the description no test changes were necessary.

PR #1703, opened 2023-05-16
Respect PAP and UBLA settings in S3 tests.

rrauber:
S3 equivalents of PAP and UBLA got enabled for our test project, which caused some tests to fail. This PR introduces some code to respect settings for PAP and UBLA passed to the CreateBucket helper method for S3, which parallels the approach we took for the GCS APIs.

Since these features aren't supported by gsutil, I had to use some low(-er) level boto methods to disable them.


NickGoog:
Can you log into the S3 project and disable these new default settings?

PR #1708, opened 2023-06-01
Adding support for HMAC auth to the shim.

thomasmaclean:
This will apply HMAC credentials to gcloud storage when appropriate. When supplied to gcloud storage, the HMAC credentials take precedence, so the shim needs to include gsutil's prioritization logic when applying those credentials. This PR encapsulates that prioritization logic in `boto_util.UsingGsHmac()` and applies it in both `cloud_api_delegator` and `shim_util`.

### DIFF
--- a/gslib/tests/test_shim_util.py
+++ b/gslib/tests/test_shim_util.py
@@ -1156,26 +1156,18 @@ class TestBotoTranslation(testcase.GsUtilUnitTestCase):
       flags, _ = self._fake_command._translate_boto_config()
       self.assertEqual(flags, ['--decryption-keys=key1,key12,key100'])
 
-  @mock.patch.object(boto_util, 'UsingGsHmac')
-  def test_gs_hmac_auth_env_set_correctly(self, mock_using_gs_hmac):
+  @mock.patch.object(boto_util, 'UsingGsHmac', return_value=False)
+  def test_gs_hmac_auth_env_when_not_using_gs_hmac(self, mock_using_gs_hmac):
     with _mock_boto_config({
         'Credentials': {
-            'gs_access_key_id': 'unmapped',
-            'gs_secret_access_key': 'mapped',
+            'gs_access_key_id': 'foo',
+            'gs_secret_access_key': 'bar',
         }
     }):
-      # Another use case that would have been good for parameterized.
-      # UsingGsHmac ultimately controls whether the gs HMAC creds
-      # are mapped, so the first key, gs_access_key_id, will not be
-      # mapped, while the second, gs_secret_access_key will.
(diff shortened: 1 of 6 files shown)


gslib/tests/test_shim_util.py
```diff
@@ -1155,12 +1156,41 @@ def test_decryption_key_gets_converted_to_flag(self):
...
-            'use_client_certificate': True,
+            'gs_access_key_id': 'unmapped',
+            'gs_secret_access_key': 'mapped',
+        }
+    }):
+      # Another use case that would have been good for parameterized.
+      # UsingGsHmac ultimately controls whether the gs HMAC creds
+      # are mapped, so the first key, gs_access_key_id, will not be
+      # mapped, while the second, gs_secret_access_key will.
+      mock_using_gs_hmac.side_effect = [False, True]
```
rrauber, line 1171: I think this would be better as two separate tests, or maybe even just one -- it looks like the True case is covered below, so we could probably just check the False case here.

rrauber (approved):

gslib/tests/test_shim_util.py
thomasmaclean: Just made this a single test for the negative case.

gslib/utils/boto_util.py
```diff
@@ -650,3 +650,13 @@ def HasUserSpecifiedGsHost():
...
+
+def UsingGsHmac():
+  # NOTE: External credentials are omitted intentionally as HMAC takes priority
+  # over the external credentials types.
+  config = boto.config
+  return (not config.has_option('Credentials', 'gs_oauth2_refresh_token') and
+        not (config.has_option('Credentials', 'gs_service_client_id') and
+             config.has_option('Credentials', 'gs_service_key_file')) and
+        (config.has_option('Credentials', 'gs_access_key_id') and
+         config.has_option('Credentials', 'gs_secret_access_key')))
```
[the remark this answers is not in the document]
thomasmaclean: Done.

rrauber (approved):

gslib/utils/shim_util.py
```diff
@@ -578,6 +592,12 @@ def translate_to_gcloud_storage_if_requested(self):
               ' You can make gsutil use the same credentials by running:\n'
               '{} config set pass_credentials_to_gsutil True'.format(
                   gcloud_binary_path))
+        elif (boto_util.UsingGsHmac and
```
rrauber, line 595: I think this should be boto_util.UsingGsHmac()

thomasmaclean: Done.

PR #1714, opened 2023-06-09
Add shim support for signurl.

rrauber:


gslib/commands/signurl.py
```diff
@@ -403,6 +404,39 @@ class UrlSignCommand(Command):
       subcommand_help_text={},
   )
 
+  _COMMAND_MAP = GcloudStorageMap(
+      gcloud_command=[
+          'alpha', 'storage', 'sign-url',
+          '--format=csv[separator="\\t"](resource, http_verb, expiration, signed_url)'
```
NickGoog, line 410: What's this going to look like versus gsutil?

gslib/commands/signurl.py
```diff
@@ -403,6 +404,39 @@ class UrlSignCommand(Command):
...
+  def get_gcloud_storage_args(self):
+    self._COMMAND_MAP.gcloud_command += ['--private-key-file=' + self.args[0]]
+    self.args = self.args[1:]
+
+    retention_arg_idx = 0
+    while retention_arg_idx < len(self.sub_opts):
+      if self.sub_opts[retention_arg_idx][0] == '-d':
+        break
+      retention_arg_idx += 1
+    if retention_arg_idx < len(self.sub_opts):
```
NickGoog, line 431: I think it's a bit less error prone if you initially set the idx variable to `None`, set it on index match, and proceed with the conversion if not `None`.

NickGoog:
No more E2E tests needed modification?

gslib/commands/signurl.py
rrauber: Done.

gslib/commands/signurl.py
rrauber: The data itself is exactly the same. Here's gsutil's format: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/commands/signurl.py#L620

The only difference is the header row. Here's what gsutil puts there: https://github.com/GoogleCloudPlatform/gsutil/blob/master/gslib/commands/signurl.py#L558 

gcloud storage has tab separated `resource`, `http_verb`, `expiration`, `signed_url`. I didn't see a good way to print a row like this only once with the gcloud format flag, so to introduce a header row like gsutil's I think we'd need to do it with logic in the gcloud implementation. Think I should do that?

rrauber:
No other E2E tests needed modification.

gslib/commands/signurl.py
NickGoog: Might be able to use a `multi` format where the first value is a `value` format with the desired header string and the second format is the CSV logic above

NickGoog (approved):

PR #1717, opened 2023-06-14
Allow setting schema in gs_json_host

ramir-savvy:
Allowing the user to specify the schema when providing the `gc_json_host`


ramir-savvy:
Found that this is enough for ls command, but not for upload and download files. Closing for now

PR #1727, opened 2023-07-11
Shim iam ch command

rrauber:


gslib/commands/iam.py
```diff
@@ -363,8 +373,106 @@ def get_gcloud_storage_args(self):
...
+
+        urls = ls_process.stdout.strip('\n').split('\n')
+      else:
+        urls = [str(url_pattern)]
+
+      # With the -r flag, container headings are printed ending in '/:'.
+      # All headings other than the first are preceded by a new line.
+      # These need to be skipped, without skipping objects that end in
+      # '/:'.
+      next_url_could_be_container_line = True
```
NickGoog, line 439: This would be more robust if it was based on reading object attributes from `ls --json` or `objects list` instead of determining data types from output formatting.

gslib/commands/iam.py
```diff
@@ -528,7 +636,14 @@ def _PatchIamHelperInternal(self,
     orig_bindings = list(bindings)
 
     for (is_grant, diff) in bindings_tuples:
-      bindings = PatchBindings(bindings, BindingsTuple(is_grant, diff))
```
NickGoog, line 531: Did you change the inputs?

gslib/commands/iam.py
```diff
@@ -578,16 +693,22 @@ def _PatchIam(self):
 
     # All following arguments are urls.
     for token in it:
-      patterns.append(token)
+      url_pattern_strings.append(token)
+
+    url_objects = list(map(StorageUrlFromString, url_pattern_strings))
+    _RaiseErrorIfUrlsAreMixOfBucketsAndObjects(url_objects,
```
NickGoog, line 699: Shouldn't this be higher level?

gslib/utils/shim_util.py
```diff
@@ -575,17 +587,8 @@ def translate_to_gcloud_storage_if_requested(self):
...
-        env_variables.update(env_vars_from_boto)
-
-        gcloud_binary_path = _get_gcloud_binary_path()
-        gcloud_storage_command = ([gcloud_binary_path] +
-                                  self.get_gcloud_storage_args() +
-                                  top_level_flags + header_flags +
-                                  flags_from_boto)
+        env_variables, gcloud_storage_command = self._get_full_gcloud_storage_execution_information(
+            self.get_gcloud_storage_args())
         if hidden_shim_mode == HIDDEN_SHIM_MODE.DRY_RUN:
```
NickGoog, line 592: Will this work with dry run mode?

NickGoog (commented):
Will review tests and iam_helper after

gslib/utils/shim_util.py
rrauber: It won't. Added a warning.

gslib/commands/iam.py
rrauber: You mean higher in the file? I moved it.

gslib/commands/iam.py
rrauber: I did. You'll see this when you get to iam_helper, but previously PatchBindings took a bindings message, passed it to BindingsToDict which transformed it to a `role: set(members)` dict which is convenient for updates. Then, PatchBindings would perform the updates, and convert the bindings back to a bindings message.

The shim doesn't have much use for the bindings messages, since we pass and receive strings directly from subcommands, so changing PatchBindings to use `role: set(members)` dicts directly cut down on a lot of unnecessary conversions.

gslib/commands/iam.py
rrauber: Agreed! `objects list` doesn't quite work since it only lists the contents of directories with a wildcard, so it functions differently than recursive listing:

```
$ gcloud storage objects list gs://bucket/dir
$ gcloud storage objects list gs://bucket/dir/*
...
name: dir/file
...
```

ls --json seems to work though!

gslib/commands/iam.py
```diff
@@ -363,8 +373,106 @@ def get_gcloud_storage_args(self):
...
+    if url.IsFileUrl():
+      error_msg = 'Invalid Cloud URL "%s".' % url.object_name
+      if set(url.object_name).issubset(set('-Rrf')):
+        error_msg += (
+            ' This resource handle looks like a flag, which must appear'
+            ' before all bindings. See "gsutil help iam ch" for more details.')
+      raise CommandException(error_msg)
+
+  def _run_gcloud_subprocess(self, args, stdin=None):
+    subprocess_envs = os.environ.copy()
```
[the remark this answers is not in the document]
rrauber: Swapped the name.

gslib/commands/iam.py
```diff
@@ -363,8 +373,106 @@ def get_gcloud_storage_args(self):
...
+  def _RaiseIfInvalidUrl(self, url):
+    if url.IsFileUrl():
+      error_msg = 'Invalid Cloud URL "%s".' % url.object_name
+      if set(url.object_name).issubset(set('-Rrf')):
+        error_msg += (
+            ' This resource handle looks like a flag, which must appear'
+            ' before all bindings. See "gsutil help iam ch" for more details.')
+      raise CommandException(error_msg)
+
+  def _run_gcloud_subprocess(self, args, stdin=None):
```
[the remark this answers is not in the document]
rrauber: Here we go!

gslib/commands/iam.py
```diff
@@ -363,8 +373,106 @@ def get_gcloud_storage_args(self):
...
+      return []  # Translation occurs in self.run_gcloud_storage.
+
     return super().get_gcloud_storage_args(gcloud_storage_map)
 
+  def _RaiseIfInvalidUrl(self, url):
+    if url.IsFileUrl():
+      error_msg = 'Invalid Cloud URL "%s".' % url.object_name
+      if set(url.object_name).issubset(set('-Rrf')):
+        error_msg += (
+            ' This resource handle looks like a flag, which must appear'
```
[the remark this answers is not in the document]
rrauber: For this command that restriction is relaxed a bit. The reason is that added bindings are not introduced by a flag, and users are allowed to mix them with deleted bindings, which are introduced by a flag. In other words, something like this is possible, where a flag follows something that looks like a positional argument:

`iam ch allAuthenticatedUsers:objectCreator -d allUsers:objectCreator gs://bucket`

gslib/commands/iam.py
```diff
@@ -316,9 +320,15 @@ class IamCommand(Command):
...
+  def _get_resource_type(self, url_patterns):
+    if self.recursion_requested or url_patterns[0].IsObject():
+      return 'objects'
+    else:
+      return 'buckets'
+
   def get_gcloud_storage_args(self):
-    sub_command = self.args.pop(0)
-    if sub_command == 'get':
+    self.sub_command = self.args.pop(0)
```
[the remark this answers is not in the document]
rrauber: It's used in `run_gcloud_storage`; we defer to the parent class's implementation of `run_cloud_storage` if self.sub_command is not `ch`.

gslib/commands/iam.py
NickGoog: No, like, earlier in the logic before we delve into `ch`

rrauber: Ah, I see. In this case we can't really do that, since there's ch-specific logic to figure out what the URLs even are. This is related to the problem that I mentioned here: https://github.com/GoogleCloudPlatform/gsutil/pull/1727#discussion_r1261418660

Bindings that we add look like positional arguments/URLs, so we need to separate them out before we get the URLs we can pass to this method.

gslib/commands/iam.py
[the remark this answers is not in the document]
NickGoog: I meant the logic

gslib/commands/iam.py
NickGoog: Weren't we planning on deprecating Apitools completely? We might not have time to do that, but would it be simpler to make dict-only tools instead of mixing this logic?

gslib/commands/iam.py
```diff
@@ -363,8 +374,153 @@ def get_gcloud_storage_args(self):
...
+            'The shim maps iam ch commands to several gcloud storage commands,'
+            ' which cannot be determined without running gcloud storage.')
+      return []  # Translation occurs in self.run_gcloud_storage.
+
     return super().get_gcloud_storage_args(gcloud_storage_map)
 
+  def _RaiseIfInvalidUrl(self, url):
+    if url.IsFileUrl():
+      error_msg = 'Invalid Cloud URL "%s".' % url.object_name
+      if set(url.object_name).issubset(set('-Rrf')):
```
NickGoog, line 391: Can you just do `url.object_name in ('-R', '-r', '-f')`? What about `-d`?

gslib/commands/iam.py
```diff
@@ -363,8 +374,153 @@ def get_gcloud_storage_args(self):
...
+    # parse self.args for a mix of both bindings and CloudUrls. We are not
+    # expecting to come across the -r, -f flags here.
+    it = iter(self.args)
+    for token in it:
+      if (STORAGE_URI_REGEX.match(token) and
+          IsKnownUrlScheme(GetSchemeFromUrlString(token))):
+        url_pattern_strings.append(token)
+        break
+      if token == '-d':
+        patch_bindings_tuples.append(BindingStringToTuple(False, next(it)))
```
NickGoog, line 426: The empty iterator error message might be a bit confusing here

gslib/commands/iam.py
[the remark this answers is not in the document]
rrauber: Really only two lines are, the ones managing environment variables. The subprocess.run call here has to look quite different, and instead of returning just the return code we need to return the entire process. I could move the lines setting up the subprocess envs to a helper function if you think that'd be better.

gslib/commands/iam.py
rrauber: I guess I didn't quite complete the thought in the earlier comment. The changes to PatchBindings do move it towards only using dicts. It no longer takes apitools messages.

gslib/commands/iam.py
rrauber: This is the existing gsutil check, and I'd prefer to keep it the same for consistency/compatibility as it concerns `-d`. Additionally, `url.object_name in ('-R', '-r', '-f')` does have different behavior than what's here, since the existing check handles combined flags like `-rf`.

gslib/commands/iam.py
rrauber: Fixed.

gslib/commands/iam.py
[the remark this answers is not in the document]
NickGoog: A DRY helper function in `shim_util.py` would be good

gslib/utils/iam_helper.py
```diff
@@ -95,18 +95,22 @@ def DeserializeBindingsTuple(serialized_bindings_tuple):
 
 
 def BindingsToDict(bindings):
-  """Converts a list of BindingsValueListEntry to a dictionary.
+  """Reformats policy bindings metadata.
 
   Args:
-    bindings: A list of BindingsValueListEntry instances.
+    bindings: Either a list of BindingsValueListEntry instances,
+    or a list of analogous dictionaries.
```
NickGoog, line 102: Can you add the expected structure of the dictionaries? Mentioned in another comment that maybe there should be separate functions for dicts and Apitools objects like `BindingsValueListEntry`

gslib/utils/iam_helper.py
```diff
@@ -151,44 +155,35 @@ def DiffBindings(old, new):
...
       # Drop all members with the DROP_ALL role specifed from input.
-      tmp_base[role].difference_update(tmp_diff[DROP_ALL])
-
-  # Construct the BindingsValueListEntry list
-  bindings = [
-      apitools_messages.Policy.BindingsValueListEntry(role=r, members=list(m))
-      for (r, m) in six.iteritems(tmp_base)
-      if m
-  ]
+      base[role].difference_update(diff[DROP_ALL])
```
NickGoog, line 184: What happens here? `DROP_ALL = ''`. When would empty string be in `base[role]`?

gslib/tests/test_iam.py
```diff
@@ -226,106 +227,110 @@ def test_diff_bindings_swap_member(self):
...
   def test_patch_bindings_grant(self):
     """Tests patching a grant binding."""
-    base = [
+    base_list = [
         bvle(role='a', members=['user:foo@bar.com']),
         bvle(role='b', members=['user:foo@bar.com']),
         bvle(role='c', members=['user:foo@bar.com']),
     ]
-    diff = [
+    base = BindingsToDict(base_list)
```
NickGoog, line 235: Would it make more sense to initially declare these variables as dicts instead of making lists and then converting them to dicts?

gslib/tests/test_iam.py
```diff
@@ -1555,3 +1580,525 @@ def test_shim_translates_iam_set_handles_empty_etag(self):
...
+      }):
+        self.RunCommand('iam', [
+            'ch', 'allAuthenticatedUsers:modified-role', '-d',
+            'user:deleted-user@example.com', '-d', 'allUsers:deleted-role',
+            'gs://b'
+        ])
+
+      self.assertEqual(mock_run.call_args_list, [
+          self._get_run_call([
+              os.path.join('fake_dir', 'bin', 'gcloud'), 'alpha', 'storage',
```
NickGoog, line 1656: `os.path.join('fake_dir', 'bin', 'gcloud')` -> `shim_util._get_gcloud_binary_path('fake_dir')`
after
https://github.com/GoogleCloudPlatform/gsutil/pull/1728

gslib/tests/test_iam.py
```diff
@@ -1555,3 +1580,525 @@ def test_shim_translates_iam_set_handles_empty_etag(self):
...
+      with SetEnvironmentForTest({
+          'CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL': 'True',
+          'CLOUDSDK_ROOT_DIR': 'fake_dir',
+      }):
+        with self.assertRaisesRegex(CommandException, 'An error.'):
+          self.RunCommand(
+              'iam', ['ch', 'allAuthenticatedUsers:modified-role', 'gs://b/o'])
+        self.assertEqual(mock_run.call_count, 2)
+
+  @mock.patch.object(iam.IamCommand, 'RunCommand', new=mock.Mock())
```
NickGoog, line 1916: Why does `RunCommand` execute if the shim is active?

gslib/tests/test_iam.py
```diff
@@ -1555,3 +1580,525 @@ def test_shim_translates_iam_set_handles_empty_etag(self):
...
+      self.assertEqual(mock_run.call_args_list, [
+          self._get_run_call([
+              os.path.join('fake_dir', 'bin', 'gcloud'), 'alpha', 'storage',
+              'ls', '--json', 'gs://b/o1'
+          ]),
+          self._get_run_call([
+              os.path.join('fake_dir', 'bin', 'gcloud'), 'alpha', 'storage',
+              'objects', 'get-iam-policy', 'gs://b/o1', '--format=json'
+          ]),
+          self._get_run_call([
```
NickGoog, line 2090: Is the error logged anywhere?

rrauber: It is now!

gslib/tests/test_iam.py
rrauber: I've removed those mocks.

gslib/tests/test_iam.py
rrauber: Nice!

gslib/tests/test_iam.py
rrauber: I was going in that direction initially, but these are default dicts which makes the declaration a bit more of a pain, plus it's replicating an implementation detail in BindingsToDict which didn't seem ideal.

gslib/utils/iam_helper.py
rrauber: Users can specify DROP_ALL by omitting part of the bindings descriptor passed to ch, that's how the empty string ends up here. All those members will be grouped under this value, so difference update will remove them from base[role].

gslib/utils/iam_helper.py
rrauber: Done.

gslib/commands/iam.py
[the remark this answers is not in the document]
rrauber: Done!

gslib/commands/iam.py
```diff
@@ -363,8 +374,158 @@ def get_gcloud_storage_args(self):
...
+        ls_output = json.loads(ls_process.stdout)
+        urls = [
+            resource['url']
+            for resource in ls_output
+            if resource['type'] == 'cloud_object'
+        ]
+      else:
+        urls = [str(url_pattern)]
+
+      # With the -r flag, container headings are printed ending in '/:'.
```
NickGoog, line 497: Is this comment still necessary with JSON formatting?

gslib/utils/iam_helper.py
```diff
@@ -52,11 +52,18 @@
...
+# bindings are to be granted or removed from a bucket / object.
+# bindings varies by type.
+# For BindingsTuple, it is a list of BindingsValueListEntry
+# For BindingsDictTuple it is a list of analogous dicts, e.g.:
+# {
+#   'members': ['member'],
+#   'role': 'role',
+# }
 BindingsTuple = namedtuple('BindingsTuple', ['is_grant', 'bindings'])
+BindingsDictTuple = namedtuple('BindingsDictTuple', ['is_grant', 'bindings'])
```
NickGoog, line 66: The identical `namedtuple` is kind of a code smell

NickGoog (approved):

gslib/commands/iam.py
rrauber: Good catch!

gslib/utils/iam_helper.py
rrauber: Agreed, though "bindings" has different semantics here. I think ultimately it will help us keep message-dependent and non-message-dependent code separate.

PR #1728, opened 2023-07-11
Make shim format strings compatible with Windows CMD.

NickGoog:


gslib/utils/shim_util.py
```diff
@@ -185,6 +186,18 @@ def get_flag_from_header(raw_header_key, header_value, unset=False):
...
 
+def get_format_flag_caret():
+  if system_util.IS_WINDOWS:
+    return '^^^^'
+  return '^'
+
+
+def get_format_flag_newline():
+  if system_util.IS_WINDOWS:
+    return '^\^n'
```
rrauber, line 197: Did you mention at one point that the number of escapes needed depended on the formatters we used? Or was that just for notifications.

rrauber (approved):

gslib/utils/shim_util.py
NickGoog: It does depend on the formatter. This seems to work for most formatters because most seem to be processed in the CMD and then Python (versus CMD, then Python, then CMD again, I think). I don't remember the incompatible formatters (maybe the `format` formatter?)

PR #1751, opened 2023-10-23
Added Python 3.12 support

thomasmaclean:


dilipped (approved):
Couple of things to verify

1. Make sure that pip install works on 3.12, just to make sure that we have the right dependency versions there.
2. Re-run parity check for hash.py

PR #1754, opened 2023-11-04
Allow accessing public data with anonymous user  when shim is turned ON

dilipped:
Accessing public data does not require authentication. In gsutil, this works out-of-the-box. In `gcloud storage`, the user has to set `auth/disable_credentials` property to use it without authentication (i.e without setting up an active account). For the shim, in order to not break the workflow of existing users who rely on the gsutil behavior, we will set the `auth/disable_credentials` property automatically if we detect that `gcloud` doesn't have any active account configured.

The change has been tested manually

### Before:
```
$ CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL=1 GCLOUD_BINARY_PATH=/usr/bin/gcloud python3 gsutil.py -o "GSUtil:use_gcloud_storage=True"  ls  gs://pub/test.txt
ERROR: (gcloud.storage.ls) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials.

If you have already logged in with a different account, run:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.

```

### After:
```
$ CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL=1 GCLOUD_BINARY_PATH=/usr/bin/gcloud python3 gsutil.py -o "GSUtil:use_gcloud_storage=True"  ls  gs://pub/test.txt
gs://pub/test.txt
```

### DIFF
--- a/gslib/tests/test_shim_util.py
+++ b/gslib/tests/test_shim_util.py
@@ -406,6 +406,20 @@ class TestTranslateToGcloudStorageIfRequested(testcase.GsUtilUnitTestCase):
         parallel_operations=True,
         bucket_storage_uri_class=mock.ANY,
         gsutil_api_class_map_factory=mock.MagicMock())
+    
+    # Translator calls `gcloud config get account` to check active account
+    # using subprocess.run().
+    # We don't care about this call for most of the tests below, so we are
+    # simply patching this here.
+    # There are separate tests to check this call is being made.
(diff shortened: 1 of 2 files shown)


dilipped:
Closing in favor of https://github.com/GoogleCloudPlatform/gsutil/pull/1755

PR #1755, opened 2023-11-04
Allow accessing public data with anonymous user when shim is turned ON

dilipped:
Accessing public data does not require authentication. In gsutil, this works out-of-the-box. In `gcloud storage`, the user has to set `auth/disable_credentials` property to use it without authentication (i.e without setting up an active account). For the shim, in order to not break the workflow of existing users who rely on the gsutil behavior, we will set the `auth/disable_credentials` property automatically if we detect that `gcloud` doesn't have any active account configured.

The change has been tested manually

### Before:
```
$ CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL=1 GCLOUD_BINARY_PATH=/usr/bin/gcloud python3 gsutil.py -o "GSUtil:use_gcloud_storage=True"  ls  gs://pub/test.txt
ERROR: (gcloud.storage.ls) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials.

If you have already logged in with a different account, run:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.

```

### After:
```
$ CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL=1 GCLOUD_BINARY_PATH=/usr/bin/gcloud python3 gsutil.py -o "GSUtil:use_gcloud_storage=True"  ls  gs://pub/test.txt
gs://pub/test.txt
```

### DIFF
--- a/gslib/tests/testcase/shim_unit_test_base.py
+++ b/gslib/tests/testcase/shim_unit_test_base.py
@@ -44,3 +44,4 @@ class ShimUnitTestBase(GsUtilUnitTestCase):
   def tearDown(self):
     if self._subprocess_run_patcher is not None:
       self._subprocess_run_patcher.stop()
+    super().tearDown()
--- a/gslib/tests/testcase/shim_unit_test_base.py
+++ b/gslib/tests/testcase/shim_unit_test_base.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# Copyright 2023 Google LLC. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Contains base class to be used for shim unit tests."""
+
+import subprocess
+from unittest import mock
+
+from gslib.tests.testcase.unit_testcase import GsUtilUnitTestCase
+
+
+class ShimUnitTestBase(GsUtilUnitTestCase):
+  """Base class for unit testing shim behavior.
+  
+  This class mocks the `subprocess.run()` method because it gets called
+  for all shim operations to check if there is an active account configured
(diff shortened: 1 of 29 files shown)


gslib/tests/testcase/shim_unit_test_base.py
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+"""Contains base class to be used for shim unit tests."""
+
+import subprocess
+from unittest import mock
+
+from gslib.tests.testcase.unit_testcase import GsUtilUnitTestCase
+
+
+class ShimUnitTestBase(GsUtilUnitTestCase):
```
thomasmaclean, line 23: This seems a bit heavy-handed. What was the reason you didn't just add this to `GsUtilUnitTestCase` rather than creating a new class?

gslib/utils/shim_util.py
```diff
@@ -607,6 +629,9 @@ def translate_to_gcloud_storage_if_requested(self):
       try:
         env_variables, gcloud_storage_command = self._get_full_gcloud_storage_execution_information(
             self.get_gcloud_storage_args())
+        if not self.gcloud_has_active_account():
```
thomasmaclean, line 632: What if the boto file has credentials? Are those effectively ignored by the shim? If so, we should probably be clear about that, as I don't believe we specifically mention that in the shim documentation (except that the options are missing under the `[Credentials]` heading).

thomasmaclean (approved):

gslib/tests/testcase/shim_unit_test_base.py
dilipped: This base class mock the `subprocess.run` call. I wanted to limit that to the shim-related tests only as it might be a bit confusing to have that mocked for all other tests that don't even call that function.

gslib/utils/shim_util.py
dilipped: Yes, anything in the boto file gets ignored as we are solely depending on the credentials used by gcloud (that's one of the conditions to use the shim, i.e gsutil and gcloud should be using the same creds.)  I think we have that mentioned for the internal docs, but for external docs, it's missing. I will file a bug.

dilipped: Actually, ended up creating a separate PR for the help text update: https://github.com/GoogleCloudPlatform/gsutil/pull/1756

dilipped:
All kokoro failures are because of known issues. We had a bug filed for one of them. I filed a couple of new bugs to track the other kokoro failures.

PR #1756, opened 2023-11-06
Add credentials information for shim

dilipped:
Added some details related to authentication and credentials when the shim is in use.

### DIFF
--- a/gslib/addlhelp/shim.py
+++ b/gslib/addlhelp/shim.py
@@ -228,8 +228,9 @@ If you do not have ``gcloud`` installed, follow the
 <B>BOTO CONFIGURATION</B>
   Configuration found in the boto file is mapped 1:1 to gcloud environment
   variables where appropriate.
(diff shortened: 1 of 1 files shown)


gslib/addlhelp/shim.py
```diff
@@ -219,6 +228,8 @@
 <B>BOTO CONFIGURATION</B>
   Configuration found in the boto file is mapped 1:1 to gcloud environment
   variables where appropriate.
+The ``gs_oauth2_refresh_token`` and ``gs_service_key_file`` fields are ignored
```
thomasmaclean, line 231: Add a space to make this its own paragraph, or potentially add `Important: ` as a prefix so that it comes out looking like a call out (like this: https://screenshot.googleplex.com/B9yP3tcaAM8N2ag)

thomasmaclean (approved):

gslib/addlhelp/shim.py
dilipped: Done

PR #1767, opened 2024-04-22
Fixes the bug related to download, and upload files with .p12 authentication.

googlyrahman:
1. Fixed a bug with downloading files from cloud with .p12 authentication.
2. Skip test which tests the caching behaviour of .p12 creds.

### DIFF
--- a/gslib/gcs_json_api.py
+++ b/gslib/gcs_json_api.py
@@ -59,6 +59,7 @@ from gslib.cloud_api import ResumableUploadStartOverException
 from gslib.cloud_api import ServiceException
 from gslib.exception import CommandException
 from gslib.gcs_json_credentials import SetUpJsonCredentialsAndCache
+from gslib.gcs_json_credentials import isP12Credentials
 from gslib.gcs_json_media import BytesTransferredContainer
 from gslib.gcs_json_media import DownloadCallbackConnectionClassFactory
 from gslib.gcs_json_media import HttpWithDownloadStream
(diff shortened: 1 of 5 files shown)


gslib/gcs_json_api.py
```diff
@@ -244,15 +244,15 @@ def __init__(self,
     # sequentially, but we can share TCP warmed-up connections across calls.
     self.download_http = self._GetNewDownloadHttp()
     self.upload_http = self._GetNewUploadHttp()
+    WrapDownloadHttpRequest(self.download_http)
```
googlyrahman, line 247: Doing it above due to the mismatch of return type of authorize method in p12 credentials, json key.

setup.py
```diff
@@ -51,7 +51,7 @@
     'retry_decorator>=1.0.0',
     'six>=1.16.0',
     # aiohttp is the extra dependency that contains requests lib.
-    'google-auth[aiohttp]>=2.5.0',
+    'google-auth[aiohttp]==2.17.0',
```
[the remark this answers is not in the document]
googlyrahman: If we don't keep this, then pip would install the latest version of google-auth, due to which there's mismatch in the response, and release test would fail.

gslib/gcs_json_api.py
```diff
@@ -251,8 +252,13 @@ def __init__(self,
     else:
       self.authorized_download_http = self.download_http
       self.authorized_upload_http = self.upload_http
-    WrapDownloadHttpRequest(self.authorized_download_http)
-    WrapUploadHttpRequest(self.authorized_upload_http)
+
+    if isinstance(self.credentials, P12Credentials):
```
shilpi23pandey, line 256: This check is used in other places as well? Can we turn it into a function so we can re-use?

shilpi23pandey (approved):

shilpi23pandey (approved):

sushantkhanna (approved):
Shilpi has already approved this change. Going ahead and approving + merging the change as Shilpi does not have access yet.

PR #1782, opened 2024-05-18
Remove the use of load_pkcs12 from signurl command

googlyrahman:


gslib/commands/signurl.py
```diff
@@ -48,23 +48,33 @@
...
   HAVE_OPENSSL = True
 except ImportError:
   load_privatekey = None
-  load_pkcs12 = None
   sign = None
   HAVE_OPENSSL = False
   FILETYPE_PEM = None
 
+try:
+  from cryptography.hazmat.primitives import hashes
```
shilpi23pandey, line 68: We are removing dependency on pyopenssl and adding dependency on cryptography? Is cryptograpgy included as dependency?

gslib/commands/signurl.py
```diff
@@ -313,18 +323,29 @@ def _GenSignedUrl(key,
...
-  ks = load_pkcs12(ks_contents, passwd)
-  client_email = ks.get_certificate().get_subject().CN
-
-  return ks.get_privatekey(), client_email
-
+def _ReadKeystore(key_string, passwd):
+  key_string = to_bytes(key_string)
+  passwd = to_bytes(passwd)
+  # Cryptography package is not bundled with gsutil, Keeping it at top would throw error.
+  if not HAVE_CRYPTO:
```
shilpi23pandey, line 341: What is the rationale for not bundling it? If I try to use gsutil without this package , this will still result in an error.

gslib/tests/test_signurl.py
shilpi23pandey, line 1: Can we have a test which checks the exception thrown when a user doesnt have a cryptography?

gslib/commands/signurl.py
googlyrahman: We're not removing the dependency of openSSL, we're removing the dependency on load_pkcs12 attribute of openSSL as it's deprecated. We're using cryptography module to decode to .p12 keys.

gslib/commands/signurl.py
googlyrahman: The reasons for keeping it this way are as follows:
1. The gcloud CLI also does not bundle the pyca/cryptography package (probably because of saving space). See the documentation for details: https://cloud.google.com/sdk/crypto
Screenshot 2024-05-29 at 1 03 42 PM

2. We only use this package for dealing with .p12 keys (authentication and signurl flow), and since .p12 keys are being deprecated, and is not recommended, my thought was to avoid including it as a dependency.

gslib/tests/test_signurl.py
```diff
@@ -109,6 +110,19 @@ def testSignUrlInvalidDurationWithUseServiceAccount(self):
     self.assertIn('CommandException: Max valid duration allowed is 12:00:00',
                   stderr)
 
+  @unittest.skipUnless(not HAVE_CRYPTO, 'signurl requires crypto to decode .p12 keys')
```
rrauber, line 113: Have you confirmed that the package is present in the test environment?

gslib/tests/test_signurl.py
```diff
@@ -144,6 +158,7 @@ def testSignUrlWithJSONKeyFileAndObjectGeneration(self):
     self.assertIn('x-goog-credential=' + TEST_EMAIL, stdout)
     self.assertIn('generation=' + object_uri.generation, stdout)
 
+  @unittest.skipUnless(HAVE_CRYPTO, 'signurl requires crypto to decode .p12 keys.')
```
rrauber, line 161: This test doesn't look like it's covering anything specific to P12, so maybe change self._GetKsFile() to self._GetJSONKsFile()

Same with some of the tests below.

gslib/commands/signurl.py
```diff
@@ -313,18 +323,29 @@ def _GenSignedUrl(key,
         signed_headers=signed_headers,
         billing_project=billing_project,
         string_to_sign_debug=string_to_sign_debug)
-    raw_signature = sign(key, string_to_sign, digest)
+    if isinstance(key, PKey):
+      raw_signature = sign(key, string_to_sign, digest)
+    else: 
+      if not HAVE_CRYPTO:
+        raise CommandException(_CRYPTO_IMPORT_ERROR)
+      raw_signature = key.sign(to_bytes(string_to_sign), padding.PKCS1v15(), hashes.SHA256())
```
rrauber, line 331: Could you tell me a bit more about why this `to_bytes` function is necessary? Is it possible for CreatePayload to return either a string or bytes?

rrauber (commented):
If you haven't already, it would probably be worth giving this a go with some actual .p12 files to make sure the output URLs work.

googlyrahman:
> If you haven't already, it would probably be worth giving this a go with some actual .p12 files to make sure the output URLs work.

I've did some manual testing with .p12 keys, and .json keys. Everything works fine. Though i would request you as well to pull changes locally, and try this out!

googlyrahman:
Adressed!

gslib/tests/test_signurl.py
rrauber: My point was if our tests only ever run in an environment without crypto or with crypto, a lot of the behavior relating to .p12 files won't ever be verified as either the tests that require crypto will silently be skipped, or the tests that require the absence of crypto will be silently skipped. Have you confirmed that there are kokoro builds, or other parts of our testing infrastructure, that we run this on that contain crypto and others that don't?

gslib/tests/test_signurl.py
rrauber: That's right, but this test only covers URL encoding behavior -- that's not behavior we need a .p12 file to confirm. 

This concern is related to the one I raised above -- if for some reason the test environment shifts such that it doesn't have crypto, the behavior this test covers will never be verified since it will always be skipped. Removing the skip decorator here and switching to JSON keys means it's more likely to continue running.

To be clear, this comment did not apply to the test cases  above, like testSignUrlRaiseErrorForP12KeysWithoutCrypto or testSignUrlOutputP12, as those explicitly cover behavior related to p12 files.

googlyrahman: Done! Thanks for clarification

gslib/tests/test_signurl.py
googlyrahman: I see, I've raised a CL for adding cryptography into test environment!

rrauber: Approved!

rrauber:
It looks like the tests aren't providing much signal right now because of an org policy constraint. For example, the python 3.11 tests end with 404/930 finished and 323 failures with the following message:

```
ERROR 0610 09:51:24.661415 retry_decorator.py] Retrying in 7.37 seconds ...
  File "/tmpfs/src/github/src/gsutil/third_party/retry-decorator/retry_decorator/retry_decorator.py", line 20, in f_retry
    return f(*args, **kwargs)
  File "/tmpfs/src/github/src/gsutil/gslib/tests/testcase/integration_testcase.py", line 630, in _CreateBucketWithExponentialBackoff
    bucket_uri.create_bucket(storage_class=storage_class,
  File "/tmpfs/src/github/src/gsutil/gslib/vendored/boto/boto/storage_uri.py", line 556, in create_bucket
    return conn.create_bucket(self.bucket_name, headers, location, policy,
  File "/tmpfs/src/github/src/gsutil/gslib/vendored/boto/boto/gs/connection.py", line 116, in create_bucket
    raise self.provider.storage_response_error(
boto.exception.GSResponseError: GSResponseError: 412 Precondition Failed
<?xml version='1.0' encoding='UTF-8'?><Error><Code>OrgPolicyConstraintFailed</Code><Message>Request violates OrgPolicy constraint.</Message><Details>'0' violates constraint 'constraints/storage.softDeletePolicySeconds'</Details></Error>
```

It'd be best if we were able to resolve this before merging, either by opting the test project out of the org policy constraint, or by giving the create bucket call in tests a default value for the field it constrains that's allowed.

googlyrahman:
> File "/tmpfs/src/github/src/gsutil/gslib/vendored/boto/boto/gs/connection.py", line 116, in create_bucket

We've removed the org policy constraint, and it should be fine now.

rrauber (approved):

PR #1791, opened 2024-07-30
Fixes parity failures of cat and rm commands.

kautikdk:
Recently, Error message for non-matching URLs has been modified in `gcloud storage` tool. This change caused failures in parity tests. 
This PR fixes the failures for `gcloud` validation only and doesn't affect any `gsutil` command implementation or gsutil specific validations.

### DIFF
--- a/gslib/tests/test_rm.py
+++ b/gslib/tests/test_rm.py
@@ -745,7 +745,7 @@ class TestRm(testcase.GsUtilIntegrationTestCase):
         expected_status=1)
     if self._use_gcloud_storage:
       self.assertIn(
-          'The following URLs matched no objects or files:\n{}\n-{}'.format(
+          'The following URLs matched no objects or files:\n{}\n{}'.format(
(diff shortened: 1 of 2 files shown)


gslib/tests/test_rm.py
```diff
@@ -745,7 +745,7 @@ def test_rm_multiple_nonexistent_objects(self):
         expected_status=1)
     if self._use_gcloud_storage:
       self.assertIn(
-          'The following URLs matched no objects or files:\n-{}\n-{}'.format(
+          'The following URLs matched no objects or files:\n{}\n-{}'.format(
```
shilpi23pandey, line 748: This one has another `-`. Is this required?

shilpi23pandey (approved):

gslib/tests/test_rm.py
kautikdk: Oops, Missed that! It should be removed.

PR #1812, opened 2024-10-21
Github CI Build only for 3.8 to 3.12

sameer-google:
This update is part of the process of transitioning to Python 3.12 while dropping support for Python 3.5, 3.6, and 3.7. The GitHub CI configuration has been modified to test against Python versions 3.8 through 3.12, ensuring compatibility across supported versions.

PR #1813, opened 2024-10-21
updating version check to 3.12

sameer-google:
This PR updates the Python version compatibility for gsutil to support Python 3.8 through 3.12, while dropping support for older versions (3.5, 3.6, and 3.7). The following changes have been made:

- Updated version check in `gsutil.py` to `3.8-3.12`.
- Updated the `python_requires` and `classifiers` in `setup.py` to specify compatibility with Python versions 3.8 to 3.12, removing references to unsupported versions (3.5, 3.6, and 3.7).

These updates align the project with the ongoing effort to support the latest Python versions while ensuring compatibility across all tested platforms.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,16 +126,16 @@ setup(
         'Natural Language :: English',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
-        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
         'Topic :: System :: Filesystems',
         'Topic :: Utilities',
     ],
-    # Gsutil supports Python 3.5+
-    python_requires='!=2.*, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4',
+    # Gsutil supports Python 3.8 to 3.12
+    python_requires='>=3.8, <=3.12',
     platforms='any',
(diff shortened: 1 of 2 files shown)


sushantkhanna:
Please squash the commits so that only a single commit gets merged into primary branch related to this change.

setup.py
```diff
@@ -126,16 +126,16 @@ def make_release_tree(self, base_dir, files):
         'Natural Language :: English',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
```
sushantkhanna, line 128: Even this line needs to be removed ?

sushantkhanna (approved):

setup.py
sameer-google: It's useful for searchability on package indexes like PyPI, as some users may filter for projects compatible with Python 3 broadly, without specifying a minor version. We should only remove this line if the project does not support all minor versions of Python 3. Correct me if i am wrong.

sameer-google:
> Please squash the commits so that only a single commit gets merged into primary branch related to this change. Please update the description.

Squashed the commits and Updated PR description.

setup.py
sameer-google: As we are updating `python_requires='>=3.8, <=3.12'`, this ensures that we are compatible only with 3.8 to 3.12.

we can keep the `'Programming Language :: Python :: 3'` line, as long as we are not claiming compatibility with all versions of `Python 3` (which could lead to confusion). IMO, It's fine to leave it, especially if we later specify the exact minor versions (e.g., Programming Language :: Python :: 3.8, Programming Language :: Python :: 3.12) as well.

sameer-google:
The kokoro tests failing in this PR are failing in the previous ones' as well.
The CI / build failures are due to a change in version check. These are resolved by [CI build PR](https://github.com/GoogleCloudPlatform/gsutil/pull/1812), merging alongside this PR.

PR #1815, opened 2024-11-11
Replaced crcmod_osx with crcmod_plus

sameer-google:


sameer-google:
Decided to continue with `crcmod_osx`, for python versions greater than `3.12` users must install `crcmod 1.7` package to avoid breakages.

PR #1825, opened 2024-12-05
update version check to 3.13

sameer-google:
Updating version check to allow users with `python 3.13` to use `gsutil`.
Drop `python 3.8` from usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,12 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        python-version: [3.8.x, 3.9.x, 3.10.x, 3.11.x, 3.12.x]
+        python-version: [3.9.x, 3.10.x, 3.11.x, 3.12.x, 3.13.x]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


sameer-google:
> I am not clear on why changes like runs-on, uses, etc are required ? Could you please provide more context ?

- `ubuntu-20.04` reaches EOL by April 2025, so it is suggested to use the latest `ubuntu-24.04`

- `actions/setup-python@v4` and above is recommended for newer python versions, specifically for 3.12+ compatibility similarly `actions/checkout@v4`

zexuan-wang:
Can I get some update of this 3.13 support?

sameer-google:
> Can I get some update of this 3.13 support?

Support for 3.13 is planned by July 2025.

.github/workflows/ci.yml
```diff
@@ -26,14 +26,14 @@ jobs:
...
         git submodule update --init --recursive
-    - name: Lint
-      if: ${{ matrix.python-version  == 3.6 }}
-      run: |
-        set -x
-        pip install pylint
-        pip install yapf
-        test/run_pylint.sh
-        yapf -dr .
+    # - name: Lint
```
sushantkhanna, line 29: I see that this was invoked with 3.6.
If my understanding is correct, this should have ideally been removed with the previous version bump, but we are doing it now since we missed it ?  
If correct, should we not just remove it rather than keep it commented ?

sushantkhanna (approved):

PR #1826, opened 2024-12-06
Added py313 kokoro test configs

sameer-google:
This update adds `Python 3.13` test configurations to the Kokoro CI pipeline for macOS, Linux, and Windows. These changes ensure comprehensive testing across all major platforms.


sushantkhanna (commented):
We do not support 3.13. Why do we want to add a Kokoro test config ?

sameer-google:
> We do not support 3.13. Why do we want to add a Kokoro test config ?

This config is being added as part of supporting 3.13

PR #1827, opened 2024-12-09
Change python_requires from <=3.12 to <3.13

sameer-google:
This pull request updates the `python_requires` field in the `setup.py` file to `<3.13`, addressing an issue in the `v5.32` release where the constraint was incorrectly set to `<=3.12`. This fix ensures that `gsutil` can be installed via `pip` on `Python` versions `3.12.1` and above.


sushantkhanna (approved):
Please ensure that we are compliant with all patches of Python 3.12

PR #1830, opened 2024-12-13
Fixes parity tests failures in windows machines.

kautikdk:
`gsutil rsync` parity tests are failing in windows machine because of pipe character. Pipe character, `|`, is a special character for winodws cmd and it should escaped while being used in command input. Windows cmd treats `^` character as escape character and can be used to escape pipe character as well but it works only for internal windows commands and fails to escape for external commands. 

It appears that `^^^` string is correctly escaping pipe character even though there is no official windows documentation related to this. 

This PR modifies rsync test check which is testing pipe character in regex input. Since it needs to be escaped for windows machines, I have added a separate condition to format input incase of windows shim only mode.

### DIFF
--- a/gslib/tests/test_rsync.py
+++ b/gslib/tests/test_rsync.py
@@ -2999,7 +2999,13 @@ class TestRsync(testcase.GsUtilIntegrationTestCase):
 
     def _Check2():
       """Tests that a regex with a pipe works as expected."""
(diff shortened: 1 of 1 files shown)


shilpi23pandey (approved):

kautikdk:
Hi @shilpi23pandey, Can you please merge this PR? I have tested on Windows instance and it's still the same behavior.

PR #1835, opened 2025-02-05
Add shim support for mb --autoclass

pujawadare:
Added flag translation for `--autoclass` flag of `mb` command. (b/289106309)


shilpi23pandey (commented):
Can we add tests?

pujawadare:
I did not find as such specific tests for flag translation for other flags except rpo and retention flag. I still added basic test to verify the translation.

shilpi23pandey:
> I did not find as such specific tests for flag translation for other flags except rpo and retention flag. I still added basic test to verify the translation.

As discussed, let's add e2e tests for this functionality so we can add them for parity testing.

pujawadare:
Added and tested the integration tests locally to ensure they pass. Will raise a CL to include these tests in the parity pipeline once the PR is merged.

gslib/tests/test_mb.py
```diff
@@ -401,6 +401,26 @@ def test_multiple_json_only_flags_raise_error_with_xml_api(self):
...
+    self.assertRegex(stdout, r"Autoclass:.*Enabled on *.")
+
+  @SkipForXML('The --autoclass flag only works for GCS JSON API.')
+  def test_create_with_invalid_storage_class_with_autoclass_flag(self):
+    bucket_name = self.MakeTempName('bucket')
+    bucket_uri = boto.storage_uri('gs://%s' % (bucket_name.lower()),
+                                  suppress_consec_slashes=False)
+    stderr = self.RunGsUtil(['mb', '--autoclass', '-c', 'archive', suri(bucket_uri)], return_stderr=True, expected_status=1)
+    if self._use_gcloud_storage:
+      self.assertIn('HTTPError 400: Cannot set default storage class on bucket with Autoclass enabled to storage class other than STANDARD.', stderr)
```
shilpi23pandey, line 420: Can we use regexes here and assert only on the most significant part. Otherwise, We'll have a hard time maintaining this test for even smallest changes in error messages.

gslib/tests/test_mb.py
```diff
@@ -401,6 +401,26 @@ def test_multiple_json_only_flags_raise_error_with_xml_api(self):
         ' -b option(s) can only be used for GCS Buckets with the JSON API',
         stderr)
 
+  @SkipForXML('The --autoclass flag only works for GCS JSON API.')
+  def test_create_with_autoclass_flag(self):
+    bucket_name = self.MakeTempName('bucket')
```
shilpi23pandey, line 406: I am assuming that the bucket cleanup will be handled by the test framework. Is that correct? If not, we should ensure that in our tests.

pujawadare: Yes, the test framework handles the test setup and the cleanup after execution of test.

shilpi23pandey (approved):

PR #1855, opened 2025-04-02
Increase timeout for tests in gsutil from 180s to 360s

sameer-google:


shilpi23pandey (commented):
Is this number based on some investigation or is it trial and error? 
Do we know if this would fix the timeout issue?

sameer-google:
> Is this number based on some investigation or is it trial and error? Do we know if this would fix the timeout issue?

The number is experimental as of now we do not have a way to decide the timeout number. Referred to other pipelines such as autogen to get to this number.

PR #1859, opened 2025-04-15
Updating handwritten shim tests for ls -L flag.

pujawadare:


googlyrahman (approved):

gslib/tests/test_ls.py
```diff
@@ -679,20 +679,19 @@ def test_logging(self):
     stdout = self.RunGsUtil(['ls', '-lb', bucket_suri], return_stdout=True)
     self.assertNotIn('Logging configuration', stdout)
 
-    spacing = '       ' if self._use_gcloud_storage else '\t\t'
     # Logging configuration is absent by default
     stdout = self.RunGsUtil(['ls', '-Lb', bucket_suri], return_stdout=True)
-    self.assertIn('Logging configuration:{}None'.format(spacing), stdout)
```
sushantkhanna, line 685: You might want to maintain using the `.format()` method and the `spacing` variable.  
Keep only the `\t\t` as value there.

The reason is that it helps add conditional spacings in the future if at all required.

pujawadare: We would not make changes with gsutil/shim in future. Would future proofing be still required?

sushantkhanna: Not future proofing per say, but just a good practice. We may still want to update tests/make some changes as we go along with migration.

pujawadare: Done.

PR #1869, opened 2025-06-10
Add -m option when it is used to user agent to track usage

sushantkhanna:
This change appends an addition `-m` to the command information set in UserAgent.

This is being added to understand the usage of the `-m` option across commands.


sameer-google (commented):
LGTM. The pipeline should run once the branch is rebased to origin/master.

shilpi23pandey (approved):

PR #1907, opened 2026-03-05
Update migration guide to add caveat for local copy in gcloud storage cp

sushantkhanna:


sameer-google (approved):
Does this change also reflect the public migration docs.

sushantkhanna:
> Does this change also reflect the public migration docs.

Technically, this is the detailed part of the public docs. But you are right, maybe adding this in the main - https://docs.cloud.google.com/storage/docs/gsutil-transition-to-gcloud is beneficieal as well. I will reach out to the team that maintains that doc.

shilpi23pandey (approved):

PR #1914, opened 2026-04-15
update oauth2client submodule to latest commit

gurusai-voleti:
Update third party dependencies in gsutil for PyOpenSSL 26.0.0 support.

- oauth2client - upgraded to latest commit d37c2bebffd9ce360a566ffa01f9024719b745ef


sameer-google (approved):

sameer-google:
The windows-py314-xml test is being failing since it was introduced, we have to take a look and fix that test as a separate task later.

PR #1915, opened 2026-04-15
chore: adding pyopenssl 26.0.0 support

gurusai-voleti:


setup.py
```diff
@@ -47,7 +47,8 @@
     # 3.0.5 is the last version that supports Python 3.3 or lower.
     'mock>=2.0.0, <=3.0.5; python_version < "3.3"',
     'monotonic>=1.4',
-    'pyOpenSSL>=0.13, <=24.2.1',
+    'pyOpenSSL>=26.0.0',
+    'cryptography>=46.0.0',
```
sameer-google: do we have to explicitly ass cryptography as a dependency or pyOpenSSL does come with crytography?

gslib/tests/test_signurl.py
```diff
@@ -240,7 +240,43 @@ def testSignUrlOfNonObjectUrl(self):
     self.RunGsUtil(['signurl', self._GetJSONKsFile(), 'gs://'],
                    expected_status=1)
```
sameer-google, line 240: New tests looks good but does these tests cover all possible scenarios of sign url, if not please add e2e tests that cover all cases of the command.

gslib/commands/signurl.py
```diff
@@ -76,6 +64,7 @@
   HAVE_CRYPTO = False
   _CRYPTO_IMPORT_ERROR = "pyca/cryptography is not available. Either install it, or please consider using the .json keyfile"
 
+HAVE_OPENSSL = HAVE_CRYPTO
```
sameer-google: I understand this is a hack to avoid refactoring, this can lead to a confusion too, as we are completely moving away from openSSL to crypto in these files, can you check if it is viable to have a single boolean HAVE_CRYPTO and get rid of HAVE_OPENSSL?

sameer-google (commented):
Overall these changes work for the pyOpenSSL upgrade, can you check gslib/command.py to understand if we still need _CryptoRandomAtFork()?

sameer-google:
Also, can you update the PR title to a more clear name says upgrading the pyopenssl or adding pyopenssl 26.0.0 support etc.

gurusai-voleti:
> Also, can you update the PR title to a more clear name says upgrading the pyopenssl or adding pyopenssl 26.0.0 support etc.

Updated the title

setup.py
gurusai-voleti: cryptography is transitive dependency which comes with pyOpenSSL, removing it from setup.py

gslib/commands/signurl.py
gurusai-voleti: updated the logic to completely avoid HAVE_CRYPTO flag

setup.py
```diff
@@ -47,7 +47,7 @@
     # 3.0.5 is the last version that supports Python 3.3 or lower.
     'mock>=2.0.0, <=3.0.5; python_version < "3.3"',
     'monotonic>=1.4',
-    'pyOpenSSL>=0.13, <=24.2.1',
+    'pyOpenSSL>=26.0.0',
```
sameer-google: We should just increase the upper limit to 26.0.0 rather than setting the lower limit as 26.0.0, users can use any pyOpenSSL version below this. so can we change this to `'pyOpenSSL>=0.13, <=26.0.0',`, One more reason to do this is we would never know what future pyOpenSSL versions changes which could break gsutil again.

gslib/commands/signurl.py
sameer-google: As this refactoring introduces new changes or misses any potential references to the removed `HAVE_OPENSSL`, once all the changes are done can you attach a testing log of 5 sign url scenarios discussed with all combinations hiding sensitive data?

setup.py
gurusai-voleti: updated the version as suggested

gslib/commands/signurl.py
gurusai-voleti: attached the logs on the ticket and tagged you

gslib/tests/test_signurl.py
gurusai-voleti: I see all the existing e2e test cases satisfies all functionalities, correct me if I am wrong

sameer-google: LGTM

gslib/commands/signurl.py
sameer-google: Please add the manual testing logs too with command invocation locally and results are verified.

gurusai-voleti: added manual testing logs link in the ticket, please verify

sameer-google: thanks i see the manual testing logs now

sameer-google:
> Overall these changes work for the pyOpenSSL upgrade, can you check gslib/command.py to understand if we still need _CryptoRandomAtFork()?

@gurusai-voleti please raise a seperate PR for this change, this is a dead code removal which is optional but still some scanners might report usage of `crypto.Random` in gslib/command.py as we are importing it, given the import is already included in try except block the application won't crash.

sameer-google (approved):

PR #1917, opened 2026-04-23
chore: remove unused code in command.py as upgraded to pyOpenSSL 26.0.0

gurusai-voleti:
The original code in gslib/command.py was wrapped in a try...except ImportError: block. Because we removed pycrypto from the dependencies, if gsutil ran that code today, it would simply fail to find the library, catch the ImportError gracefully, and move on. It wouldn't crash the application.

Even though it is "dead code" now, leaving from Crypto import Random in gslib/command.py is a huge risk for security scanners. If a scanner sees from Crypto ..., it will flag gsutil for using pycrypto (CVE-2018-6594), causing false positives and blocking customer deployments.

By physically deleting the _CryptoRandomAtFork function and its imports, we guarantee a clean audit and completely sever gsutil's ties to the deprecated library. Furthermore, modern cryptographic libraries (like cryptography) handle process forking natively and securely, so the workaround is completely obsolete anyway.

### DIFF
--- a/gslib/command.py
+++ b/gslib/command.py
@@ -102,13 +102,6 @@ from gslib.utils.translation_helper import PRIVATE_DEFAULT_OBJ_ACL
 from gslib.wildcard_iterator import CreateWildcardIterator
 from six.moves import queue as Queue
 
-# pylint: disable=g-import-not-at-top
-try:
-  from Crypto import Random as CryptoRandom
-except ImportError:
-  CryptoRandom = None
-# pylint: enable=g-import-not-at-top
-
 OFFER_GSUTIL_M_SUGGESTION_THRESHOLD = 5
 OFFER_GSUTIL_M_SUGGESTION_FREQUENCY = 1000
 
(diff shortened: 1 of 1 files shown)


sameer-google:
can you update the description with the following:

```
The original code in gslib/command.py was wrapped in a try...except ImportError: block. Because we removed pycrypto from the dependencies, if gsutil ran that code today, it would simply fail to find the library, catch the ImportError gracefully, and move on. It wouldn't crash the application.

Even though it is "dead code" now, leaving from Crypto import Random in gslib/command.py is a huge risk for security scanners. If a scanner sees from Crypto ..., it will flag gsutil for using pycrypto (CVE-2018-6594), causing false positives and blocking customer deployments.

By physically deleting the _CryptoRandomAtFork function and its imports, we guarantee a clean audit and completely sever gsutil's ties to the deprecated library. Furthermore, modern cryptographic libraries (like cryptography) handle process forking natively and securely, so the workaround is completely obsolete anyway.
```

sameer-google:
I see all necessary tests are green, can you perform manual sign url testing logs for this change too and close this

sameer-google (approved):

gurusai-voleti:
> I see all necessary tests are green, can you perform manual sign url testing logs for this change too and close this

all automation and manual test cases are green

PR #1932, opened 2026-06-23
test: add and expand unit, integration, and shim tests for various gsutil components

risha-vijayvargiya-22:
S3 Test Invocation = https://paste.googleplex.com/5866418145984512

### DIFF
--- a/gslib/tests/test_bucketpolicyonly.py
+++ b/gslib/tests/test_bucketpolicyonly.py
@@ -16,12 +16,32 @@
 
 from __future__ import absolute_import
 import re
+from unittest import mock
 
+from gslib import exception
+from gslib.commands import bucketpolicyonly
 from gslib.cs_api_map import ApiSelector
 import gslib.tests.testcase as testcase
 from gslib.tests.util import ObjectToURI as suri
+from gslib.tests.util import SetBotoConfigForTest
+from gslib.tests.util import SetEnvironmentForTest
 from gslib.tests.util import unittest
+from gslib.utils import shim_util
 from gslib.utils.retry_util import Retry
+
+
+class TestBucketPolicyOnlyUnit(testcase.GsUtilUnitTestCase):
+
+  def test_set_invalid_mode_fails(self):
+    with self.assertRaisesRegex(exception.CommandException,
+                                'Only on and off values allowed'):
+      self.RunCommand('bucketpolicyonly', ['set', 'invalid_mode', 'gs://bucket'])
+
+  def test_s3_fails(self):
+    bucket_uri = self.CreateBucket(provider='s3')
+    with self.assertRaisesRegex(exception.CommandException,
+                                'only be used with gs:// bucket URLs'):
+      self.RunCommand('bucketpolicyonly', ['set', 'on', suri(bucket_uri)])
 
 
 class TestBucketPolicyOnly(testcase.GsUtilIntegrationTestCase):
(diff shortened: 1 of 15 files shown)


gslib/tests/testcase/integration_testcase.py
```diff
@@ -320,6 +320,38 @@ def SetPOSIXMetadata(self,
                                        object_name,
                                        obj_metadata,
                                        provider=provider)
+    if provider != 'gs':
```
sameer-google: PR focuses on JSON API test coverage. Modifying S3 (provider != 'gs') POSIX metadata retry logic is completely unrelated to the added JSON tests. Can you explain why these changes required? If optional, we can revert these?

gslib/tests/test_gcs_json_media.py
```diff
@@ -175,3 +179,53 @@ def test_incomplete_download_sets_content_length_correctly(self):
...
+
+  @mock.patch('six.moves.http_client.HTTPConnection.getresponse')
+  def testReadUpdatesDigesterAndCallback(self, mock_getresponse):
+    mock_response = mock.Mock()
+    mock_response.status = http_client.OK
+    mock_response.read.return_value = b'sample-data'
+    mock_getresponse.return_value = mock_response
+
+    wrapped_response = self.instance.getresponse()
+    data = wrapped_response.read(11)
```
sameer-google, line 223: can you add a test case with sequential chunked reads (e.g., read(5) then read(6)) to explicitly verify the hashlib digester and callbacks accumulate state correctly. Currently just a single read call is being tested and production downloads are chunked.

gslib/tests/test_gcs_json_media.py
```diff
@@ -175,3 +179,53 @@ def test_incomplete_download_sets_content_length_correctly(self):
...
+  def testReadInvalidAmtRaisesBadRequest(self, mock_getresponse):
+    mock_response = mock.Mock()
+    mock_response.status = http_client.OK
+    mock_getresponse.return_value = mock_response
+
+    wrapped_response = self.instance.getresponse()
+
+    # 1. Size exceeding TRANSFER_BUFFER_SIZE should raise BadRequestException
+    with self.assertRaises(BadRequestException):
+      wrapped_response.read(TRANSFER_BUFFER_SIZE + 1)
```
sameer-google, line 209: The assert that read(TRANSFER_BUFFER_SIZE + 1) and read(None) raise exceptions, but did we miss read(-1) (read to EOF) and read(0). Add coverage for these cases.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -230,3 +230,29 @@ def testOauth2ServiceAccountAndOauth2UserCredential(self):
...
+            "keyfile": tmpfile,
+            "client_id": "?",
+        })):
+      with self.assertRaisesRegex(Exception, 'did not contain the required entry: private_key'):
+        gcs_json_api.GcsJsonApi(None, logging.getLogger(), None, None)
+
+  @unittest.skipUnless(HAS_CRYPTO, 'p12credentials requires cryptography.')
+  def testPKCS12SignerInvalidKeyfileRaises(self):
+    with self.assertRaisesRegex(CommandException, 'Unable to load the keyfile'):
+      gcs_json_credentials.CreateP12ServiceAccount(b'invalid-p12-data', b'password')
```
sameer-google, line 270: Passing entirely malformed bytes (b'invalid-p12-data') bypasses the password validation logic. Add a test case that provides a structurally valid P12 file but an incorrect password.

gslib/tests/test_gcs_json_credentials.py
```diff
@@ -230,3 +230,29 @@ def testOauth2ServiceAccountAndOauth2UserCredential(self):
...
+    tmpfile = self.CreateTempFile(contents=b'{"invalid_json": ')
+    with SetBotoConfigForTest(
+        getBotoCredentialsConfig(service_account_creds={
+            "keyfile": tmpfile,
+            "client_id": "?",
+        })):
+      with self.assertRaisesRegex(Exception, 'Could not parse JSON keyfile'):
+        gcs_json_api.GcsJsonApi(None, logging.getLogger(), None, None)
+
+  def testServiceAccountJsonMissingFieldsRaises(self):
```
sameer-google, line 244: Testing only for the omission of private_key is insufficient. Parameterize this test to drop other mandatory fields (e.g., client_email) to ensure the validation isn't hardcoded to a single key.

gslib/tests/test_bucketpolicyonly.py
```diff
@@ -95,3 +115,110 @@ def testTooFewArgumentsFails(self):
                             return_stderr=True,
                             expected_status=1)
     self.assertIn('command requires at least', stderr)
+
+  def test_get_nonexistent_bucket(self):
+    if self.test_api == ApiSelector.XML:
+      return unittest.skip('XML API has no concept of Bucket Policy Only')
```
sameer-google: Using return unittest.skip(...) inside a test method simply returns the decorator object and exits the function. This causes the test runner to incorrectly report the test as 'PASSED' instead of 'SKIPPED', hiding the skip status entirely. Please use the @SkipForXML decorator exactly like you did in test_autoclass.py, or raise unittest.SkipTest(...). Same for next 2 tests.

gslib/tests/test_cloud_api_delegator.py
```diff
@@ -51,3 +54,61 @@ def testRaisesErrorIfMtlsUsedWithXml(self, mock_get_context_config):
 
     with self.assertRaises(cloud_api.ArgumentException):
       delegator.GetApiSelector(provider='s3')
+
+  @mock.patch.object(boto_util, 'UsingGsHmac')
+  @mock.patch.object(config, 'has_option')
```
sameer-google, line 59: Mocking boto.config.has_option to blindly return True or False for all arguments hides bugs if the code under test checks other configuration sections/options simultaneously (e.g., gs_oauth2_refresh_token). Use side_effect with a lambda to strictly only return the mocked value when the specific option (e.g., 'encryption_key') is queried. Same forother tests in this file.

sameer-google (commented):
Other files LGTM, address the ones has comments or explanations.

gslib/tests/test_gcs_json_media.py
risha-vijayvargiya-22: These cases are not covered in the main function itself, do you want me to add that cases in gcs_json_media.py and then I can add in test file as well.

sameer-google: Let's keep this PR strictly scoped to adding test coverage without altering any production tool functionality. Since read(-1) and read(0) expose a pre-existing flaw in the gcs_json_media.py logic, let's intentionally leave those tests out of this PR so we don't block it. Please revert any production code changes if you started them. I will approve this PR as-is and we can track the read(-1) bug in a separate issue.

gslib/tests/testcase/integration_testcase.py
```diff
@@ -338,6 +337,7 @@ def ClearPOSIXMetadata(self, obj):
         suri(obj)
     ])
 
+
```
sameer-google: can you clear the blank changes here so this file is completeyl out of this PR.

sameer-google (approved):
Address the pending comments LGTM.

PR #1934, opened 2026-07-06
test: add comprehensive unit test coverage for stet, tracker file, and retention utility modules

risha-vijayvargiya-22:
S3 Test Invocation = https://paste.googleplex.com/5360855947542528

### DIFF
--- a/gslib/tests/test_retention_util.py
+++ b/gslib/tests/test_retention_util.py
@@ -16,6 +16,8 @@
 
 from __future__ import absolute_import
 
+from unittest import mock
+
 import gslib.tests.testcase as testcase
 from gslib.utils.retention_util import _RetentionPeriodToString
 from gslib.utils.retention_util import DaysToSeconds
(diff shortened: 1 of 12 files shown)


gslib/tests/test_retention_util.py
```diff
@@ -16,6 +16,8 @@
 
 from __future__ import absolute_import
 
+from unittest import mock
```
sameer-google: As discussed in previous PRs, the gsutil standard for mocking is to use from six.moves import mock rather than importing unittest.mock directly. Please fix this here, as well as the identical issues introduced in gslib/tests/test_rsync_funcs.py and gslib/tests/test_versioning.py.

gslib/tests/test_versioning.py
```diff
@@ -19,9 +19,14 @@
...
 from __future__ import unicode_literals
 
+from gslib.commands import versioning
 import gslib.tests.testcase as testcase
 from gslib.tests.util import ObjectToURI as suri
+from gslib.tests.util import SetBotoConfigForTest
+from gslib.tests.util import SetEnvironmentForTest
+from gslib.utils import shim_util
 from gslib.utils.retry_util import Retry
+from unittest import mock
```
sameer-google: When moving to from six.moves import mock please ensure the import is placed in the correct group at the top of the file. Standard library imports and third-party dependencies (six) should be grouped together and placed before local gslib.* imports.

gslib/tests/test_retention_util.py
```diff
@@ -155,3 +157,151 @@ def testRetentionPeriodToString(self):
     retention_str = _RetentionPeriodToString(SECONDS_IN_YEAR + 1)
     self.assertRegex(retention_str,
                      r'Duration: 31557601 Seconds \(~365 Day\(s\)\)')
+
+  def testRetentionInSecondsInvalidRaises(self):
+    from gslib.exception import CommandException
```
sameer-google: Please move local imports like from gslib.exception import CommandException to the top of the file rather than importing them inline inside the test method. This applies to several other tests introduced in this file (e.g., inside testRetentionPolicyToString, testConfirmLockRequest, testUpdateObjectMetadataExceptionHandler, etc.).

gslib/tests/test_stet_cp.py
```diff
@@ -151,3 +151,29 @@ def test_does_not_seek_ahead_for_bytes_if_stet_transform(self):
       # Post-encryption total file size:
       # +-0.1 KiB b/c of rounding and different platforms.
       self.assertRegex(stderr, r'2\.\d KiB]')
+
+  def test_stet_failure_raises(self):
+    # Create a STET binary that exits with error code 1
+    failing_stet_binary = """#!/bin/sh
```
sameer-google: Creating a dummy script with #!/bin/sh will fail when these tests run on Windows machines since /bin/sh is not available natively. Consider using a Python script with #!/usr/bin/env python for the dummy STET binary, or explicitly skip this test on Windows using @unittest.skipIf(IS_WINDOWS, ...).

gslib/tests/test_ubla.py
```diff
@@ -100,3 +101,34 @@ def testTooFewArgumentsFails(self):
...
+      # Command fails on XML immediately. We force XML API for gs provider by
+      # using HMAC credentials.
+      with SetBotoConfigForTest([
+          ('Credentials', 'gs_access_key_id', 'dummy_key'),
+          ('Credentials', 'gs_secret_access_key', 'dummy_secret'),
+          ('Credentials', 'gs_oauth2_refresh_token', None),
+          ('Credentials', 'gs_service_client_id', None),
+          ('Credentials', 'gs_service_key_file', None),
+      ]):
+        stderr = self.RunGsUtil(['ubla', 'get', 'gs://foo'], expected_status=1, return_stderr=True)
```
sameer-google: This line is 105 characters long, which exceeds our standard 80-character limit. Please break this and other long self.RunGsUtil calls across multiple lines. Similar violations exist on other lines, as well as in test_versioning.py and test_web.py.

gslib/tests/test_rsync_funcs.py
```diff
@@ -80,3 +81,106 @@ def test_compute_needed_file_checksums(self):
...
+    from gslib.commands.rsync import _BatchSort
+    import tempfile
+
+    lines = [
+        'gs://bucket/object_c 10 123456789 - - - - - -\n',
+        'gs://bucket/object_a 20 123456789 - - - - - -\n',
+        'gs://bucket/object_b 30 123456789 - - - - - -\n',
+    ]
+
+    with tempfile.NamedTemporaryFile(mode='w+', delete=False) as out_file:
```
sameer-google: Instead of manually creating a tempfile.NamedTemporaryFile and wrapping it in a try/finally block to call os.unlink(), you can use the test framework's helper self.CreateTempFile(). It automatically handles test file creation and cleanup.

sameer-google (approved):
update the branch to master, LGTM

sameer-google:
Windows test failure in `test_update.py`
The `test_EnsureDirsSafeForUpdate` test is consistently failing on Windows with `AssertionError: CommandException not raised`. 
The `gsutil` method `_EnsureDirsSafeForUpdate` safely normalizes directories using `d.lstrip(os.sep)`. Because the test hardcodes forward-slashes (e.g. `'/usr'`), it strips properly on Linux (`os.sep == '/'`), but fails to strip on Windows (`os.sep == '\\'`). As a result, the check is skipped and the exception is never raised.
**Suggested Fix:** Use standard path joining or `os.sep` for the test values.
```python
import os
...
for unsafe_dir in (os.sep + 'usr', os.sep + 'etc', 'BOOT', os.sep + 'lib32'):

PR #1935, opened 2026-07-08
test: add comprehensive unit test coverage for various files

risha-vijayvargiya-22:
S3 Test Invocation = https://paste.googleplex.com/5777959045693440

### DIFF
--- a/gslib/tests/test_command_runner.py
+++ b/gslib/tests/test_command_runner.py
@@ -481,6 +481,52 @@ class TestCommandRunnerUnitTests(testcase.unit_testcase.GsUtilUnitTestCase):
     with self.assertRaisesRegex(CommandException, r'Invalid non-ASCII'):
       HandleHeaderCoding(headers)
 
+  def test_skip_update_check(self):
+    from gslib.utils import boto_util
+    prev_loglevel = logging.getLogger().getEffectiveLevel()
+    try:
+      # Scenario 1: Standard interactive + INFO logging -> returns False
+      self.running_interactively = True
+      logging.getLogger().setLevel(logging.INFO)
+      with mock.patch.object(boto_util, 'HasUserSpecifiedGsHost', return_value=False):
+        self.assertFalse(self.command_runner.SkipUpdateCheck())
+
+      # Scenario 2: Non-interactive -> returns True
+      self.running_interactively = False
+      with mock.patch.object(boto_util, 'HasUserSpecifiedGsHost', return_value=False):
+        self.assertTrue(self.command_runner.SkipUpdateCheck())
+
+      # Scenario 3: Interactive but quiet log level -> returns True
+      self.running_interactively = True
+      logging.getLogger().setLevel(logging.WARNING)
+      with mock.patch.object(boto_util, 'HasUserSpecifiedGsHost', return_value=False):
+        self.assertTrue(self.command_runner.SkipUpdateCheck())
+
+      # Scenario 4: User specified custom GS host -> returns True
+      self.running_interactively = True
+      logging.getLogger().setLevel(logging.INFO)
+      with mock.patch.object(boto_util, 'HasUserSpecifiedGsHost', return_value=True):
+        self.assertTrue(self.command_runner.SkipUpdateCheck())
+    finally:
+      logging.getLogger().setLevel(prev_loglevel)
+
+  def test_run_invalid_command_suggests_match(self):
+    import io
+    mock_stderr = io.StringIO()
+    with mock.patch('sys.stderr', mock_stderr):
+      with self.assertRaisesRegex(CommandException, 'Invalid command "lss".'):
+        self.command_runner.RunNamedCommand('lss')
+    self.assertIn('Did you mean this?', mock_stderr.getvalue())
+    self.assertIn('\tls', mock_stderr.getvalue())
+
+  def test_run_named_command_translates_help(self):
+    mock_help_class = mock.Mock()
+    with mock.patch.dict(self.command_runner.command_map, {'help': mock_help_class}):
+      self.command_runner.RunNamedCommand('ls', args=[command_runner._StringToSysArgType('--help')])
+      mock_help_class.assert_called_once()
+      args, _ = mock_help_class.call_args
+      self.assertEqual(args[1], ['ls'])
+
 
 class TestCommandRunnerIntegrationTests(testcase.GsUtilIntegrationTestCase):
   """Integration tests for gsutil update check in command_runner module."""
--- a/gslib/tests/test_command_runner.py
+++ b/gslib/tests/test_command_runner.py
@@ -46,6 +46,7 @@ import gslib.tests.util as util
 from gslib.tests.util import ARGCOMPLETE_AVAILABLE
 from gslib.tests.util import SetBotoConfigForTest
 from gslib.tests.util import unittest
+from gslib.utils import boto_util
 from gslib.utils import system_util
 from gslib.utils.constants import GSUTIL_PUB_TARBALL
 from gslib.utils.text_util import InsistAscii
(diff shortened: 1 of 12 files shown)


gslib/tests/test_command_runner.py
```diff
@@ -481,6 +481,52 @@ def test_invalid_header_coding_fails(self):
     with self.assertRaisesRegex(CommandException, r'Invalid non-ASCII'):
       HandleHeaderCoding(headers)
 
+  def test_skip_update_check(self):
+    from gslib.utils import boto_util
```
sameer-google: We just discussed this in the previous PR (#1934). Please do not place standard, non-optional imports like import io or from gslib.utils import boto_util inline inside test methods. They must go at the top of the file. (Note: The inline import of google_reauth in test_gsutil.py is perfectly fine since it's wrapped in a try/except ImportError for an optional dependency, but io and boto_util are standard/mandatory).

gslib/tests/test_kms.py
```diff
@@ -255,6 +256,37 @@ def testEncryptionSetKeyFailsWhenUpdateKeyPolicyFailsWithoutWarningFlag(
...
 
+  def test_kms_invalid_subcommand(self):
+    try:
+      self.RunCommand('kms', ['invalid_subcommand'])
+      self.fail('Did not get expected CommandException')
+    except CommandException as e:
+      self.assertIn('Invalid subcommand "invalid_subcommand" for the kms command.', e.reason)
+
+  def test_warn_without_encryption_subcommand_fails(self):
+    try:
```
sameer-google: Instead of manually writing a try... except CommandException as e: block and calling self.fail(), please use the with self.assertRaisesRegex(CommandException, '...'): context manager. This is the idiomatic unittest standard and is used everywhere else in the codebase. Please apply this to test_kms_invalid_subcommand, test_warn_without_encryption_subcommand_fails, test_warn_without_key_option_fails, and test_authorize_without_key_fails.

gslib/tests/test_du.py
```diff
@@ -21,12 +21,21 @@
...
+from gslib.commands import du
 import gslib.tests.testcase as testcase
 from gslib.tests.testcase.integration_testcase import SkipForS3
 from gslib.tests.util import GenerationFromURI as urigen
 from gslib.tests.util import ObjectToURI as suri
+from gslib.tests.util import SetBotoConfigForTest
+from gslib.tests.util import SetEnvironmentForTest
 from gslib.utils.constants import UTF8
 from gslib.utils.retry_util import Retry
+from gslib.utils import shim_util
```
sameer-google: The import from gslib.utils import shim_util is out of alphabetical order relative to the other gslib imports block. It should be placed above from gslib.utils.constants import UTF8.

gslib/tests/test_du.py
```diff
@@ -292,3 +301,67 @@ def _Check():
...
+
+  @mock.patch.object(du.DuCommand, 'RunCommand', new=mock.Mock())
+  def test_shim_translates_du_excludes(self):
+    with SetBotoConfigForTest([('GSUtil', 'use_gcloud_storage', 'True'),
+                               ('GSUtil', 'hidden_shim_mode', 'dry_run')]):
+      with SetEnvironmentForTest({
+          'CLOUDSDK_CORE_PASS_CREDENTIALS_TO_GSUTIL': 'True',
+          'CLOUDSDK_ROOT_DIR': 'fake_dir',
+      }):
+        mock_log_handler = self.RunCommand('du', ['-e', '*.tmp', '-X', 'exclude.txt', 'gs://bucket'],
```
sameer-google: This line is 80+ characters long, significantly exceeding our 80-character limit. There are several other lines in this PR that also violate this rule and need to be wrapped. Please wrap long lines in test_copy_helper_funcs.py , test_kms.py, test_hmac.py, test_defacl.py.

gslib/tests/test_daisy_chain_wrapper.py
```diff
@@ -351,3 +354,29 @@ def testInvalidSeek(self):
       self.fail('Expected exception')
     except IOError as e:
       self.assertIn('Invalid seek during daisy chain', str(e))
+
+  def testInvalidReadSize(self):
```
sameer-google, line 358: In testInvalidReadSize, you check read(TRANSFER_BUFFER_SIZE + 1) and read(None). However, looks like we missed testing read(-1), we MUST add coverage to explicitly assert that daisy_chain_wrapper.read(-1) correctly throws a BadRequestException and doesn't bypass limits.

gslib/tests/test_cors.py
```diff
@@ -292,6 +292,19 @@ def testTooFewArgumentsFails(self):
     stderr = self.RunGsUtil(['cors'], return_stderr=True, expected_status=1)
     self.assertIn('command requires at least', stderr)
 
+  def test_cors_invalid_subcommand(self):
+    stderr = self.RunGsUtil(['cors', 'invalid', 'gs://bucket'],
+                            expected_status=1,
+                            return_stderr=True)
+    self.assertIn('Invalid subcommand "invalid"', stderr)
+
+  def test_cors_set_multi_provider_fails(self):
```
sameer-google, line 301: test_cors_set_multi_provider_fails is added to ensure cors set gs:// s3:// correctly throws an error. However, you missed adding the equivalent test for cors get gs:// s3://. Please add test_cors_get_multi_provider_fails for complete coverage.

gslib/tests/test_gsutil.py
```diff
@@ -261,3 +261,50 @@ def test_translates_oauth_error_standalone(self, mock_output_and_exit,
           'Your credentials are invalid. For more help, see '
           '"gsutil help creds", or re-run the gsutil config command (see '
           '"gsutil help config").', fake_error)
+
+  @mock.patch.object(system_util, 'InvokedViaCloudSdk', autospec=True)
+  @mock.patch.object(gslib.__main__, '_OutputAndExit', autospec=True)
+  def test_translates_saml_error_cloudsdk(self, mock_output_and_exit,
```
sameer-google: The tests test_translates_saml_error_cloudsdk and test_translates_saml_error_standalone are 15-line copy-pastes of each other (including the google_reauth import boilerplate). They only differ by a single boolean flag and the expected assert string. Please consolidate these into a single test method that loops over a list of (is_cloudsdk, expected_cmd) tuples.

gslib/tests/test_copy_helper_funcs.py
```diff
@@ -871,3 +871,93 @@ def testCheckCloudHashesIsSkippedCorrectly(self):
...
                         src_obj_metadata=FakeObject(md5Hash='a'),
                         dst_obj_metadata=FakeObject(md5Hash='b'))
+
+  def testConstructDstUrlStreamDisallowedWithPosix(self):
+    # If exp_dst_url is a stream and preserve_posix=True, should raise CommandException
+    src_url = StorageUrlFromString('file:///tmp/src')
+    exp_src_url = StorageUrlFromString('file:///tmp/src')
+    exp_dst_url = StorageUrlFromString('file://-')  # stream representation in gsutil
+    with self.assertRaisesRegex(CommandException, 'Cannot preserve POSIX attributes with a stream.'):
+      copy_helper.ConstructDstUrl(
```
sameer-google: copy_helper.ConstructDstUrl is here five separate times, only changing 1 or 2 boolean flags per test. Please abstract this into a private helper method (e.g., _call_construct_dst_url(self, src, exp_src, exp_dst, **kwargs)) that sets the defaults for the boolean flags to drastically reduce this duplication.

gslib/tests/test_command_runner.py
```diff
@@ -481,6 +481,52 @@ def test_invalid_header_coding_fails(self):
     with self.assertRaisesRegex(CommandException, r'Invalid non-ASCII'):
       HandleHeaderCoding(headers)
 
+  def test_skip_update_check(self):
```
sameer-google: In test_skip_update_check, we wrap four sequential blocks in with mock.patch.object(boto_util, 'HasUserSpecifiedGsHost', return_value=False):. Instead of repeating this block four times, mock it once at the top of the test method (e.g., as a decorator) and just update mock_has_host.return_value = True for the final scenario.

gslib/tests/test_daisy_chain_wrapper.py
risha-vijayvargiya-22: DaisyChainWrapper.read currently does not check for negative read sizes (meaning daisy_chain_wrapper.read(-1) will not throw a BadRequestException but instead bypass the check and hang in an infinite loop). In order to add this test scenario, we will need changes in daisy_chain_wrapper.py

gslib/tests/test_cors.py
risha-vijayvargiya-22: Currently, gsutil cors get only expects and processes a single bucket URL (the first argument). Any subsequent arguments are silently ignored. For example, running gsutil cors get gs://bucket1 s3://bucket2 successfully fetches the CORS configuration for gs://bucket1 and does not raise any error about the second bucket. If we want to add the test then it will need changes in the python file.

gslib/tests/test_daisy_chain_wrapper.py
sameer-google: so this is an exisitng bug/feature extension in gsutil, but we do not intend to fix this now. Let's keep this as it is.

gslib/tests/test_cors.py
sameer-google: same as above, let's not change in gsutil core logic. Let's keep this purely a test coverage addition PR.

sameer-google (approved):

